PR #7, opened 2015-08-02
Cleanup Intellij Warnings

vkostyukov:
Taking an advatage of using Intellij IDEA and cleaning up some warinings:

* Unnecessary parentheses
* Anonynous function convertible to a method
* Unnecessary `final`
* Unused imports

### DIFF
--- a/core/src/main/scala/io/jfc/CursorOp.scala
+++ b/core/src/main/scala/io/jfc/CursorOp.scala
@@ -24,9 +24,9 @@ case class El(o: CursorOpElement, succeeded: Boolean) extends CursorOp {
 }
 
 object CursorOp {
-  def apply(o: CursorOpElement): CursorOp = El(o, true)
+  def apply(o: CursorOpElement): CursorOp = El(o, succeeded = true)
   def reattemptOp: CursorOp = Reattempt
-  def failedOp(o: CursorOpElement): CursorOp = El(o, false)
+  def failedOp(o: CursorOpElement): CursorOp = El(o, succeeded = false)
 
   implicit val eqCursorOp: Eq[CursorOp] = Eq.instance {
     case (Reattempt, Reattempt) => true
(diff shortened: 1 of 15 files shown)


core/src/main/scala/io/jfc/CursorOp.scala
```diff
@@ -24,9 +24,9 @@ case class El(o: CursorOpElement, succeeded: Boolean) extends CursorOp {
 }
 
 object CursorOp {
-  def apply(o: CursorOpElement): CursorOp = El(o, true)
+  def apply(o: CursorOpElement): CursorOp = El(o, succeeded = true)
```
travisbrown: What's the motivation for these?

travisbrown:
Thanks! This looks good to me, with the exception of the introduction of named parameters in a few places where I don't think they add much.

core/src/main/scala/io/jfc/Printer.scala
```diff
@@ -129,13 +129,13 @@ final case class Printer(
       jsonString.span(Printer.isNormalChar) match {
         case (prefix, suffix) =>
           builder.append(prefix)
-          if (suffix.nonEmpty) appendJsonString(suffix, false)
+          if (suffix.nonEmpty) appendJsonString(suffix, normalChars = false)
```
travisbrown: Oh, so it looks like it just does these for booleans? I guess it kind of makes sense for clarity…

travisbrown:
Okay, I'm sold on the named parameter for booleans—will merge when green. Thanks!

vkostyukov:
Wait. I'm actually not sure we need them. Well Intellij suggests to use them for Boolean params, since it's not clear what `true` or `false` means without looking at the  function signature. But I recall some discussions about named params in Scala. As far as I remember, they considered "inpure".

travisbrown:
I wouldn't want an API that was painful without named parameter, but I'm fine with this kind of usage—the reasoning makes sense to me.

PR #8, opened 2015-08-02
Begin using Scaladoc groups

travisbrown:
This is just a start, but I find that grouping methods makes the API docs [much more useful](https://travisbrown.github.io/jfc/api/#io.jfc.Encode$).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -35,7 +35,6 @@ lazy val baseSettings = Seq(
     }
   ),
   libraryDependencies ++= Seq(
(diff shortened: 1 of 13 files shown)


vkostyukov:
Great idea!

Just out of curiosity, why give up on kind-projector?

travisbrown:
@vkostyukov Scaladoc crashes on kind-projector's `?`. I was only using it in a couple of places, but if we want to reintroduce it we can look for workarounds.

PR #14, opened 2015-08-03
Fix #1: Rename toJson to asJson

### LINKED ISSUE #1: Rename toJson to asJson
@travisbrown what do you think about renaming `toJson` to `asJson`? I'm a huge fan of symmetric names for methods that do symmetirc things. Now you can go from `A` to `Json` with `toJson`, but in opposite way (`Json` to `A`) is `as[A]`.

If it sounds good to you I can put together a PR.

vkostyukov:
I went ahead and closed the first (#1) ticket in jfc!

Please, note that I had to refactor the benchamark a bit to workaround the fact that `asJson` is available in both libraries Argonaut and jfc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,10 +34,10 @@ defined class Qux
 scala> val foo: Foo = Qux(13, Some(14.0))
 foo: Foo = Qux(13,Some(14.0))
 
-scala> foo.toJson.noSpaces
+scala> foo.asJson.noSpaces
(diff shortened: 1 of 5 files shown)


travisbrown:
Thanks!—this looks good to me, except that one import seems to be missing. I've just pushed a fix, and will merge if it passes.

PR #15, opened 2015-08-03
Included method for creating `Json` from `BigDecimal`s

propensive:


propensive:
I called the method `bigDecimal` rather than overloading `number`, which was about the only choice I had.

travisbrown:
The `Json` API might change a bit more before the 0.1 release, but for now at least `bigDecimal` sounds right to me—I'd like to avoid overloading `number`. So thanks!—will merge when green.

travisbrown:
I love how @codecov-io just makes up numbers sometimes…

propensive:
Maybe they should give the percentage to 1 significant figure...

PR #16, opened 2015-08-04
Some efficiency improvements for jfc.

non:
Here's a list of changes:

 1. Introduce typesafe === and =!= equality methods.
 2. Introduce static True and False values.
 3. Avoid reifying Eq[List[Json]] in each equality check.
 4. Avoid allocating a List[Json] for each array comparison.
 5. Switch parser from eager to lazy number parsing.
 6. Parse to static True/False values for less memory use.

Full disclosure: I haven't set up benchmarks for these yet.
I'm pretty sure these changes will help reduce the memory
footprint of using jfc, and speed up things like equality
tests, but obviously more testing is needed.

### DIFF
--- a/core/src/test/scala/io/jfc/test/CodecTests.scala
+++ b/core/src/test/scala/io/jfc/test/CodecTests.scala
@@ -33,7 +33,11 @@ trait CodecTests[A] extends Laws {
   def codec(implicit A: Arbitrary[A], eq: Eq[A]): RuleSet = new DefaultRuleSet(
     name = "codec",
     parent = None,
-    "roundTrip" -> Prop.forAll((a: A) => isEqToProp(laws.codecRoundTrip(a)))
+    "roundTrip" -> Prop.forAll { (a: A) =>
+      val result: IsEq[Xor[DecodeFailure, A]] = laws.codecRoundTrip(a)
+      if (result.lhs != result.rhs) println(result)
+      result
+    }
   )
 }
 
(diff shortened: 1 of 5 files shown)


travisbrown:
Thanks! One question: do you mind setting the `ReturnChecker` level to warning in the Scalastyle config? I think it's reasonable here, and that's the only reason Travis CI is failing.

core/src/test/scala/io/jfc/test/CodecTests.scala
```diff
@@ -33,7 +33,11 @@ trait CodecTests[A] extends Laws {
   def codec(implicit A: Arbitrary[A], eq: Eq[A]): RuleSet = new DefaultRuleSet(
     name = "codec",
     parent = None,
-    "roundTrip" -> Prop.forAll((a: A) => isEqToProp(laws.codecRoundTrip(a)))
+    "roundTrip" -> Prop.forAll { (a: A) =>
+      val result: IsEq[Xor[DecodeFailure, A]] = laws.codecRoundTrip(a)
+      if (result.lhs != result.rhs) println(result)
```
travisbrown: Do we need the `println`?

non:
Oh hey, I just saw you merged this.

I was using the `println` to debug a test failure during development. I should have removed it.

travisbrown:
No worries! The project's three days old—I figured the build could be broken for a few minutes. Thanks again—

PR #23, opened 2015-08-10
Rename all uses of `Jfc` in *.scala files to `Circe`

propensive:


travisbrown:
Thanks, @propensive! At least it's all tests except for `JfcSupportParser`, and that's not really public-facing API, so I guess I'll just let this wait for the other stuff I want to get into 0.2.0.

travisbrown:
…unless someone needs to refer to `JfcSupportParser` and would prefer a release with this change sooner rather than later.

PR #24, opened 2015-08-11
Add support for patching case classes with JSON

travisbrown:
As [requested](https://gitter.im/travisbrown/circe?at=55c8e6750c29567545d98e4a) by @vkostyukov, this addition supports the derivation of case class "updaters":

```scala
scala> import io.circe.generic.auto._, io.circe.jawn._
import io.circe.jawn._

scala> case class Foo(i: Int, xs: List[String])
defined class Foo

scala> val foo = Foo(0, Nil)
foo: Foo = Foo(0,List())

scala> decode[Foo => Foo]("""{ "i": 13 }""").map(_(foo))
res0: cats.data.Xor[io.circe.Error,Foo] = Right(Foo(13,List()))

scala> decode[Foo => Foo]("""{ "xs": ["a", "b"] }""").map(_(foo))
res1: cats.data.Xor[io.circe.Error,Foo] = Right(Foo(0,List(a, b)))

scala> decode[Foo => Foo]("""{ "xs": ["a", "b"], "i": 1 }""").map(_(foo))
res2: cats.data.Xor[io.circe.Error,Foo] = Right(Foo(1,List(a, b)))
```

This is closely related to ["incomplete" decoder derivation](https://meta.plasm.us/posts/2015/06/21/deriving-incomplete-type-class-instances/) and the motivation is similar: we want to avoid lots of case classes like `PatchedTodo` here:

```scala
case class Todo(id: String, title: String, completed: Boolean, order: Int)
case class PatchedTodo(title: Option[String], completed: Option[Boolean], order: Option[Int])
```

`Decoder[Foo => Foo]` lets us accomplish the same thing without an extra boilerplate-y case class.

### DIFF
--- a/generic/src/main/scala/io/circe/generic/IncompleteInstances.scala
+++ b/generic/src/main/scala/io/circe/generic/IncompleteInstances.scala
@@ -10,4 +10,10 @@ trait IncompleteInstances {
     complement: Complement.Aux[T, P, R],
     d: Decoder[R]
   ): Decoder[F] = d.map(r => ffp(p => gen.from(complement.insert(p, r))))
+
+  implicit def decodeCaseClassPatch[A, R <: HList, O <: HList](implicit
+    gen: LabelledGeneric.Aux[A, R],
+    patch: PatchWithOptions.Aux[R, O],
+    d: Decoder[O]
+  ): Decoder[A => A] = d.map(o => a => gen.from(patch(gen.to(a), o)))
 }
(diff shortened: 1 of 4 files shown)


vkostyukov:
Omg! This is awesome! Thanks a lot.

Mind publishing a snapshot version? I'd love to use this feature in the workshop.

travisbrown:
@vkostyukov Yep, can definitely publish a snapshot. I also just found a few minor bugs in some of the cursor operations (a couple carried over from Argonaut and a couple of my own invention). These don't interfere with "normal" usage, but I might still try to pull together an 0.2.0 tonight. I'm stuck in LAX for two more hours thanks to another delayed plane, so I've got plenty of time…

PR #28, opened 2015-08-12
More lightweight core

alexarchambault:
Opening this PR to discuss the points below, which the code illustrates (although it could be merged if you're fine with it).

I've been experimenting a bit with [upickle](https://github.com/lihaoyi/upickle-pprint) recently (mainly for its scalajs support and its test suite), and I wish circe could be as lightweight as it, with almost no dependencies.

It turns out it can be done, which this PR illustrates. It puts all the cats related parts of core into a subproject of its own (circe-cats), resulting in a lightweight core with no dependency. It makes circe even more lightweight, it makes it have the best features of both argonaut and upickle (this just rocks), and it doesn't prevent it becoming the JSON library of choice when using cats (thanks to the circe-cats module).

Only possible issue I see is if you were planning to rely more on cats in the very core of circe. That would make this a bit more complicated.


alexarchambault:
(I kept the full history of my commits during the dev of this, I could squash them differently to make their changes more clear.)

alexarchambault:
Build failing because of transient resolution errors... (Travis-CI issue)

travisbrown:
@alexarchambault Wow, thanks for putting this together! I vaguely thought about going this route, but liked the convenience of `Xor`, keeping type class instances in one place, etc. I had some other ideas about using cats in core, but nothing definite enough to rule this change out.

I've created a new issue (#29) for the question so we can keep this PR focused on the implementation.

(Also: I restarted the failed build and it passed the second time around, but dropped the coverage dramatically. This should be a pretty easy fix—I'm pretty sure I've figured out how to get scoverage to count tests in a separate project before.)

alexarchambault:
Just split the only commit of this PR into three (hopefully more readable) ones.

Issue #30, opened 2015-08-12
Use export-hook in core


travisbrown:
I'm creating this issue to follow up on a comment by @dwijnand [on gitter](https://gitter.im/travisbrown/circe?at=55cb4fe20c29567545d9d7c0):

> The only thought I had about #29 is the need for the imports, and I that made me think if maybe core should use export-hook

I've played around with [export-hook](https://github.com/milessabin/export-hook) in core, but I really want to push on the experiment of keeping generic derivation modular for as long as possible. I don't love my current approach of getting priorities right by balancing a tower of instance-providing traits and using a `Secondary` wrapper, but it seems to work reasonably well, and is still (I think) not completely unintelligible.

travisbrown:
I'm removing the question tag since @milessabin has just convinced me that export-hook can do what I want. I'll tackle this next week unless someone beats me to it.

travisbrown:
Fixed by #43.

travisbrown:
Fixed by #43.

dwijnand:
:+1:

PR #43, opened 2015-08-27
Use export-hook for generic derivation

travisbrown:
This is a first stab at using [export-hook](https://github.com/milessabin/export-hook) for generic derivation (as discussed in #30). It lets us get rid of the `Secondary` wrappers and some of the complex machinery for getting implicit prioritization right.

Usage is the same, except now we don't need generic derivation for tuples, and I've also removed support for incomplete instances from `semiauto` (it seemed a little out of place).

I also went ahead and updated the Jawn version from 0.8.0 to 0.8.3, but that's completely unrelated to the rest of the changes.


travisbrown:
Oh, also I moved the Shapeless type classes that aren't really circe-specific into a new `io.circe.generic.util` package.

PR #46, opened 2015-08-29
Rename projects to make JVM usage less verbose

travisbrown:
I think I saw @non suggest this renaming for cats a few days ago (thought it was in gitter but not seeing it at the moment). In any case I'd like to borrow the idea for circe.


travisbrown:
Oh, this doesn't actually work. Will fix when I'm back at a computer.

PR #47, opened 2015-08-30
Remove io.circe.auto._ from example in README

groz:
Replaced by import io.circe.generic.auto._

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,6 @@ root directory of this repository):
 ```scala
 scala> import io.circe._, io.circe.generic.auto._, io.circe.jawn._, io.circe.syntax._
 import io.circe._
-import io.circe.auto._
 import io.circe.jawn._
 import io.circe.syntax._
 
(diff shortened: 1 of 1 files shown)


travisbrown:
Thanks, @groz! One thing—do you mind changing the line to `import io.circe.generic.auto._` instead of deleting it, so that this represents a complete REPL session?

groz:
@travisbrown isn't it already imported on the first line?

travisbrown:
@groz The idea is that that whole code section should represent a copy-and-pasted REPL session, so when you see this:

```scala
scala> sealed trait Foo
defined trait Foo
```

The second line is what the REPL prints after you enter everything after the prompt in the first.

The Scala REPL allows you to paste in a complete REPL session like this (it'll ignore the prompts and result lines), so using it in documentation has the advantage of both showing what will happen and being directly copy-and-paste-able.

groz:
Ah, understood. Thanks.

travisbrown:
@groz Thanks again for catching this! Will merge when green.

PR #66, opened 2015-09-18
Fix StackOverflowError when decoding large JSON arrays

alexarchambault:
This PR fixes a StackOverflowError raised while decoding large JSON arrays. See the added test, which doesn't pass without this.


travisbrown:
Thanks for catching this!

PR #68, opened 2015-10-02
Add UUID support

vkostyukov:
I was in the middle of migrating the [Todo application](https://github.com/vkostyukov/finch-101/blob/master/src/main/scala/i/f/workshop/finch/Todo.scala) to UUIDs, which are [natively supported in Finch] (https://github.com/finagle/finch/pull/399) and realized there is no encoders/decoders available in Circe for the `java.util.UUID` type.

UUIDs are very very poluar in the modern web so it would be nice to support them out of the box in Circe, given that the decoder implementation [might be tricky](http://mark.koli.ch/thoughts-on-using-raw-uuids-in-your-web-application-or-web-service-paths).

Although, I'd totally understand if UUID codecs is something that's considered as out-of-the-scope of the library. So, please feel free drop this PR if you don't think this should be part of the `core` package.

### DIFF
--- a/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -1,5 +1,7 @@
 package io.circe
 
+import java.util.UUID
+
 import cats.Monad
 import cats.data.{ Kleisli, NonEmptyList, Validated, Xor }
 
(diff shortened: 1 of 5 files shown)


vkostyukov:
That's weird. I have not idea why it keeps failing:

```
[/home/travis/build/travisbrown/circe/tests/jvm/target/scala-2.10/scoverage-data/scoverage.coverage.xml]
java.io.EOFException
	at java.io.ObjectInputStream$BlockDataInputStream.peekByte(ObjectInputStream.java:2601)
	at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1319)
	at java.io.ObjectInputStream.readObject(ObjectInputStream.java:371)
	at org.scalatest.tools.Framework$ScalaTestRunner$Skeleton$1$React.react(Framework.scala:949)
	at org.scalatest.tools.Framework$ScalaTestRunner$Skeleton$1.run(Framework.scala:938)
	at java.lang.Thread.run(Thread.java:745)
java.io.EOFException
```
In the meantime, I will keep hitting the "rebuild" button.

travisbrown:
@vkostyukov I'd been vaguely thinking of providing instances for some `java.time` classes and other Java types in a new circe-java module, but this is one that I agree should be in core. I'm curious about whether it passes tests on Scala.js—will check shortly.

About the test flakiness: I've also noticed a lot of these errors recently—usually one or sometimes two per build. I'm not sure why they're happening, and I haven't been able to reproduce them locally, but we (I) should look into it.

ceedubs:
:+1:

travisbrown:
Thanks, @vkostyukov! This passes Scala.js tests locally, so I'm merging.

PR #72, opened 2015-10-22
Update versions (with one other small clarification)

travisbrown:


fthomas:
This build https://travis-ci.org/travisbrown/circe/jobs/86886404 which should use Scala 2.10.6 compiles with 2.11.7. I think you are also hitting https://github.com/scoverage/sbt-scoverage/issues/146

vkostyukov:
:+1:

travisbrown:
@fthomas Oh, ouch! I guess the best thing to do is roll back to 1.2.0 for now?

fthomas:
@travisbrown That is what I did in refined.

travisbrown:
@fthomas Thanks for catching that! I've just pushed a fix.

PR #74, opened 2015-10-26
Add codecs for JsonObject and JsonNumber

ochrons:


travisbrown:
Thanks! Looks good to me—I'll merge when green and publish a new snapshot shortly.

One thing: do you mind adding an issue noting that we should have an `Arbitrary` instance for `JsonNumber` so that we can test its codecs? I don't think that needs to happen in this PR, but I'd like to remember that it should happen eventually.

tests/shared/src/test/scala/io/circe/CodecSuites.scala
```diff
@@ -75,6 +75,7 @@ class CatsCodecSuite extends CirceSuite {
 
 class CirceCodecSuite extends CirceSuite {
   checkAll("Codec[Json]", CodecTests[Json].codec)
+  checkAll("Codec[JsonObject]", CodecTests[JsonObject].codec)
```
vkostyukov: Sorry for the silly question: Why don't we test `Codec[JsonNumber]`?

travisbrown: It's definitely a good question (see my comment above). We should test it, but writing a good `Arbitrary` instance for `JsonNumber` is non-trivial, and since the codec is extremely simple I don't think it should hold up this PR.

travisbrown:
Also, for the record, these additions are discussed briefly [here](https://gitter.im/travisbrown/circe?at=562e2ef1f2adcffe367faf4a).

tests/shared/src/test/scala/io/circe/CodecSuites.scala
vkostyukov: Oh, got it. Thanks Travis!

PR #77, opened 2015-10-29
Update to export-hook 1.0.2 and fix #73

### LINKED ISSUE #73: Generic derivation chokes on recursive types in certain situations
The generic derivation mechanism can handle recursive type definitions, but it will choose the wrong codecs when there's an intervening non-derived type. For example (the error messages also need some work, but I'll save that for another issue):

```scala
scala> import io.circe._, io.circe.jawn.decode, io.circe.generic.auto._
import io.circe._
import io.circe.jawn.decode

scala> case class Foo(o: Option[Foo])
defined class Foo

scala> decode[Foo]("{}")
res0: cats.data.Xor[io.circe.Error,Foo] = Left(io.circe.DecodingFailure: Attempt to decode value on failed cursorEl(DownField(o),false))

scala> decode[Foo]("""{ "o": null }""")
res1: cats.data.Xor[io.circe.Error,Foo] = Left(io.circe.DecodingFailure: CNilEl(DownField(o),true))
```

But this works:

```scala
scala> decode[Foo]("""{ "o": { "None": {} } }""")
res2: cats.data.Xor[io.circe.Error,Foo] = Right(Foo(None))
```

And similarly:

```scala
scala> Foo(None).asJson
res3: io.circe.Json =
{
  "o" : {
    "None" : {

    }
  }
}
```

This makes it clear that circe is using the derived instance for `Option[A: Decoder]` instead of the usual one. This is bad, and should be fixed.

Thanks to [Caballero](http://stackoverflow.com/users/717839/caballero) on Stack Overflow for [noticing this](http://stackoverflow.com/questions/33318802/scala-parse-json-of-more-than-22-elements-into-case-class/33319168?noredirect=1#comment54441746_33319168).

travisbrown:
Fixes #73 (and adds a test to verify), so there's now no issue with recursive nesting of types with a mix of derived and non-derived instances:

```scala
scala> import io.circe._, io.circe.jawn.decode, io.circe.generic.auto._
import io.circe._
import io.circe.jawn.decode

scala> case class Foo(o: Option[Foo])
defined class Foo

scala> decode[Foo]("{}")
res0: cats.data.Xor[io.circe.Error,Foo] = Right(Foo(None))

scala> decode[Foo]("""{ "o": null }""")
res1: cats.data.Xor[io.circe.Error,Foo] = Right(Foo(None))
```

Will update, merge, and publish a new snapshot as soon as export-hook 1.0.2 is published.

### DIFF
--- a/tests/shared/src/test/scala/io/circe/generic/AutoDerivedSuite.scala
+++ b/tests/shared/src/test/scala/io/circe/generic/AutoDerivedSuite.scala
@@ -55,12 +55,28 @@ class AutoDerivedSuite extends CirceSuite {
       Arbitrary(atDepth(0))
   }
 
+  case class RecursiveWithOptionExample(o: Option[RecursiveWithOptionExample])
+
+  object RecursiveWithOptionExample {
+    implicit val eqRecursiveWithOptionExample: Eq[RecursiveWithOptionExample] =
+      Eq.fromUniversalEquals
+
+    private def atDepth(depth: Int): Gen[RecursiveWithOptionExample] = if (depth < 3)
+      Gen.option(atDepth(depth + 1)).map(
+        RecursiveWithOptionExample(_)
+      ) else Gen.const(RecursiveWithOptionExample(None))
+
+    implicit val arbitraryRecursiveWithOptionExample: Arbitrary[RecursiveWithOptionExample] =
+      Arbitrary(atDepth(0))
+  }
+
   checkAll("Codec[Tuple1[Int]]", CodecTests[Tuple1[Int]].codec)
   checkAll("Codec[(Int, Int, Foo)]", CodecTests[(Int, Int, Foo)].codec)
   checkAll("Codec[Qux[Int]]", CodecTests[Qux[Int]].codec)
   checkAll("Codec[Foo]", CodecTests[Foo].codec)
   checkAll("Codec[OuterCaseClassExample]", CodecTests[OuterCaseClassExample].codec)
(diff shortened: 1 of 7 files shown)


tests/shared/src/test/scala/io/circe/generic/AutoDerivedSuite.scala
```diff
@@ -55,12 +55,28 @@ class AutoDerivedSuite extends CirceSuite {
...
+      Eq.fromUniversalEquals
+
+    private def atDepth(depth: Int): Gen[RecursiveWithOptionExample] = if (depth < 3)
+      Arbitrary.arbitrary[Option[RecursiveWithOptionExample]].map(
+        RecursiveWithOptionExample(_)
+      ) else Gen.const(RecursiveWithOptionExample(None))
+
+    implicit val arbitraryRecursiveWithOptionExample: Arbitrary[RecursiveWithOptionExample] =
+      Arbitrary(atDepth(0))
+  }
```
xuwei-k: `depth: Int` is always `0` ?

alexarchambault:
:+1: for the removal of the way `Priority` was used. It wasn't intended to be used this way, only when the return type of the implicit method at hand was the same as the first type arg of `Priority` (like [here](https://github.com/milessabin/shapeless/blob/master/core/src/test/scala/shapeless/priority.scala#L272-L275)).

tests/shared/src/test/scala/io/circe/generic/AutoDerivedSuite.scala
travisbrown: @xuwei-k Ah, thanks! I've just pushed a fix.

Issue #39, opened 2015-08-20
Monocle lenses for Circe


adelbertc:
Dependent on: https://github.com/julien-truffaut/Monocle/issues/235

travisbrown:
I'd love to see this happen and would happily add a circe-lenses module once Monocle's ready.

travisbrown:
Marking this _in progress_ since I've started on it in #78 (which currently requires both cats and Scalaz).

travisbrown:
#78 is now merged and `circe-optics` is available in the 0.3.0 snapshot.

PR #78, opened 2015-10-29
Initial sketch of optics module

travisbrown:
This is a partial port [from Argonaut](https://github.com/argonaut-io/argonaut/tree/sliceitallup/argonaut-monocle) in response to #39 and [this question](https://twitter.com/schrepfler/status/659803714614599682). If we can get some docs and tests in place before Shapeless 2.3.0 is out, I'd be happy to publish this with 0.2.0, but I don't want to wait for it.

### DIFF
--- a/optics/src/main/scala/io/circe/optics/JsonPath.scala
+++ b/optics/src/main/scala/io/circe/optics/JsonPath.scala
@@ -1,0 +1,27 @@
+package io.circe.optics
+
+import io.circe.{ Decoder, Encoder, Json }
+import io.circe.optics.JsonOptics._, io.circe.optics.JsonObjectOptics._
+import monocle.{ Optional, Prism }
+import monocle.function.index
+import monocle.std.list._
+import scala.language.dynamics
+
+case class JsonPath(opt: Optional[Json, Json]) extends Dynamic {
+  def selectDynamic(field: String): JsonPath =
+    JsonPath(opt.composePrism(jsonObject).composeOptional(index(field)))
+
+  def at(i: Int): JsonPath =
+    JsonPath(opt.composePrism(jsonArray).composeOptional(index(i)))
+
+  def as[A](implicit decode: Decoder[A], encode: Encoder[A]): Optional[Json, A] =
+    opt.composePrism(
+      Prism((j: Json) => decode.decodeJson(j).toOption)(encode(_))
+    )
+
+  def `*`: Optional[Json, Json] = opt
+}
+
+object JsonPath {
+  val root: JsonPath = JsonPath(Optional.id)
+}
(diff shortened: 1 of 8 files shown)


travisbrown:
Example usage:

```scala
import cats.data.Xor
import io.circe._, generic.auto._, jawn._, optics._
import monocle.macros.Lenses

@Lenses case class Foo(count: Int)
@Lenses case class Qux(foo: Foo, s: String)

val toQux = JsonPath.root.x.y.at(2).as[Qux]
val incrementCount = toQux.composeLens(Qux.foo).composeLens(Foo.count).modify(_ + 1)

val Xor.Right(json) = parse("""
  { "x" : { "y" : [0, 1, { "foo" : { "count" : 3  }, "s" : "abcde" }] } }
""")
```

And then:

```scala
scala> json
res0: io.circe.Json =
{
  "x" : {
    "y" : [
      0,
      1,
      {
        "foo" : {
          "count" : 3
        },
        "s" : "abcde"
      }
    ]
  }
}

scala> incrementCount(json)
res1: io.circe.Json =
{
  "x" : {
    "y" : [
      0,
      1,
      {
        "s" : "abcde",
        "foo" : {
          "count" : 4
        }
      }
    ]
  }
}
```

optics/src/main/scala/io/circe/optics/JsonPath.scala
```diff
@@ -0,0 +1,27 @@
...
+import scala.language.dynamics
+
+case class JsonPath(opt: Optional[Json, Json]) extends Dynamic {
+  def selectDynamic(field: String): JsonPath =
+    JsonPath(opt.composePrism(jsonObject).composeOptional(index(field)))
+
+  def at(i: Int): JsonPath =
+    JsonPath(opt.composePrism(jsonArray).composeOptional(index(i)))
+
+  def as[A](implicit decode: Decoder[A], encode: Encoder[A]): Optional[Json, A] =
```
julien-truffaut: I believe that `as` is not lawful because `decoder` is not injective e.g.
```scala
case class Point(x: Int, y: Int)

val decoder: Decoder[Point] = ...
decoder.decodeJson(obj("x" -> int(1), "y" -> int(2))) == decoder.decodeJson(obj("x" -> int(1), "y" -> int(2), "z" -> int(5))) 
```

julien-truffaut:
You can upgrade monocle to 1.2.0-M2 and define a `Plated` instance

travisbrown:
I'm going to merge this when it's green on the understanding that the `JsonPath` API is subject to change.

dwijnand:
The readme brought me here ("We are likely to add an experimental optics subproject in 0.3.0"), so given this is merged I think the readme should be updated.

PR #82, opened 2015-11-02
Update to cats 0.3.0

travisbrown:
It [sounds likely](https://github.com/non/cats/issues/594) that cats 0.3.0 will be released in time for circe 0.2.0, so I think it makes sense to go ahead and update to the snapshot.


travisbrown:
Superseded by #90.

PR #84, opened 2015-11-03
Add spray-json benchmark

liff:
Another point of comparison.

Here are the results from my laptop:

```
info] Benchmark                       Mode  Cnt      Score      Error  Units
[info] DecodingBenchmark.decodeFoosA  thrpt   20   1411,502 ±   62,909  ops/s
[info] DecodingBenchmark.decodeFoosC  thrpt   20   1983,638 ±   72,286  ops/s
[info] DecodingBenchmark.decodeFoosP  thrpt   20   1740,702 ±   21,146  ops/s
[info] DecodingBenchmark.decodeFoosS  thrpt   20   8596,047 ±  146,904  ops/s
[info] DecodingBenchmark.decodeIntsA  thrpt   20   7722,981 ±  208,717  ops/s
[info] DecodingBenchmark.decodeIntsC  thrpt   20  11226,785 ±  665,015  ops/s
[info] DecodingBenchmark.decodeIntsP  thrpt   20  14598,128 ±  144,416  ops/s
[info] DecodingBenchmark.decodeIntsS  thrpt   20  32657,558 ± 2755,072  ops/s
[info] EncodingBenchmark.encodeFoosA  thrpt   20   5493,093 ±   98,725  ops/s
[info] EncodingBenchmark.encodeFoosC  thrpt   20   6198,165 ±   52,079  ops/s
[info] EncodingBenchmark.encodeFoosP  thrpt   20   2230,278 ±   35,819  ops/s
[info] EncodingBenchmark.encodeFoosS  thrpt   20   5196,096 ±   30,894  ops/s
[info] EncodingBenchmark.encodeIntsA  thrpt   20  71120,576 ± 1242,305  ops/s
[info] EncodingBenchmark.encodeIntsC  thrpt   20  88010,821 ± 1199,913  ops/s
[info] EncodingBenchmark.encodeIntsP  thrpt   20  52768,451 ± 4772,556  ops/s
[info] EncodingBenchmark.encodeIntsS  thrpt   20  31652,292 ±  746,873  ops/s
[info] ParsingBenchmark.parseFoosA    thrpt   20   2201,660 ±  209,243  ops/s
[info] ParsingBenchmark.parseFoosC    thrpt   20   2947,012 ±   97,429  ops/s
[info] ParsingBenchmark.parseFoosP    thrpt   20   1781,731 ±   84,997  ops/s
[info] ParsingBenchmark.parseFoosS    thrpt   20   3845,950 ±   60,479  ops/s
[...]

PR #86, opened 2015-11-03
Add Spray JSON benchmark results to README

travisbrown:
I'm pushing these results as a follow-up to #84, but some of the Spray allocation numbers are so low that I'm a little suspicious that the benchmarks may not be doing what we think they're doing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,37 +132,45 @@ better):
 ```
 Benchmark                       Mode  Cnt      Score     Error  Units
 
-DecodingBenchmark.decodeFoosA  thrpt   80   1185.694 ±   1.947  ops/s
-DecodingBenchmark.decodeFoosC  thrpt   80   1972.754 ±   6.829  ops/s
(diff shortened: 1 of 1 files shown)


travisbrown:
There's some discussion of this issue [here](https://gitter.im/travisbrown/circe?at=56410aba565da33267ff63d2), with confirmation via Google's allocation-instrumenter by @liff, so I'm convinced that the fact that Spray just throws exceptions on decoding failures accounts for its advantage.

To quote myself from Gitter:

> It's a big enough difference API-wise that I think it needs some specific discussion to make sure that people evaluate the benchmarks fairly. My plan is to add a new docs/benchmarks.md page that includes the full benchmark results, and to have a link and summary in the README.

I'll close this issue when that's done.

As a side note, if we want circe to be competitive here, we could add a `private[circe] def decodeUnsafe(c: HCursor): A` method to `Decoder` and then use it in the collection decoders. I'm not convinced this is worth it at the moment, but I'd listen if someone wanted to make a case.

travisbrown:
This is finally done in #233.

PR #79, opened 2015-11-02
Initial sketch of generic derivation with default values

travisbrown:
This needs some work but I wanted to run the API past the people who've been asking for this feature.

It's now possible to derive decoders that will use any available default values for case class members:

```scala
scala> import io.circe.generic.auto.defaults._, io.circe.jawn.decode
import io.circe.generic.auto.defaults._
import io.circe.jawn.decode

scala> case class Foo(a: String, b: Int = 0, c: String, d: Char = 'x')
defined class Foo

scala> decode[Foo]("""{ "a": "X", "c": "Y" }""")
res0: cats.data.Xor[io.circe.Error,Foo] = Right(Foo(X,0,Y,x))

scala> decode[Foo]("""{ "a": "X", "c": "Y", "d": "y" }""")
res1: cats.data.Xor[io.circe.Error,Foo] = Right(Foo(X,0,Y,y))
```

There's also a semi-automatic flavor:

```scala
scala> import io.circe._, io.circe.generic.semiauto._, io.circe.jawn.decode
import io.circe._
import io.circe.generic.semiauto._
import io.circe.jawn.decode

scala> case class Foo(a: String, b: Int = 0, c: String, d: Char = 'x')
defined class Foo

scala> implicit val decodeFoo: Decoder[Foo] = deriveFor[Foo].decoderWithDefaults
decodeFoo: io.circe.Decoder[Foo] = io.circe.Decoder$$anon$2@74e8303d

scala> decode[Foo]("""{ "a": "X", "c": "Y" }""")
res0: cats.data.Xor[io.circe.Error,Foo] = Right(Foo(X,0,Y,x))

scala> decode[Foo]("""{ "a": "X", "c": "Y", "d": "y" }""")
res1: cats.data.Xor[io.circe.Error,Foo] = Right(Foo(X,0,Y,y))
```

There are probably a few `Lazy`s in there that could be removed, and there's still an issue (flagged in a comment in `SemiautoDerivedSuite`) with semi-automatic instances defined in the companion object.

Issue #57, opened 2015-09-02
Add non-compilation tests


travisbrown:
In #55 @tylerprete points out that `generic.auto` will find an `Encoder` instance for `Map[String, Map[String, Object]]`, which is definitely not a good thing.

We should add tests to confirm that circe won't find encoders or decoders for types like `Object`, `Product`, etc. (we might need to wait for a fix for [this Shapeless bug](https://github.com/milessabin/shapeless/issues/453) before we can merge).

travisbrown:
Adding this to the 0.2.0 milestone since we're waiting on Shapeless 2.3.0 anyway.

PR #87, opened 2015-11-04
Use Shapeless 2.2.5 for 0.2.0

### LINKED ISSUE #55: Odd incorrect encoding of heterogenous map
With circe 0.1.1, I'm running the following program:

```
import io.circe._
import io.circe.jawn._
import io.circe.syntax._

object Main {
  def main(args: Array[String]) {
    val funnyMap1 = Map("a" -> Map("str" -> "b", "map" -> Map("d" -> 1)))
    val weirdJs1 = funnyMap1.asJson.noSpaces 

    println(weirdJs1)
  }
}
```

With the result:
```
{"a":{"str":{},"map":{}}}
```

travisbrown:
After publishing 0.1.2 this morning I realized that after some recent changes to `generic` and the upgrade to export-hook 1.0.2, we no longer absolutely need new changes in Shapeless 2.3.0.

The changes here would allow us to publish circe 0.2.0 today against cats 0.2.0 and Shapeless 2.2.5. We'd then be able to spend some time with the bigger changes in #78, #79, and #85 and publish them in circe 0.3.0 when Shapeless 2.3.0 and cats 0.3.0 are available.

I hadn't fixed #55 and #57 because I was expecting them to be resolved when Shapeless's [#453](https://github.com/milessabin/shapeless/issues/453) is closed, but if we're publishing 0.2.0 against Shapeless 2.2.5 it makes sense to fix them now (so I did). If we go this route I'll open an issue about removing the ambiguous instances before 0.3.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,9 +18,9 @@ you can just add the following to your build:
 
 ```scala
 libraryDependencies ++= Seq(
-  "io.circe" %% "circe-core" % "0.1.1",
-  "io.circe" %% "circe-generic" % "0.1.1",
-  "io.circe" %% "circe-jawn" % "0.1.1"
+  "io.circe" %% "circe-core" % "0.2.0",
+  "io.circe" %% "circe-generic" % "0.2.0",
+  "io.circe" %% "circe-parse" % "0.2.0"
 )
 ```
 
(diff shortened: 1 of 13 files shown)


travisbrown:
For some reason the non-compilation tests that confirm that there are no `Decoder[Object]`, etc. instances when using fully-automatic derivation are failing on 2.10 (using either ScalaTest's `assertTypeError` or Shapeless's `illTyped`), even though these instances aren't available in the REPL.

Because we're blocking these instances primarily to avoid confusing users, and since all other tests pass, I'm tempted to write this off as an issue with export-hook (or the non-compilation test mechanisms) and simply skip the tests on 2.10 (which is what I'm currently doing in this PR).

travisbrown:
Okay, I'm going to fix #55 here but kick #57 down the road because I don't want to get bogged down in fighting with test-related issues that only show up on 2.10.

ngbinh:
:+1: would love to use circe now.

yanana:
:+1:

travisbrown:
This PR has become a kind of grab-bag of last little things for 0.2.0, but I think they're mostly uncontroversial, so once it's green I'll publish 0.2.0.

Issue #88, opened 2015-11-04
Stringified object representations don't preserve field order


longcao:
Currently, stringified `io.circe.Json` doesn't appear to preserve field order. This doesn't matter for machines but it would greatly help human readability.

```scala
scala> import io.circe._, io.circe.generic.auto._, io.circe.jawn._, io.circe.syntax._
import io.circe._
import io.circe.generic.auto._
import io.circe.jawn._
import io.circe.syntax._

scala> case class MyClass(a: Int, b: String, c: List[String], d: Boolean)
defined class MyClass

scala> val m = MyClass(5, "asdf", List("zxcv", "qwer", "hjkl"), false)
m: MyClass = MyClass(5,asdf,List(zxcv, qwer, hjkl),false)

scala> m.asJson
res0: io.circe.Json =
{
  "d" : false,
  "c" : [
    "zxcv",
    "qwer",
    "hjkl"
  ],
  "b" : "asdf",
  "a" : 5
}
```

alexarchambault:
@longcao Order is preserved internally, it's just that the default printers don't preserve it in their output, see https://github.com/travisbrown/circe/blob/master/core/shared/src/main/scala/io/circe/Printer.scala#L215.

The following gives back the right order for me:
```scala
Printer.spaces2.copy(preserveOrder = true).pretty(m.asJson)
```
giving
```
res1: String = """
{
  "a" : 5,
  "b" : "asdf",
  "c" : [
    "zxcv",
    "qwer",
    "hjkl"
  ],
  "d" : false
}
"""
```

travisbrown:
Fixed by #89—thanks @longcao, @alexarchambault!

PR #89, opened 2015-11-04
preserve field order when printing

longcao:
This PR addresses #88 and defaults printing to preserve field order. I went ahead and flagged it the same for spaces/no spaces for consistency but that can be up for discussion if anyone feels otherwise.


travisbrown:
Looks reasonable to me!

PR #98, opened 2015-11-10
Fix printing null strings

vkostyukov:
This PR fixes a bug discussed [on Gitter](https://gitter.im/travisbrown/circe?at=56417d9404a883da3838cea4):

Before:

```scala
scala> case class Foo(s: String)
defined class Foo

scala> Foo(null).asJson.noSpaces
java.lang.NullPointerException
  at scala.collection.immutable.StringOps$.length$extension(StringOps.scala:48)
  at scala.collection.immutable.StringOps.length(StringOps.scala:48)
````

After:

```scala
scala> case class Foo(s: String)
defined class Foo

scala> Foo(null).asJson.noSpaces
res0: String = {"s":null}
```

Please note that I'm not quite sure this the right place for the fix.

### DIFF
--- a/tests/shared/src/main/scala/io/circe/tests/PrinterSuite.scala
+++ b/tests/shared/src/main/scala/io/circe/tests/PrinterSuite.scala
@@ -3,6 +3,7 @@ package io.circe.tests
 import io.circe.{ Parser, Printer }
 
 class PrinterSuite(printer: Printer, parser: Parser) extends CirceSuite {
+  checkAll("Printing String", PrinterTests[String].printer(printer, parser))
   checkAll("Printing Unit", PrinterTests[Unit].printer(printer, parser))
   checkAll("Printing Boolean", PrinterTests[Boolean].printer(printer, parser))
   checkAll("Printing Char", PrinterTests[Char].printer(printer, parser))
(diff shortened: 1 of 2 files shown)


tests/shared/src/main/scala/io/circe/tests/PrinterSuite.scala
```diff
@@ -3,6 +3,7 @@ package io.circe.tests
 import io.circe.{ Parser, Printer }
 
 class PrinterSuite(printer: Printer, parser: Parser) extends CirceSuite {
+  checkAll("Printing String", PrinterTests[String].printer(printer, parser))
```
vkostyukov: Question: Would Scalacheck generate `null` as a possible string value?

travisbrown:
As I mentioned on Gitter, I'm not entirely sure this is a bug. I don't think we necessarily need to think of JSON nulls as mapping directly to Scala's `null`, and circe's current approach is to pretend `null` (the Scala / Java one) doesn't exist, with `None` being the equivalent of JSON nulls.

Also Scalacheck will never generate `null` for `String` or other types (unless you define your own `Arbitrary` that does, of course).

vkostyukov:
Thanks @travisbrown! Let me drop this.

PR #99, opened 2015-11-10
[WIP]New ScalaJS Module

chandu0101:
Javascript has `undefined` type http://www.scala-js.org/doc/js-interoperability.html , I added `JUndefined` to core Json please let me know what do you think about. If it doesn't make sense i'll remove it and will document method accordingly ,if its okay to stay then i have an issue in encoding

```scala
  /**
   * @group Encoding
   */
  implicit def encodeJSUndefined[A](implicit e: Encoder[A]): Encoder[js.UndefOr[A]] =
    Encoder.instance(_.fold(Json.empty)(e(_)))
```
If i use `Json.undefined`  instead of `Json.empty` then i am getting match error in printer but i do have `case JUndefined` :s

### DIFF
--- a/core/shared/src/main/scala/io/circe/Json.scala
+++ b/core/shared/src/main/scala/io/circe/Json.scala
@@ -124,34 +124,6 @@ sealed abstract class Json extends Product with Serializable {
   def spaces4: String = Printer.spaces4.pretty(this)
 
   /**
-   * Type-safe equality method.
-   */
-  def ===(that: Json): Boolean = {
-    def arrayEq(x: Seq[Json], y: Seq[Json]): Boolean = {
-      val it0 = x.iterator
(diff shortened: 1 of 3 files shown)


julienrf:
undefined does not exist in JSON.

travisbrown:
I agree with @julienrf—we're not trying to parse arbitrary JavaScript, so I don't see why we'd need support for `undefined`.

chandu0101:
>undefined does not exist in JSON.

Yeah , but js.Objects can have undefined fields new scalajs module is only for this(js.Objects <----> Json) purpose. anyway looks like its not a good fit for core module  i'll remove it and will send a updated PR :)

chandu0101:
sorry for delay, opened new PR #112  please review it :)

PR #103, opened 2015-11-11
Add codecs for refined types

alexarchambault:
See the discussion in https://github.com/travisbrown/circe/issues/101


travisbrown:
This looks good to me, but do you think we should note that this is currently use-at-your-own-risk since the Shapeless versions are mismatched?

alexarchambault:
It is a bit risky right now with circe *snapshots*, although binary incompatibility seems unlikely (refined is only using a limited subset of shapeless features).

The next circe release awaits shapeless 2.3.0, right? I guess refined would have switched to it by that time. We just have to be careful to use the last refined version then. Maybe this should be noted in an issue for circe 0.3.0?

travisbrown:
@alexarchambault I've been meaning to add a "warnings / known issues" section to the README, and I can mention the issue there. I'd prefer not to publish a release where different modules depend on different Shapeless versions, but let's assume for now that refined will be moving to Shapeless 2.3.0 shortly after its release as well.

fthomas:
That is a safe assumption. :-)

PR #104, opened 2015-11-11
Change root project name to circe

alexarchambault:
As the root SBT project is currently called `root`, this gives circe the name `root` in IntelliJ. This PR changes that.


travisbrown:
Works for me, thanks!

PR #107, opened 2015-11-12
Rename some symbolic operators

travisbrown:
I'm not a big fan of symbolic operators, and in a couple of these cases I think they present opportunities for confusion (e.g. why does `(k -> v) +: o` work but not `o + (k -> v)`?), or that they obscure the API (I've had a couple of people say they didn't realize something that there was a method that does what `|||` does for decoders).

If anyone wants to advocate for any of these, I'm happy to discuss removing the deprecations, but otherwise I'd like to remove them in 0.4.0.


alexarchambault:
:+1: for the removal and the deprecations. I'm really no fan of operators.

PR #109, opened 2015-11-12
Add array cursor optimized for navigation

travisbrown:
This is a quick experiment in trying to improve performance ([related](https://github.com/travisbrown/circe/pull/86#issuecomment-155204248)) by adding a new `Cursor` implementation for arrays that's optimized for navigation (as opposed to modification). Because decoders generally only navigate arrays, this does help out, although not as much as I'd hoped (not really much for speed, and allocation rates are only 8-10% lower in the decoding benchmarks).

There are no changes to the public API, but there is some extra complexity, so I'm not sure I want to merge this, but I wanted to post it here for possible discussion.


travisbrown:
Something like this ended up happening in #459.

PR #114, opened 2015-11-17
Add tests for cursor operations

travisbrown:
I'm aiming for 100% coverage for cursors, and just in the first few minutes discovered that `leftAt` was completely broken (it hangs forever when the predicate holds for some value), and that it's also broken in Argonaut.

If anyone really wants the `leftAt` fix backported to 0.2, I'll do it, but given that nobody seemed to use it even in Argonaut, I think it can wait for 0.3.0.


travisbrown:
I'm going to go ahead and merge this when it's green, since it fixes the `leftAt` bug, is incrementally better test-wise, and since I've got some bigger changes to the `Cursor` tests coming.

PR #115, opened 2015-11-17
Introduce AbstractDecoder

travisbrown:
This abstraction helps with two things I'm working on—it lets us capture the behavior that `Decoder` and `AccumulatingDecoder` share, and it lets us (cleanly) introduce a new `PerfectDecoder` type class that represents decoders that can't fail, which makes some changes I'm working on for the `*Cursor` tests a little nicer.

I've introduced a new `internal` package to signal that these aren't generally user-facing types. It's possible that `GenericCursor` should go here as well (and that it should be named `AbstractCursor` for consistency—I went with "abstract" here instead of "generic" in order to avoid confusion with the `generic` module).


travisbrown:
This PR is ancient and will be superseded at least in part by #228.

PR #117, opened 2015-11-18
Use the cats 0.4.0 snapshot (do not merge)

travisbrown:
I'm opening this PR so that any issues turn up early, but I don't think we'll end up merging it unless it looks like cats 0.4.0 is going to be out before Shapeless 2.3.0.


travisbrown:
Superseded by #146.

PR #120, opened 2015-11-23
Avoid some allocations in Option decoder

travisbrown:
This takes advantage of the changes in #118 to make `Option` decoders a little more efficient (and arguably clearer).

### DIFF
--- a/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -16,10 +16,8 @@ trait Decoder[A] extends Serializable { self =>
   /**
    * Decode the given acursor.
    */
-  def tryDecode(c: ACursor): Decoder.Result[A] = c.either.fold(
-    invalid =>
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/io/circe/Decoder.scala
```diff
@@ -389,17 +389,15 @@ object Decoder extends TupleDecoders with LowPriorityDecoders {
...
-            df.history.headOption.fold[Result[Option[A]]](
-              Xor.right(None)
-            )(_ => Xor.left(df)),
-          a => Xor.right(Some(a))
-        )
-      }
-    }
+    withReattempt(c =>
+      if (c.succeeded) {
+        if (c.any.focus.isNull) Xor.right(None) else d(c.any) match {
```
vkostyukov: Low hanging fruit (-8 bytes) is `Xor.right(None)` - I think we could lift this one into a `val` and reuse it whenever possible (pretty sure there are a lot of places).

travisbrown:
@vkostyukov Good point! I've added a `rightNone` and used it here (which I think is the only place we deal with `Xor.right(None)`).

PR #122, opened 2015-11-24
upgrade monocle and add plated instance

julien-truffaut:


travisbrown:
Looks great, thanks!

optics/src/main/scala/io/circe/optics/JsonObjectOptics.scala
```diff
@@ -22,7 +22,7 @@ trait JsonObjectOptics extends CatsConversions with ListInstances {
     }
   }
 
-  implicit lazy val objectAt: At[JsonObject, String, Json] = new At[JsonObject, String, Json]{
+  implicit lazy val objectAt: At[JsonObject, String, Option[Json]] = new At[JsonObject, String, Option[Json]]{
```
travisbrown: I think the Scalastyle line-length checker will complain here—it wants lines to be shorted than 100 characters.

optics/src/main/scala/io/circe/optics/JsonPath.scala
```diff
@@ -1,10 +1,12 @@
 package io.circe.optics
 
-import io.circe.{ Decoder, Encoder, Json }
-import io.circe.optics.JsonOptics._, io.circe.optics.JsonObjectOptics._
-import monocle.{ Optional, Prism }
-import monocle.function.index
+import io.circe.optics.JsonObjectOptics._
+import io.circe.optics.JsonOptics._
+import io.circe.{Decoder, Encoder, Json}
```
travisbrown: This is super nitpicky, but do you mind maintaining the alphabetization (where capital letters precede lowercase) and putting a space inside the `{ }`? (I could also do this in a commit to the `topic/optics` branch after merging).

travisbrown:
Hey @julien-truffaut, since all tests pass and only Scalastyle fails the builds, I'm going to go ahead and merge this into my branch and then add a new commit for the formatting. Thanks again.

optics/src/main/scala/io/circe/optics/JsonPath.scala
julien-truffaut: sorry I used intellij refactor import, I didn't notice they were a specific formatting

PR #123, opened 2015-11-24
Add convenience methods for decoders

travisbrown:
`before` and `after` here are kind of like `contramap` and `andThen` for kleisli arrows, but I didn't want to use those names since these aren't literally composition.

I've been meaning to add these helper methods for a while. The general motivation is to allow definitions like this (from @etaty on [Gitter](https://gitter.im/travisbrown/circe?at=5654a74492aa9746647b61e2)):

```scala
val df = DateTimeFormat.forPattern("YYYY-MM-dd'T'HH:mm:ssZ")
val decoderDateTime: Decoder[DateTime] = Decoder.instance { cursor =>
  cursor.as[String].flatMap { s =>
    Xor.catchOnly[IllegalArgumentException] {
      DateTime.parse(s, df)
    }.leftMap(e => DecodingFailure(e.getMessage, cursor.history))
  }
}
```

Like this:

```scala
val df = DateTimeFormat.forPattern("YYYY-MM-dd'T'HH:mm:ssZ")
val decoderDateTime: Decoder[DateTime] = Decoder[String].after(s =>
  Xor.catchOnly[IllegalArgumentException](DateTime.parse(s, df)).leftMap(_.getMessage)
)
```

### DIFF
--- a/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -112,6 +112,20 @@ trait Decoder[A] extends Serializable { self =>
       a <- this(a1)
       b <- x(a2)
     } yield (a, b)
+
+  /**
+   * Create a new decoder that performs some operation on the incoming JSON before decoding.
+   */
+  final def after(f: HCursor => ACursor): Decoder[A] = Decoder.instance(c => self.tryDecode(f(c)))
+
+  /**
+   * Create a new decoder that performs some operation on the result if this one succeeds.
+   *
+   * @param f a function returning either a value or an error message
+   */
+  final def emap[B](f: A => Xor[String, B]): Decoder[B] = Decoder.instance(c =>
+    self(c).flatMap(a => f(a).leftMap(message => DecodingFailure(message, c.history)))
+  )
 }
 
 /**
@@ -172,6 +186,15 @@ object Decoder extends TupleDecoders with LowPriorityDecoders {
 
     override def tryDecode(c: ACursor) = f(c)
   }
+
+  /**
+   * Construct an instance that always fails with the given error message.
+   *
+   * @group Utilities
+   */
+  final def failWith[A](message: String): Decoder[A] = instance(_ =>
+    Xor.left(DecodingFailure(message, Nil))
+  )
 
   /**
    * @group Decoding
--- a/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -116,7 +116,7 @@ trait Decoder[A] extends Serializable { self =>
   /**
    * Create a new decoder that performs some operation on the incoming JSON before decoding.
    */
-  final def after(f: HCursor => ACursor): Decoder[A] = Decoder.instance(c => self.tryDecode(f(c)))
+  final def prepare(f: HCursor => ACursor): Decoder[A] = Decoder.instance(c => self.tryDecode(f(c)))
 
   /**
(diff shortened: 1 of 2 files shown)


etaty:
:+1:

travisbrown:
Now I'm wondering if `before` and `after` read backwards. Anyone have an opinion one way or the other, or other ideas for names?

vkostyukov:
`after` reminds me scodec's [`emap`](https://github.com/scodec/scodec/blob/master/shared/src/main/scala/scodec/Decoder.scala#L61-L63) (e for embed?). Although, I don't have a good name for `before`.

mandubian:
`after` in my mind sounds like the other way :)
`andThen` is much better but as you said has a connotation...
`thenDecode`
`andNext`
...

travisbrown:
Okay, how about `after` and `emap` (instead of `before` and `after`)?

mandubian:
+1 it's better

core/shared/src/main/scala/io/circe/Decoder.scala
```diff
@@ -112,6 +112,20 @@ trait Decoder[A] extends Serializable { self =>
       a <- this(a1)
       b <- x(a2)
     } yield (a, b)
+
+  /**
+   * Create a new decoder that performs some operation on the incoming JSON before decoding.
+   */
+  final def after(f: HCursor => ACursor): Decoder[A] = Decoder.instance(c => self.tryDecode(f(c)))
+
```
etaty: did you renamed before in after?
I am confused

etaty:
You could use `prepare(f: HCursor => ACursor)`

core/shared/src/main/scala/io/circe/Decoder.scala
travisbrown: Yeah, now it's "decode _after_ processing the incoming JSON" and `emap` for the other `flatMap`-like thing.

travisbrown:
@etaty Hmm, I like `prepare` (or maybe `preprocess`), although it doesn't sounds like you're performing a transformation, where often you'll just be navigating.

travisbrown:
@etaty (Also sorry for the confusion—I tend to force push to PR branches to keep history clean, but this is a case where the change should have been a new commit.)

travisbrown:
Okay, unless anyone speaks up soon, `prepare` and `emap` are the winners.

core/shared/src/main/scala/io/circe/Decoder.scala
```diff
@@ -112,6 +112,20 @@ trait Decoder[A] extends Serializable { self =>
...
+   * Create a new decoder that performs some operation on the incoming JSON before decoding.
+   */
+  final def prepare(f: HCursor => ACursor): Decoder[A] = Decoder.instance(c => self.tryDecode(f(c)))
+
+  /**
+   * Create a new decoder that performs some operation on the result if this one succeeds.
+   *
+   * @param f a function returning either a value or an error message
+   */
+  final def emap[B](f: A => Xor[String, B]): Decoder[B] = Decoder.instance(c =>
```
vkostyukov: Do I understand correctly that the reason for having `Xor[String, B]` (not `Xor[DecodingFailure, B]`) is because we need a cursor instance to construct a `DecodingFailure `?

etaty: yes, you can still create a DecodingFailure without the cursor but then the cursor history would be empty

travisbrown: Right—we don't really want callers to be able to throw away history here (if they really need to, they can use the more powerful `flatMap`).

PR #112, opened 2015-11-16
new scalajs module

chandu0101:
PR for https://github.com/travisbrown/circe/issues/93

PR #126, opened 2015-11-25
New Scala.js module

travisbrown:
This includes @chandu0101's commits from #112 rebased and with some follow-up changes by me, including the following:

1. I've replaced `JS` with `Js` everywhere it appears in the API, following the convention used by e.g. cats for `Js` specifically and for other acronyms in circe.
2. I've removed the unsafe `convertJSAnyToJson` (now called `unsafeConvertAnyToJson`) from the public API.
3. I've renamed `parseJs` to `convertJsToJson`, since this isn't really a "parsing" operation. It also now returns an `Xor[Throwable, Json]` instead of `Xor[ParsingFailure, Json]` for the same reason.
4. I'm now using property-based testing in the new tests.

I don't think any of these changes undermine the basic goals of the new module—let me know what you think, @chandu0101.


chandu0101:
code is more concise and clear now thanks alot for changes :)

PR #127, opened 2015-11-25
Add non-compilation tests for bad instances

### LINKED ISSUE #57: Add non-compilation tests

travisbrown:
Now that [this Shapeless bug](https://github.com/milessabin/shapeless/issues/453) is fixed in the 2.3.0 snapshot, we can remove our workaround and add non-compilation tests. Fixes #57.

### DIFF
--- a/generic/shared/src/main/scala/io/circe/generic/auto.scala
+++ b/generic/shared/src/main/scala/io/circe/generic/auto.scala
@@ -13,17 +13,4 @@ import io.circe.generic.encoding.DerivedObjectEncoder
  * sealed trait hierarchies, etc.
  */
 @reexports[DerivedDecoder, DerivedObjectEncoder]
-object auto {
-  /**
-   * Blocks derivation until Shapeless #453 is fixed.
-   */
-  implicit def decodeObject0: Decoder[Object] = sys.error("No Decoder[Object]")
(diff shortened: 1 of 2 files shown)


fthomas:
Would it make sense to use the `illTyped` version that takes an additional string argument (which is a regex) that is matched against the compiler error that scalac issues for the first string? This ensures that the code is failing for the right reason and not because there is a syntax error or a missing import. I think this usage of `illTyped` is widely underused (even in shapeless).

travisbrown:
@fthomas I agree that that feature is underused, but I'm not sure it makes sense here, since the expected error is just "could not find implicit value for parameter…", which is the same thing we'd get with a missing import. I guess there could be a syntax error, but given that the code here is only a handful of characters, I'm not sure that's worth worrying about.

fthomas:
Ok, no problem. I guess I've developed a reflex when seeing one argument `illTyped`s :-)

PR #138, opened 2015-12-14
Topic/accumulating decoders

pvillega:
Hi, 

apologies for the delay. I'm aware this will break on Map, still have to fix that one. I have a couple of questions:

* On Map, I started an approach (see line 489 on Decoder) but the *spinAcc* method I implemented doesn't build. I'm sure there's an obvious reason but I got stuck :( Any hints are appreciated so I can progress on this :)

* I added a test case for case classes. I tried to be 'fancy' with ScalaCheck creating instances from an ADT to serialise/deserialise but I was getting *SerializationExceptions* when converting to Json, so I decided to revert to a simpler scenario to start with. I'll try that again for the next commit.

* Any feedback on code formatting (I fear IDEA may mess things sometimes) or areas to revisit will be very appreciated :)

TODO (next commits):

* add tests using cursor history, that'll be the first task
* I realised I may need to add tests for JsonObject and Array!
* Try to clean test code as there's quite a bit of duplication

Thanks!

### DIFF
--- a/core/shared/src/main/scala/io/circe/AccumulatingDecoder.scala
+++ b/core/shared/src/main/scala/io/circe/AccumulatingDecoder.scala
@@ -1,6 +1,5 @@
 package io.circe
 
-import cats.Applicative
 import cats.data.ValidatedNel
 
 sealed trait AccumulatingDecoder[A] extends Serializable { self =>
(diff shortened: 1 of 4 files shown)


travisbrown:
Thanks, @pvillega! My long flight just landed, and I've got a fix for the map decoder and a few other little changes (I've rebased, amended commits to undo non-semantically significant formatting changes, etc.). I'll open a new PR with these changes once I'm back at a computer.

pvillega:
ok great. I'll continue adding the tests for history cursor in the meantime so I can PR-them after your fixes

travisbrown:

PR #139, opened 2015-12-16
Accumulating decoders

travisbrown:
This is a rebased version of #138 that includes the commits from #85 with some small changes and additions.

Your commits are here, @pvillega, but I've removed the formatting-only changes and tweaked a few other things. For example, I've swapped the "good" and "bad" types for the tests, since e.g. `[ "10" ]` would be successfully decoded as a `List[Int]`, which means having the integers be the valid values would sometimes result in spurious failures.

I'd be happy to rebase any upcoming changes in #138 against this branch, @pvillega, or you can do that—just let me know. Thanks again!

### DIFF
--- a/core/shared/src/main/scala/io/circe/AccumulatingDecoder.scala
+++ b/core/shared/src/main/scala/io/circe/AccumulatingDecoder.scala
@@ -1,0 +1,34 @@
+package io.circe
+
+import cats.Applicative
+import cats.data.ValidatedNel
+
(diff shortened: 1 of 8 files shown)


pvillega:
ok will use this branch for next steps thanks!

travisbrown:
@pvillega Great! I just pushed a merge commit with a couple of recent changes (#136 and #140) that should get the tests working again—something about an update to Travis CI's legacy infrastructure was causing them to fail on some builds.

PR #141, opened 2015-12-16
Acc dec rebased

pvillega:
As mentioned in Gitter, this contains tests to ensure history provided by accumulated errors replay correctly.

Unfortunately there's that issue with Map/Set tests, which fail for no obvious reason. I'm lost as to how to fix, so as agreed I flagged them as 'ignore' test. Going on holidays, will check again when I'm back to see if I can help somehow :)


travisbrown:

PR #147, opened 2015-12-21
Fix error in refined ScalaDoc example

fthomas:
`refineV` is a function that returns an `Either[String, Int Refined Positive]` and `refineMV` is a macro that returns an `Int Refined Positive` or fails to compile.


travisbrown:
Thanks for catching this, @fthomas!

PR #149, opened 2015-12-21
Use scalastyle's rule for imports order

vkostyukov:
Since 0.8 Scalastyle supports the rule that checks if imports are ordered according to the predefined order. I tried to pick the most simple order (everything is in one group and sorted alphabetically) for Circe so it matches the previous style.

If you think we should enable some mo complicated rule (i.e., two groups for stdlib imports and everything else as per [Finch's config](https://github.com/finagle/finch/blob/master/scalastyle-config.xml#L79-L85)) - let me know, I'm happy to change it.


travisbrown:
This is great! I think I'd prefer `'A' < 'a'`, but I can live with this version.

I don't prefer separating standard library imports, but I'm not sure I have a good reason for that.

travisbrown:
Just restarted the build—looks like a Sonatype server issue.

PR #153, opened 2015-12-22
Some edits to contributing doc

travisbrown:


vkostyukov:
Thanks @travisbrown! This is looks awesome - sorry for not mention all of these in the very begining.

PR #161, opened 2016-01-09
Make almost all implementations final

travisbrown:
This is somewhat tedious and verbose, but it keeps us honest, and based on my experience with iteratee.io it may actually have a positive impact on performance.


alexarchambault:
@travisbrown You seem to have added `final` modifiers on members of singletons, or of anonymous classes. Unless I'm mistaken, these are already final by default.

travisbrown:
@alexarchambault It's probably unnecessary, but in both cases adding `final` results in different byte code, and in the case of objects scalac does have a weird `-Yoverride-objects` option.

PR #163, opened 2016-01-11
Configuring generic derivation (demonstration; do not merge)

travisbrown:
This morning @notxcain [asked about the status](https://gitter.im/travisbrown/circe?at=56939d6a0712a5b63b4cb612) of configuration for generic derivation, and I wanted to go ahead and post this branch since I think it's the most promising approach I've tried so far (although it's a long way from done). Usage looks like this:

```scala
import io.circe._, io.circe.generic.auto._, io.circe.jawn._, io.circe.syntax._
import io.circe.generic.config.SnakeCaseKeys

case class Qux(thisIsAFieldWithALongName: Int, anotherField: String)
```

And then:

```scala
scala> val json = Qux(13, "abcd").asJson
json: io.circe.Json =
{
  "thisIsAFieldWithALongName" : 13,
  "anotherField" : "abcd"
}

scala> val snakeCaseJson = Qux(13, "abcd").asJsonConfigured[SnakeCaseKeys]
snakeCaseJson: io.circe.Json =
{
  "this_is_a_field_with_a_long_name" : 13,
  "another_field" : "abcd"
}

scala> decode[Qux](json.noSpaces)
res0: cats.data.Xor[io.circe.Error,Qux] = Right(Qux(13,abcd))

scala> decodeConfigured[Qux, SnakeCaseKeys](snakeCaseJson.noSpaces)
res1: cats.data.Xor[io.circe.Error,Qux] = Right(Qux(13,abcd))
```

The code above actually works and all tests pass unchanged, but there are a few problems:

1. Composing configuration: ideally something like `decodeConfigured[Qux, SnakeCaseKeys with FieldDiscriminators](...)` should work, but I've run into problems trying to get there.
2. Nested derived instances: getting the prioritization right (not stepping on user-defined instances but passing configuration into nested case classes) for this is a mess, and it's currently broken (I had it working at some point at the expense of some trade-off that I don't remember at the moment).
3. It makes compilation even slower, even for non-configured instances. I don't want to merge any version of this that hurts the compilation speed for non-configured instances.

### DIFF
--- a/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -10,6 +10,8 @@ import scala.collection.generic.CanBuildFrom
 import scala.collection.mutable
 
 trait Decoder[A] extends Serializable { self =>
+  type Config
+
   /**
    * Decode the given hcursor.
    */
@@ -608,3 +610,9 @@ final object Decoder extends TupleDecoders with LowPriorityDecoders {
 }
 
 @export.imports[Decoder] private[circe] trait LowPriorityDecoders
+
+trait ConfiguredDecoder[C, A] extends Decoder[A] {
+  type Config = C
+}
+
+@export.imports[ConfiguredDecoder] object ConfiguredDecoder
--- a/core/shared/src/main/scala/io/circe/Encoder.scala
+++ b/core/shared/src/main/scala/io/circe/Encoder.scala
@@ -14,6 +14,8 @@ import scala.collection.mutable.ArrayBuffer
  * @author Travis Brown
  */
 trait Encoder[A] extends Serializable {
+  type Config
+
   /**
    * Converts a value to JSON.
    */
@@ -53,8 +55,6 @@ trait Encoder[A] extends Serializable {
  * @author Travis Brown
  */
 object Encoder extends TupleEncoders with LowPriorityEncoders {
-  type Config
-
   /**
    * Return an instance for a given type `A`.
    *
--- a/core/shared/src/main/scala/io/circe/Encoder.scala
+++ b/core/shared/src/main/scala/io/circe/Encoder.scala
@@ -53,6 +53,8 @@ trait Encoder[A] extends Serializable {
  * @author Travis Brown
  */
 object Encoder extends TupleEncoders with LowPriorityEncoders {
+  type Config
+
   /**
    * Return an instance for a given type `A`.
    *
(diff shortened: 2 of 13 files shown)


alexarchambault:
Doesn't this go against point 4 of the [design document](https://github.com/travisbrown/circe/blob/master/DESIGN.md#guidelines)? :-P

notxcain:
Oh! This is just great. It just uses implicits more explicitly :) 
1. I think `decodeConfigured[Qux, SnakeCaseKeys :: FieldDiscriminators :: HNil](...)`  should work better as you can iterate over HList and compose encoders.
2. Will [export-hook](https://github.com/milessabin/export-hook) do the trick?
3. Not sure what to do here. Gonna check it tomorrow morning.

notxcain:
@alexarchambault well, it doesn't use implicits until it asked to )

travisbrown:
@alexarchambault Not really—the `asJsonConfigured` and `decodeConfigured` methods ask for instances of `ConfiguredDecoder` and `ConfiguredEncoded` type classes that extend the basic ones but add a configuration type parameter. There's no implicit configuration floating around—you always have to ask for a `Decoder[Qux]`, a `ConfiguredDecoder[SnakeCaseKeys, Qux]`, etc.

@notxcain I'll try using an `HList` for the configuration—I'm not sure it'll help, but there's a chance (the problem is avoiding an exponential blow-up of instance methods as you add configuration options). About export-hook—this approach already relies pretty heavily on it, but doesn't make use of its facilities for fine-grained prioritization, so you're right—it's something to try.

julienrf:
@travisbrown What would you think of a design like the following?

~~~ scala
class Derivation(namingConvention: NamingConvention, etc.) {
  def deriveDecoder[A](implicit decode: Lazy[DerivedDecoder[A]]): DerivedDecoder[A] =
    decode.value(namingConvention)
}
~~~

And then:

~~~ scala
val customGeneric = new Derivation(SnakeCaseKeys)

customGeneric.deriveDecoder[Qux]
~~~

And also:

~~~ scala
object semiauto extends Derivation(CamelCaseKeys)
~~~

travisbrown:
@julienrf Hmm, that might be a reasonable alternative if this general approach doesn't work out.

One thing I'm aiming for (that I don't think your proposal accommodates) is the ability to mix user-defined and derived instances as seamlessly as possible. For example, suppose I've got a big nest of case classes and ADTs, and I want to use derivation for all of it except one case class buried somewhere inside. I want to be able to configure the whole thing uniformly, so that I can generate one batch of JSON with the default object wrapper discriminators and another with field discriminators. With the current approach I just define custom `Decoder[Foo]` and `ConfiguredDecoder[FieldDiscriminator, Foo]` instances, and the appropriate one will get picked up during derivation.

I'm not sure that'll be reasonably possible, though. We might also be able to make that kind of thing work with what you're proposing. Thanks for the idea—

travisbrown:
Superseded by #164.

PR #168, opened 2016-01-15
Add beginning of java8 module

travisbrown:
Right now this only includes instances for `LocalDateTime`:

```scala
scala> import io.circe._, io.circe.java8._, java.time.LocalDateTime
import io.circe._
import io.circe.java8._
import java.time.LocalDateTime

scala> val json = Encoder[LocalDateTime].apply(LocalDateTime.now)
json: io.circe.Json = "2016-01-15T13:27:46.64"

scala> Decoder[LocalDateTime].apply(json.hcursor)
res0: io.circe.Decoder.Result[java.time.LocalDateTime] = Right(2016-01-15T13:27:46.640)
```

Note that this module won't work on Java 7 or earlier, and isn't available for Scala.js.

I'd also like to add some other things like helpers for creating encoder and decoder instances from Java with lambda expressions, but I'm happy to merge this as is for now.

I'm not 100% sure the Travis CI config will work correctly (I hate Bash), and it's taking jobs forever to get started today, so I may be force pushing fixes to this branch until I get it working.

### DIFF
--- a/java8/src/main/scala/io/circe/java8/package.scala
+++ b/java8/src/main/scala/io/circe/java8/package.scala
@@ -1,0 +1,25 @@
+package io.circe
+
+import cats.data.Xor
+import java.time.LocalDateTime
+import java.time.format.{ DateTimeFormatter, DateTimeParseException }
+
+package object java8 {
+  final def decodeLocalDateTime(formatter: DateTimeFormatter): Decoder[LocalDateTime] =
+    Decoder.instance { c =>
+      c.as[String].flatMap { s =>
+        try Xor.right(LocalDateTime.parse(s, formatter)) catch {
+          case _: DateTimeParseException => Xor.left(DecodingFailure("LocalDateTime", c.history))
+        }
+      }
+    }
+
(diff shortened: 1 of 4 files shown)


java8/src/main/scala/io/circe/java8/package.scala
```diff
@@ -0,0 +1,25 @@
...
+        try Xor.right(LocalDateTime.parse(s, formatter)) catch {
+          case _: DateTimeParseException => Xor.left(DecodingFailure("LocalDateTime", c.history))
+        }
+      }
+    }
+
+  final def encodeLocalDateTime(formatter: DateTimeFormatter): Encoder[LocalDateTime] =
+    Encoder.instance(time => Json.string(time.format(formatter)))
+
+  implicit final def decodeLocalDateTimeDefault: Decoder[LocalDateTime] =
```
hseeberger: Make these default vals in order to not construct them on every call?

travisbrown: Ah, right—good call.

PR #172, opened 2016-01-25
Deep merge operation

anakos:
Here's my initial attempt at adding the deepmerge method to the Json class. While almost identical to the corresponding method in Argonaut.Json, as per the discussion [here] (https://gitter.im/travisbrown/circe?at=56915002ee13050b38a293b3) and [here](https://github.com/travisbrown/circe/issues/169), this implementation ensures that the fields in the Json argument consistently take precedence.

### DIFF
--- a/core/shared/src/main/scala/io/circe/Json.scala
+++ b/core/shared/src/main/scala/io/circe/Json.scala
@@ -124,6 +124,29 @@ sealed abstract class Json extends Product with Serializable {
   final def spaces4: String = Printer.spaces4.pretty(this)
 
   /**
+    * Perform a deep merge of this JSON value with another JSON value.
+    *
+    * Objects are merged by key, values from the argument JSON take
+    * precedence over values from this JSON. Nested objects are
+    * recursed.
+    *
+    * Null, Array, Boolean, String and Number are treated as values,
+    * and values from the argument JSON completely replace values
+    * from this JSON.
+    */
+  def deepMerge(that: Json): Json =
+    (asObject, that.asObject) match {
+      case (Some(lhs), Some(rhs)) =>
+        fromJsonObject(
+          lhs.toList.foldLeft(rhs) {
+            case (acc, (key, value)) =>
+              rhs(key).fold(acc.add(key, value)) { r => acc.add(key, value.deepMerge(r)) }
+          }
+        )
+      case _ => that
+    }
+
+  /**
    * Compute a `String` representation for this JSON value.
    */
   override final def toString: String = spaces2
--- a/tests/shared/src/test/scala/io/circe/JsonSuite.scala
+++ b/tests/shared/src/test/scala/io/circe/JsonSuite.scala
@@ -1,0 +1,20 @@
+package io.circe
+
+import io.circe.tests.CirceSuite
+
+class JsonSuite extends CirceSuite {
+
+  test("deepmerge preserves argument order") {
+    check { (js: List[Json]) =>
+      val fields = js.zipWithIndex.map {
+        case (j, i) => i.toString -> j
+      }
+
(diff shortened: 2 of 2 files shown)


travisbrown:
Looks good to me, although Scalastyle doesn't like that `import cats.{ Show, Apply }` isn't in alphabetical order. I've got a few minor comments but I can address them in a follow-up commit, so if you don't mind pushing a fix for the import order to keep Travis CI happy, we can go ahead and merge this.

core/shared/src/main/scala/io/circe/Json.scala
```diff
@@ -1,9 +1,10 @@
 package io.circe
 
 import algebra.Eq
-import cats.Show
+import cats.{ Show, Apply }
```
travisbrown: Scalastyle is complaining that these aren't in alphabetical order.

core/shared/src/main/scala/io/circe/Json.scala
```diff
@@ -124,6 +125,27 @@ sealed abstract class Json extends Product with Serializable {
...
+    * Objects are merged by key, values from the argument JSON take
+    * precedence over values from this JSON. Nested objects are
+    * recursed.
+    *
+    * Null, Array, Boolean, String and Number are treated as values,
+    * and values from the argument JSON completely replace values
+    * from this JSON.
+    */
+  def deepmerge(that: Json): Json =
+    Apply[Option].map2(asObject, that.asObject) { (lhs, rhs) =>
```
travisbrown: I'd personally prefer using either pattern matching or a `for`-comprehension (or just `flatMap`, etc.) here, for a couple of reasons.

The first reason is that this version is probably less clear for most Scala developers. The fact that Cats provides an `Apply` instance for `Option` is great when you need to abstract over type constructors, but here we're just working with plain old `Option`, so I'm not sure `map2` is worth the readability cost.

The second reason is that using the `Apply` instance has additional costs at both compile time and runtime. This is unlikely ever to be relevant in this case specifically, but if we did this everywhere in the codebase it might start to be a problem, and I'd like to aim for consistency. Again, if we needed the abstraction, it'd totally be worth the performance cost in this context, but it doesn't really buy us anything here.

tests/shared/src/test/scala/io/circe/JsonSuite.scala
```diff
@@ -0,0 +1,21 @@
+package io.circe
+
+import io.circe.tests.CirceSuite
+import org.scalacheck.{ Arbitrary, Gen }
```
travisbrown: Unused imports.

core/shared/src/main/scala/io/circe/Json.scala
```diff
@@ -124,6 +125,27 @@ sealed abstract class Json extends Product with Serializable {
...
+    *
+    * Objects are merged by key, values from the argument JSON take
+    * precedence over values from this JSON. Nested objects are
+    * recursed.
+    *
+    * Null, Array, Boolean, String and Number are treated as values,
+    * and values from the argument JSON completely replace values
+    * from this JSON.
+    */
+  def deepmerge(that: Json): Json =
```
travisbrown: This is super minor, but I think I'd like to go with camel-case `deepMerge` for the sake of consistency.

travisbrown:
Okay, I've just made a few comments, but feel free to leave them for me to follow up on. Thanks again, @anakos!

core/shared/src/main/scala/io/circe/Json.scala
anakos: No problem, I got it - will revert to pattern matching and push the change within the next hour or two.

travisbrown:
Perfect, thanks @anakos!

PR #176, opened 2016-01-29
Renaming of cats.syntax.monoidal to cats.syntax.cartesian

anakos:
cats.syntax.monoidal has been renamed to cats.syntax.cartesian - the modifications here simply reflect that change.


travisbrown:
Thanks @anakos! Mind changing the import order in `Decoder.scala` to satisfy Scalastyle? You can confirm locally with `sbt scalastyle`.

anakos:
Sorry forgot about that again! No problem, will get that sorted as soon as I get out of work.

Issue #170, opened 2016-01-22
Provide instances for OneAnd


travisbrown:
From @mpilquist on [Gitter](https://gitter.im/travisbrown/circe?at=56a24a39dc33b33c754747e2). We currently provide encoders and decoders for `NonEmptyList`, but these instances could be for any `OneAnd` where there's an instance for the collection type.

ArtemPyanykh:
I wonder what you mean by "_where there's an instance for the collection type_"? 

As far as I understand even if we have an `Encoder` for `F[A]` in case of `OneAnd` we need to be able to put an additional value `head: A` inside a context `tail: F[A]` before applying the encoder. If we're speaking only about `NonEmptyList/Vector/Stream` it might be enough to have:
```
def encodeOneAnd[A: Encoder, F[_]: Foldable](x: OneAnd[F, A]): Encoder[OneAnd[F, A]]
```

For decoder this won't work though. Basically we can always decode `JArray` into `List`. So to implement `Decoder[OneAnd[F, A]]` we need to transform `List[A]` into `F[A]`. It seems to me that the minimum needed set of capabilities on `F` is having `empty`, `pure` and `combine`. `MonadCombine` has all these and defined for `List`, `Vector` and `Stream` which gives out of box support for all `NonEmpty*` from `cats`.

Another things is this reasoning is applicable only to _collection-like_ containers, and decoding from/into `JArray`. It's not clear what should be a generic representation for, say, `OneAnd[Id, A]` which is isomorphic to `(A, A)`.

Or am I thinking in the wrong direction?

travisbrown:
@ArtemPyanykh You're right that there are a few ways this could be done, and I think it's fine to only target collection-like `OneAnd`s.

One other approach would be to require only instances for both `A` and `F[A]` for the `OneAnd[F, A]` codecs. For the encoder you apply the `F[A]` encoder to the tail and then prepend the encoded head _if_ the `F[A]` encoder returned a `JArray` (and fail otherwise). For the decoder you first check whether the cursor points to a non-empty `JArray`. If it doesn't, you fail, but if it does you decode the first `Json` with the `A` decoder and the rest with the `F[A]` decoder.

This approach takes advantage of the optimized `CanBuildFrom` decoder for all the `NonEmpty*` types in cats.

What do you think?

ArtemPyanykh:
[...]

PR #178, opened 2016-01-30
Implement OneAnd[F, A] instances for collection-like F

ArtemPyanykh:
Addresses #170 

Some reasoning behind changes:

1. I'm targeting only collection-like `F`s as it's not perfectly clear to me what should be a generic representation of `OneAnd[F, A]` where `F` is just some arbitrary type constructor. Anyway all the fun starts only when `F` is at least `Foldable`.
2. I decided to target collections from `std` lib, instead of implementing instances in terms of `cats` typeclasses for a couple reasons:
    1. Getting instances for `NonEmptyList/Vector/Stream` will require end-users to import `cats.std.list/vector/stream._` which is not convenient. Targeting `IsTraversableOnce` on the other hand will automagically generate instances for standard collections, which is a major use-case AFAIU.
    2. Just replacing an instance for `NonEmptyList[A]` with an instance for `OneAnd[F, A]` can break code in a subtle way if automatic derivation is involved. Consider the following snippet:

```scala
import io.circe._

val nel: NonEmptyList[Int] = NonEmptyList(1, 2)

// Before: encoder for NEL is implemented directly and doesn't require any additional imports
val e = Encoder[NonEmptyList[Int]]
e(nel) // "[1, 2]"

// After: encoder for NEL is implemented in terms of OneAnd and requires import from cats
val e = Encoder[NonEmptyList[Int]] // generic derivation kicks in because no foldable for list is provided
e(nel) // "{head: 1, tail: [2]}"
```

Might not be very generic. But on the bright-side we get very good performance comparable to that of regular collections. Locally I got ~3% difference:
```
[info] Benchmark                          Mode  Cnt     Score    Error  Units
[info] EncodingBenchmark.encodeFooNelsC  thrpt   20  4875.958 ± 74.492  ops/s
[info] EncodingBenchmark.encodeFoosC     thrpt   20  5039.914 ± 72.708  ops/s
```

### DIFF
--- a/benchmark/src/main/scala/io/circe/benchmark/Benchmark.scala
+++ b/benchmark/src/main/scala/io/circe/benchmark/Benchmark.scala
@@ -2,6 +2,7 @@ package io.circe.benchmark
 
 import algebra.Eq
 import argonaut.{ Json => JsonA, _ }, argonaut.Argonaut._
+import cats.data.NonEmptyList
 import io.circe.{ Decoder, Encoder, Json => JsonC }
 import io.circe.generic.semiauto._
 import io.circe.jawn._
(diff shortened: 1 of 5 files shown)


kevinmeredith:
Note - this a comment for my own learning, not a review comment.

Per your [comment](https://github.com/travisbrown/circe/issues/170#issuecomment-174019224) in #170, I thought that you needed `MonadCombine`? Searching for it in this page (in my browser), I don't see it. Why not?

ArtemPyanykh:
@kevinmeredith yep, no `MonadCombine` here. 

Basically you can implement `Decoder` for `OneAnd` in terms of `MonadCombine`, but the consequences are:
1. library end-users will have to import `cats.std.list/vector/stream._` to to get `MonadCombine` instance in scope for standard library collections,
2. just replacing current decoder for `NonEmptyList` with the one based on `MonadCombine` will break existing code (again, because of the required extra import).
3. there will probably be some performance penalties, however I would not like to speculate on this wo/ any benchmarks.

`CanBuildFrom[Nothing, A, List[A]]` on the other hand is kind of analogous to `MonadCombine` in a sense that it can give us an empty collection of a given type and append new elements to the collection. But in addition it's also automagically available in the implicit scope for all standard collections, which is quite handy.

Of course, we can have `CanBuildFrom` based implementation for std-lib collections and also `MonadCombine` based implementation for other stuff, but that would require rearranging priority of implicits to avoid ambiguity in resolution, and I'm not sure whether this is really needed atm.

Does it make sense?

travisbrown:
Thanks, @ArtemPyanykh—this looks perfect to me, with the one exception that `Encoder.fromFoldable` is part of the public API and potentially has uses other than the internal one that's going away, so I think we should keep it around.

ArtemPyanykh:
@travisbrown right, was a bit too eager. I put it back.

ArtemPyanykh:
@travisbrown is there anything else missing here?

travisbrown:
@ArtemPyanykh Nope, this looks good, thanks—I've just been swamped with Cats release stuff. Merging now.

ArtemPyanykh:
@travisbrown cool! Noticed interesting activity in Cats -- gotta check out recent changes.

PR #180, opened 2016-02-01
Fixed error message in the unit decoder.

alexknvl:


travisbrown:
Oh, good catch! Will merge shortly.

Issue #38, opened 2015-08-19
Codecs for maps with non-string keys


travisbrown:
@koshelev asked today [on gitter](https://gitter.im/travisbrown/circe?at=55d4d70a11c7afc249785e77) about `Decoder` instances for a `Map` with non-string keys.

It'd be nice to have a clear story for these cases. This new [Argonaut pull request](https://github.com/argonaut-io/argonaut/pull/189/files) by @xuwei-k is one solution: you have to provide a mapping from keys to strings, and then you get the instance for free. Alternatively uPickle provides `Map[NonString, Bar]` instances that serialize to JSON arrays of pairs.

I'm leaning toward not providing instances for maps with non-string keys by default (for the reason given [here](https://gitter.im/travisbrown/circe?at=55d4ee971ba62e4a21fa72b2)), but making it easy for users to define uPickle-style instances if they want them.

koshelev:
I like the first approach, with mapping keys to string. It could be very handy in case where a key is a scala value class, wrapping strings.

ngbinh:
Yes, 1st approach sounds nice and easy! :)

travisbrown:
Okay, sounds like people want this, so if someone wants to port the Argonaut implementation (with appropriate attribution), I'd be happy to merge it.

lukewyman:
@travisbrown, I'd like to take a crack at this.

travisbrown:
@lukewyman Oh, this is now done—I wrapped up some work that @pjan started today. Thanks, and sorry for the confusion—I should have closed the issue this morning.

PR #186, opened 2016-02-02
adds a KeyEncoder type class to encode a custom value as Json key

pjan:
Initial shot at addressing #38 

Not sure about how to do the attribution correctly,
nor as to where (and how) you would like to see it be tested (don't seem to fine a suite that is testing the encoder instances).

Can iterate over this according to feedback/answers.


travisbrown:
Thanks for getting this started, @pjan. I think a note and link in the Scaladoc would work for attribution, since we credit Argonaut in the top-level `NOTICE` file now.

About testing: round-trip tests for encoders come from the `CodecTest` laws, and most of the tests for specific encoders are in `CodecSuites.scala`. This would be a rare case where we don't provide matching decoders (except when the key is `String`), but you could define a custom key type like `case class MyKey(s: String)`, give it a `KeyEncoder` instance, and then write a custom decoder for `Map[MyKey, A: Decoder]` (using a `map` on the ordinary `Map` decoder).

One question: is there a reason to keep the `Map[String, A: Encoder]` instance and this one, instead of providing a `KeyEncoder` instance for strings (which is what Argonaut does)?

travisbrown:
Another thought: we could use `cats.Show` instead of `KeyEncoder`, which lets us avoid including a fairly meaningless type class in our API, and lets users get instances for standard types a little more easily.

pjan:
Great idea on the use of `cats.Show` instead of defining another typeclass.

One question: if I were to remove the Encoder instance for `Map[String, V]` in favour of the more generic instance for `Map[K, V]` where the instance observes an implicit `Show[String]`, where am I to put that implicit so it doesn't require everybody to import it/put it in scope?

pjan:
@travisbrown shot at using `Show` for this. The `SerializableSuite` currently fails as the compiler can't resolve an implicit instance of `Encoder[Map[String, Int]]` in scope, now that there's no concrete instance of it.

At the same time, I'm not really sure whether bringing `Show[String]` implicitly in scope is what I'd want to do.

Thoughts?

travisbrown:
@pjan Sorry about the long delay in review (I've been traveling). I agree that we probably don't want to put a `Show[String]` instance into scope for people, so we could either go back to a custom type class or treat `Map[String, V]` separately. I'll be home tomorrow and will think about this in the meantime—

travisbrown:
These commits were included in #213, which is now merged.

Issue #187, opened 2016-02-02
Ugh signed zeros


travisbrown:
Let me start by saying that I spent my entire life up until about an hour ago without realizing what a shitshow signed zeros are, and I wish I could go back to a state of innocence, so you might want to close the tab and move on.

First for the basics:

```scala
scala> 0.0 == -0.0
res0: Boolean = true

scala> Double.box(0.0) == Double.box(-0.0)
res1: Boolean = false

scala> java.lang.Double.compare(0.0, -0.0) == 0
res2: Boolean = false
```

Hurray, that makes no sense (but at least you might have seen it before).

Cats and Scalaz also disagree:

```scala
scala> import scalaz.Equal, scalaz.std.anyVal._
import scalaz.Equal
import scalaz.std.anyVal._

scala> Equal[Double].equal(0.0, -0.0)
res3: Boolean = true

scala> import cats.Eq, cats.std.double._
import cats.Eq
import cats.std.double._

scala> Eq[Double].eqv(0.0, -0.0)
res4: Boolean = false
```

Apparently IEEE 754 says something like "positive and negative zeros are distinct but equal", and that's why Java's primitive floating point types make them equal. Cats (and Algebra, although each defines its own instance) seems to be taking a more principled stand than Scalaz here on "distinct but equal" not making any sense (I don't know what the actual motivations for the choices here are, but I'm curious).

Argonaut says all zero JSON numbers are equal:

```scala
scala> import argonaut._, Argonaut._, scalaz._, Scalaz._
import argonaut._
import Argonaut._
import scalaz._
import Scalaz._

scala> Parse.parse("0.0") === Parse.parse("-0.0")
res0: Boolean = true

scala> Parse.parse("0") === Parse.parse("-0")
res1: Boolean = true
```

But it's perfectly happy to round-trip negative zeros if they have a fractional part:

```scala
scala> Parse.parse("-0").map(_.nospaces)
res2: scalaz.\/[String,String] = \/-(0)

scala> Parse.parse("-0.0").map(_.nospaces)
res3: scalaz.\/[String,String] = \/-(-0.0)
```

This seems really wrong to me—if two things serialize differently, the `scalaz.Equal` instance shouldn't say they're the same.

circe currently round-trips both cases (at least with Jawn as the parser—other parsers may behave in such a way that this isn't possible), but also says they're all equal:

```scala
scala> import io.circe._, io.circe.jawn._, cats.syntax.eq._
import io.circe._
import io.circe.jawn._
[...]

Issue #173, opened 2016-01-25
Try to convert string value in Double/Float decoders


knutwalker:
From the [Gitter conversation](https://gitter.im/travisbrown/circe?at=56a69f40c54bc2bf180bfcd6).

The Decoders for `Int`, `Long` and other integral types try to convert a string into their number type.

```scala
import io.circe._, parse._, cats.data.Xor
scala> val Xor.Right(i) = decode[Int]("\"42\"")
i: Int = 42
```

The decoder for `Double` and `Float` don't try this conversion because of this:

```
scala> "Infinity".toDouble
res1: Double = Infinity
```

However, we could use `JsonNumber.fromString` and `_.toDouble` to try to safely convert a String into a Double, for example:

```scala
scala> val doubleWithFromString = Decoder.instance(c => c.as[String].flatMap(s => Xor.fromOption(JsonNumber.fromString(s).map(_.toDouble), DecodingFailure("Double", c.history))))
doubleWithFromString: io.circe.Decoder[Double] = io.circe.Decoder$$anon$8@47016c8b

scala> val Xor.Right(d) = decode("\"13.37\"")(doubleWithFromString)
d: Double = 13.37

scala> val Xor.Left(e) = decode("\"Infinity\"")(doubleWithFromString)
e: io.circe.Error = io.circe.DecodingFailure: Double

scala> val Xor.Left(e) = decode("\"NaN\"")(doubleWithFromString)
e: io.circe.Error = io.circe.DecodingFailure: Double
```

If `"Infinty"` and `"NaN"` are valid to be parsed as Double, one can fallback to using a `Decoder[String].map(_.toDouble)` approach.

MartinSeeler:
I think this is a good idea!

travisbrown:
Fixed in #189.

Issue #124, opened 2015-11-24
Conversion from JsonNumber to BigDecimal can throw exceptions


travisbrown:
JSON allows numbers with exponents larger than `Int.MaxValue`, and `JsonNumber.fromString` will happily accept them:

```scala
val Some(okay) = JsonNumber.fromString(s"1E${ Int.MaxValue.toLong }")
val Some(tooBig) = JsonNumber.fromString(s"1E${ Int.MaxValue.toLong + 1L }")
```

Unfortunately the `toBigDecimal` method just crashes on `tooBig` here:

```scala
scala> okay.toBigDecimal
res47: BigDecimal = 1E+2147483647

scala> tooBig.toBigDecimal
java.lang.NumberFormatException
  at java.math.BigDecimal.<init>(BigDecimal.java:491)
  at java.math.BigDecimal.<init>(BigDecimal.java:824)
  at scala.math.BigDecimal$.apply(BigDecimal.scala:289)
  at io.circe.JsonDecimal.toBigDecimal$lzycompute(JsonNumber.scala:187)
  at io.circe.JsonDecimal.toBigDecimal(JsonNumber.scala:187)
  ... 43 elided
```

`toDouble` also crashes, as does `toLong`, etc. (even though `toLong` returns an `Option`). The same thing happens with these methods in Argonaut.

This is bad, and needs to be fixed. Here's my proposal:

1. `toBigDecimal` should return an `Option` that's empty if the value can't be parsed as a `BigDecimal`.
2. We add a new `approximateBigDecimal` that uses one of the various `pow(BigDecimal, BigDecimal)` implementations floating around (e.g. a copy-paste-and-port job from [here](http://arxiv.org/src/0908.3030v2/anc) or some other implementation with a friendly license).
3. The current contract of `toDouble` says that values outside the range of `Double` will be rounded to positive or negative infinity, so in the case that `toBigDecimal` is `None`, we use `approximateBigDecimal` and truncate.
4. The behavior of `toLong`, `toInt`, etc. is unchanged, except that they no longer throw exceptions.
5. All the `truncateTo` methods use `approximateBigDecimal` if necessary.

etaty:
I think your range is too large

[...]

PR #189, opened 2016-02-04
New BiggerDecimal and overhaul of JsonNumber

### LINKED ISSUE #83: Reconsider integral decoders
circe currently follows Argonaut in truncating, etc. when decoding integral types:

```scala
scala> io.circe.parse.decode[Short]("0.1")
res0: cats.data.Xor[io.circe.Error,Short] = Right(0)

scala> io.circe.parse.decode[Short]("123456")
res1: cats.data.Xor[io.circe.Error,Short] = Right(32767)
```

I don't really like this, and would prefer both of the above to result in failures. Any objections?

travisbrown:
This is still kind of messy but it fixes #83, #124, #173, and #187, and a lot of other probably or definitely broken things I've come across in the past couple of days that I didn't get around to reporting.

The key idea is a new `BiggerDecimal` class that's like `java.math.BigDecimal` except the scale is a `BigInteger` and it doesn't provide any operations except safe conversions to other numeric types (also it distinguishes between positive and negative zeros).

## Highlights

All of the following issues are inherited from Argonaut and at least some of them probably ought to be fixed there as well at some point.

The following fails immediately instead of running forever (see argonaut-io/argonaut#211):

```scala
io.circe.jawn.decode[BigInt]("1e2147483647")
````

The following doesn't crash at runtime (#124):

```scala
io.circe.jawn.decode[BigDecimal]("1e2147483648").map(_.toDouble)
```

The following fails instead of returning `32767` (#83):

```scala
io.circe.jawn.decode[Short]("32768")
```

The following fails instead of returning `10` (#83):

```scala
io.circe.jawn.decode[Int]("10.999")
```

Also `Decoder[Double]` now follows the integral decoders in attempting to parse JSON strings (#173) and `"-0.0"` and `"0.0"` are now different JSON documents.

All of this behavior is now covered by tests.

## Stuff that still needs fixing

Scala.js tests are failing because of [an apparent `BigInt` bug](https://twitter.com/travisbrown/status/695067947258220544)—I'll probably just make those tests JVM-only until the bug gets fixed.

Update: even after making the `NumberParsing` tests JVM-only, Scala.js fails because of missing `BigInteger` methods:

```scala
[error] Referring to non-existent method java.math.BigInteger.longValueExact()scala.Long
[error]   called from io.circe.util.SigAndExp.toLong()scala.Option
[error]   called from io.circe.JsonBigDecimal.toLong()scala.Option
```

We should probably just have platform-specific `BiggerDecimal` implementations, but I'm not looking forward to this.

## Other issues

The following will still crash with a runtime exception instead of failing appropriately

```scala
io.circe.jawn.decode[Double]("-.1")
```

This is because of a bug in Jawn (non/jawn#47). This isn't ideal, but I'm leaning toward waiting for the bug to be fixed instead of working around it.


travisbrown:
P.S. I know `BiggerDecimal` is a really stupid name—it was a working title that I was too tired to come up with a replacement for before posting this.

core/shared/src/main/scala/io/circe/Decoder.scala
```diff
@@ -301,6 +301,10 @@ final object Decoder extends TupleDecoders with LowPriorityDecoders {
     c.focus match {
       case JNull => Xor.right(Float.NaN)
       case JNumber(number) => Xor.right(number.toDouble.toFloat)
+      case JString(string) => Xor.fromOption(
+        JsonNumber.fromString(string).map(_.toDouble.toFloat),
```
knutwalker: `toDouble.toFloat` may overflow and result in `Infinity`, e.g. on `Double.MaxValue = 1.7976931348623157E308`. Since the other number conversions now fail instead of truncating/overflowing, I think this one should too.

travisbrown: @knutwalker Thanks for pointing this out.

`Double` is kind of different from the integral decoders, though. `BiggerDecimal.toDouble` isn't partial, so it does truncate (or rather overflow to infinity) and can lose precision, and I think `Float` should probably follow suit here, although I'm happy to hear counter-arguments, of course. This behavior should be much better documented, too.

easel:
Looks like a solid improvement to me.

core/shared/src/main/scala/io/circe/Decoder.scala
travisbrown: @knutwalker I've added some documentation that explains the current behavior.

non:
This looks good to me.

Is there anything Jawn should be doing to make this easier on you?

core/shared/src/main/scala/io/circe/Decoder.scala
knutwalker: Thanks for the documentation :+1:

travisbrown:
@non My only complaint about Jawn is non/jawn#47, which I haven't had time to take a look at yet. At first I thought some of the work here could potentially be moved to Jawn, but now I don't, really—I think the current division of responsibilities makes sense.

sjrd:
> Update: even after making the NumberParsing tests JVM-only, Scala.js fails because of missing BigInteger methods:

I'm happy to implement `BigInteger.longValueExact` if you [file an issue](https://github.com/scala-js/scala-js/issues). Note that it's a JDK8-only method. If you want to support JVM 7, you can't use it, anyway.

sjrd:
The `BigInteger` issue from Scala.js was fixed and the fix is available in 0.6.7, which we will announce in a couple of hours (but it's already on Maven Central). It's probably better that you just upgrade.

travisbrown:
Okay, I think everything's sorted here, so I'll be rebasing in a couple of minutes and then merging once it's green.

PR #190, opened 2016-02-04
Remove unnecessary type lambda in Decoder.kleisli

ceedubs:
Is there any reason to keep this?


travisbrown:
Nope, this is a leftover from before the type alias—thanks for catching it. (For what it's worth I'm only not using kind projector here because it doesn't work well with Scaladoc `-groups`.)

PR #191, opened 2016-02-04
Fix of broken link to benchmarks

Eternity-Yarr:


travisbrown:
Thanks, @Eternity-Yarr! Will merge when green.

PR #193, opened 2016-02-11
Add Encoder/Decoder for java.nio.ByteBuffer

BenWhitehead:


BenWhitehead:
I'm not totally sure I've got the generators setup correctly as I've seen an error similar to the following multiple times, but I'm not sure what I'm doing wrong.

```
[info] - Codec[ByteBuffer].codec.consistency with accumulating *** FAILED ***                                                                                                                                                                                                            
[info]   IllegalArgumentException was thrown during property evaluation.                                                                                                                                                                                                                 
[info]     Message: Input byte[] should at least have 2 bytes for base64 bytes                                                                                                                                                                                                           
[info]     Occurred when passed generated values (                                                                                                                                                                                                                                       
[info]       arg0 = "办" // 1 shrink                                                                                                                                                                                                                                                     
[info]     )
```

Any guidance would be greatly appreciated.

core/shared/src/main/scala/io/circe/Decoder.scala
```diff
@@ -1,11 +1,13 @@
 package io.circe
 
+import java.nio.ByteBuffer
```
vkostyukov: Just merge this with the rest of the imports. I'm not sure Circe's scalastyle settings allow this.

core/shared/src/main/scala/io/circe/Decoder.scala
```diff
@@ -1,11 +1,13 @@
...
+
 import cats.Monad
 import cats.data._
 import cats.std.list._
 import cats.syntax.cartesian._
 import cats.syntax.functor._
-import java.util.UUID
+import java.util.{ Base64, UUID }
```
vkostyukov: Don't we only have `Base64` in Java 8? If so maybe it makes sense to these instances under `circe-java8`?

core/shared/src/main/scala/io/circe/Decoder.scala
```diff
@@ -617,6 +619,10 @@ final object Decoder extends TupleDecoders with LowPriorityDecoders {
     override final def product[A, B](fa: Decoder[A], fb: Decoder[B]): Decoder[(A, B)] = fa.and(fb)
     final def flatMap[A, B](fa: Decoder[A])(f: A => Decoder[B]): Decoder[B] = fa.flatMap(f)
   }
+
+  implicit val decodeByteBuffer: Decoder[ByteBuffer] = decodeString.map { b64String =>
```
vkostyukov: I'd probably vote for using `decodeString.map(s => ....)` style (no curly braces) given that there is just one simple expression in the closure.

core/shared/src/main/scala/io/circe/Encoder.scala
```diff
@@ -1,9 +1,11 @@
 package io.circe
 
+import java.nio.ByteBuffer
```
vkostyukov: Merge imports, please.

core/shared/src/main/scala/io/circe/Encoder.scala
```diff
@@ -253,6 +255,27 @@ object Encoder extends TupleEncoders with LowPriorityEncoders {
   implicit final val contravariantEncode: Contravariant[Encoder] = new Contravariant[Encoder] {
     final def contramap[A, B](e: Encoder[A])(f: B => A): Encoder[B] = e.contramap(f)
   }
+
+  implicit val encodeByteBuffer: Encoder[ByteBuffer] = instance { bb =>
+    val bytes = if (bb.hasArray){
```
vkostyukov: Why we don't care about `position` and `remaining` in case if there is an array behind `ByteBuffer`? I'm not sure I have a good answer though, but maybe should just encode the remaining bytes then? I guess that's what users's would expect to happen.

core/shared/src/main/scala/io/circe/Decoder.scala
travisbrown: Ugh, yeah, I don't think we should drop Java 7 support quite yet, so it looks like if we want these instances in core, we'd need to use the `javax.xml.bind.DatatypeConverter` approach, which isn't going to work on Scala.js (where it looks like we'd need to use [`btoa`](https://groups.google.com/forum/#!msg/scala-js/En8WL-WIY_Q/147WEVIIOKAJ)).

I'd personally prefer to have these instances in core, since `ByteBuffer` itself isn't Java 8-specific. Maybe we could switch to `DatatypeConverter` for this PR, and I can follow up with a Scala.js-specific implementation?

core/shared/src/main/scala/io/circe/Encoder.scala
travisbrown: Yeah, this is a little more complex than I anticipated when I suggested providing `ByteBuffer` rather than `Array[Byte]` instances yesterday. My feeling is that we should respect the position and limit of the buffer to be encoded, and that encoding should not change these values.

tests/shared/src/test/scala/io/circe/ByteBufferCodecSuite.scala
```diff
@@ -0,0 +1,36 @@
...
+    Gen.oneOf(Seq(
+      'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M',
+      'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z',
+      'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm',
+      'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z',
+      '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '+', '/'
+    ))
+  )
+
+  implicit val base64StringArbitrary: Arbitrary[String] = Arbitrary(
```
travisbrown: I think it would be better to get arbitrary base 64 strings by getting an arbitrary array of bytes and then encoding them, but that gets us back into the land of needing multiple ways to do this to support Java <8 and Scala.js. Let me think a bit about this…

core/shared/src/main/scala/io/circe/Decoder.scala
vkostyukov: I think `DatatypeConverter` makes sense until we drop Java 7 support.

Any reason we don't want to drop it now? As I recall, 1.7 eol-ed on Apr 2015.

BenWhitehead:
Sorry for not checking in sooner, been busy with a release.

@travisbrown Did you come to a decision regarding java8 deps in the core module?

travisbrown:
@BenWhitehead We want to keep supporting Java 7 for now. I'd be happy to see these go in circe-java8, either for `ByteBuffer` or as a simpler (non-implicit) `decodeByteArray: Decoder[Array[Byte]]` that users could put into implicit scope themselves (so that we don't smash the default `Array[Byte]` instance for everyone).

travisbrown:
@BenWhitehead I am planning to publish 0.4.0 this evening or tomorrow morning at the latest, though.

BenWhitehead:
@travisbrown I won't have enough time to update the PR before then unfortunately. 

I'm just going to close this for now and open a new one in the future when I get time to do it.

travisbrown:
@BenWhitehead Do you mind if I re-open this to track the need for these instances? You needn't feel like it's your responsibility. 😄

BenWhitehead:
@travisbrown Sure, feel free. Let me know if you want/need help.

travisbrown:
Oh, sorry—just noticed that this is a PR, not an issue. I'll close again and open an issue later.

travisbrown:
Maybe we should revisit this now that we require Java 8 for circe-core?

BenWhitehead:
Yay, Java 8 core! Happy to help where I can.

PR #194, opened 2016-02-12
Update Scala.js to 0.6.7

travisbrown:


travisbrown:
I've verified this locally and Travis CI doesn't test on Scala.js anyway, so I'm going to go ahead and merge.

PR #195, opened 2016-02-13
Preparation for 0.3.0

travisbrown:
No more snapshots.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,10 +19,6 @@ script:
     then sbt ++$TRAVIS_SCALA_VERSION clean validateJS;
     else
       sbt ++$TRAVIS_SCALA_VERSION clean coverage validateJVM &&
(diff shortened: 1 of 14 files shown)


sjrd:
In `build.sbt`, replace `scalaJSStage in Global := FastOptStage` by `scalaJSUseRhino in Global := false`. Typical breaking change of 0.6.5 -> 0.6.6, as highlighted in the [release notes](https://www.scala-js.org/news/2016/01/25/announcing-scalajs-0.6.6/).

travisbrown:
@sjrd Thanks, but that doesn't seem to help. The strange thing is that I'd been able to run the tests locally until this morning (they were slow, but I could live with that), and the only change here that seems at all possibly relevant is the Shapeless version.

travisbrown:
Just confirmed that changing the Shapeless version back to the 2.3.0 snapshot fixes the Scala.js tests.

Issue #166, opened 2016-01-15
Provide a macro-annotation to reduce defining Encoder/Decoder pairs


dwijnand:
From your blog post https://meta.plasm.us/posts/2016/01/14/configuring-generic-derivation/

I was wondering if perhaps

``` scala
import io.circe.generic.semiauto._

case class Foo(thisIsSomeNumber: Int = 0)
object Foo {
  implicit val decodeFoo: Decoder[Foo] = deriveConfiguredDecoder[Foo, SnakeCaseKeys]
  implicit val encodeFoo: Encoder[Foo] = deriveConfiguredEncoder[Foo, SnakeCaseKeys]
}
```

could be further boilerplate-reduced to

```
import io.circe.macros.MkCodec // or some other name
import io.circe.generic.semiauto._

@MkCodec case class Foo(thisIsSomeNumber: Int = 0)
```

I've read your "The public API should not contain unnecessary types" point in [DESIGN.md](https://github.com/travisbrown/circe/blob/94379728f8c4d3c799c076d0257b191ce58195c7/DESIGN.md) and #133, and I think this doesn't go against those wishes.

travisbrown:
I've been resisting adding macros to circe, but the seal is broken now, so I think we should consider this.

We'd have to change the language in the first item in the design doc a bit. The prohibition against annotations there is primarily intended to target member-level annotations, and I don't think this kind of thing (where the macro annotation is just nice syntax for `semiauto` definitions) is so bad.

We'd probably want something like `@ConfiguredJsonCodec[C]` in addition to `@JsonCodec` (or whatever they end up being called).

Are you interested in adding this, @dwijnand? I'm not sure I'd be able to get to it for a week or so.

dwijnand:
I'm not against rolling up my sleeves and contributing, but I've a stack of sbt things I'd like to tend to, so I don't think I'll be able to get to it in a week or so.

ngbinh:
This is something we love to have also.

dwijnand:
Initial support for this is now in master (via #203).

Support for a proposed `@ConfiguredJsonCodec[C]` one could say is blocked on still pending #164.

dwijnand:
I should rivisit the tail part of this given https://github.com/circe/circe/pull/429 is merged.

travisbrown:
#429 includes `@ConfiguredJsonCodec` so I think we can close this.

PR #203, opened 2016-02-16
Initial implementation of @JsonCodec

dwijnand:
Refs #166

---

This is only an initial cut, submitted for early feedback.

Please feel free to review aggressively, I'm happy to tweak any minor detail.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -184,8 +184,7 @@ lazy val genericBase = crossProject.in(file("generic"))
   .settings(
     libraryDependencies ++= Seq(
       "com.chuusai" %%% "shapeless" % shapelessVersion,
-      "org.scala-lang" % "scala-compiler" % scalaVersion.value % "provided",
-      "org.typelevel" %%% "macro-compat" % "1.1.1"
+      "org.scala-lang" % "scala-compiler" % scalaVersion.value % "provided"
     )
   )
   .jsSettings(commonJsSettings: _*)
--- a/build.sbt
+++ b/build.sbt
@@ -182,7 +182,11 @@ lazy val genericBase = crossProject.in(file("generic"))
   )
   .settings(allSettings: _*)
   .settings(
-    libraryDependencies += "com.chuusai" %%% "shapeless" % shapelessVersion
+    libraryDependencies ++= Seq(
+      "com.chuusai" %%% "shapeless" % shapelessVersion,
+      "org.scala-lang" % "scala-compiler" % scalaVersion.value % "provided",
+      "org.typelevel" %%% "macro-compat" % "1.1.1"
+    )
   )
   .jsSettings(commonJsSettings: _*)
   .jvmConfigure(_.copy(id = "generic"))
--- a/generic/shared/src/main/scala/io/circe/generic/JsonCodecMacros.scala
+++ b/generic/shared/src/main/scala/io/circe/generic/JsonCodecMacros.scala
@@ -31,8 +31,8 @@ private[generic] class JsonCodecMacros(val c: blackbox.Context) {
       q"""
        $clsDef
        object $objName extends { ..$objEarlyDefs} with ..$objParents { $objSelf =>
-         ..${codec(clsDef)}
          ..$objDefs
+         ..${codec(clsDef)}
        }
        """
     case _ => c.abort(c.enclosingPosition,
--- a/generic/shared/src/main/scala/io/circe/generic/JsonCodecMacros.scala
+++ b/generic/shared/src/main/scala/io/circe/generic/JsonCodecMacros.scala
@@ -37,15 +37,14 @@ private[generic] class JsonCodecMacros(val c: blackbox.Context) {
 
   private[this] val DecoderClass = symbolOf[Decoder[_]]
   private[this] val EncoderClass = symbolOf[Encoder[_]]
+  private[this] val semiautoObject = symbolOf[semiauto.type].asClass.module
 
   private[this] def codec(tpname: TypeName): List[Tree] = {
     val decodeNme = TermName("decode" + tpname.decodedName)
     val encodeNme = TermName("encode" + tpname.decodedName)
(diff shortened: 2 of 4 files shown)


dwijnand:
Ok finally, as side-effect indicated by codecov, this is now passing TravisCI.

ngthanhtrung:
I think:

1. We should reject classes which have type parameters with a more detailed compiler message.
1. We should not enforce the target to be a case classes. Sometimes a sealed trait/class is fine.
1. Codec definitions should be placed at the end of the companion object. This is for the case which sub-classes are defined inside the companion object.

build.sbt
```diff
@@ -182,7 +182,11 @@ lazy val genericBase = crossProject.in(file("generic"))
   )
   .settings(allSettings: _*)
   .settings(
-    libraryDependencies += "com.chuusai" %%% "shapeless" % shapelessVersion
+    libraryDependencies ++= Seq(
+      "com.chuusai" %%% "shapeless" % shapelessVersion,
+      "org.scala-lang" % "scala-compiler" % scalaVersion.value % "provided",
+      "org.typelevel" %%% "macro-compat" % "1.1.1"
```
travisbrown: We get this transitively from Shapeless, right? If we need to include it here, we should probably define a `macroCompatVersion` since we use it twice now.

tests/shared/src/test/scala/io/circe/generic/SemiautoDerivedSuite.scala
```diff
@@ -48,7 +48,7 @@ class SemiautoDerivedSuite extends CirceSuite {
     implicit val encodeRecursiveAdtExample: Encoder[RecursiveAdtExample] = deriveEncoder
   }
 
-  case class RecursiveWithOptionExample(o: Option[RecursiveWithOptionExample])
+  @JsonCodec case class RecursiveWithOptionExample(o: Option[RecursiveWithOptionExample])
```
travisbrown: I think I'd prefer to have a new case class (or possibly even a new suite?) rather than change existing tests.

generic/shared/src/main/scala/io/circe/generic/JsonCodecMacros.scala
```diff
@@ -0,0 +1,50 @@
...
+class JsonCodec extends scala.annotation.StaticAnnotation {
+  def macroTransform(annottees: Any*): Any = macro JsonCodecMacros.jsonCodecAnnotationMacro
+}
+
+@bundle
+private[generic] class JsonCodecMacros(val c: blackbox.Context) {
+  import c.universe._
+
+  def jsonCodecAnnotationMacro(annottees: Tree*): Tree = annottees match {
+    case List(clsDef: ClassDef) if clsDef.mods hasFlag Flag.CASE =>
```
travisbrown: This is nitpicky, but I'd like to stick to a no-infix-syntax-for-non-operator-methods rule. That really should be enforced by Scalastyle, though, so we could merge this and fix once a rule is in place and complains about it.

generic/shared/src/main/scala/io/circe/generic/JsonCodecMacros.scala
```diff
@@ -0,0 +1,50 @@
...
+       """
+    case _ => c.abort(c.enclosingPosition, "Invalid annotation target: must be a case class")
+  }
+
+  private[this] val DecoderClass = symbolOf[Decoder[_]]
+  private[this] val EncoderClass = symbolOf[Encoder[_]]
+  private[this] val semiautoObj = symbolOf[semiauto.type].asClass.module
+
+  private[this] def codec(tpname: TypeName): List[Tree] = {
+    val decodeNme = TermName("decode" + tpname.decodedName)
```
travisbrown: I haven't really thought this through yet, but what do we want to happen in the case of name collisions? Right now there's just a compiler error, and that may be fine, but it might also make sense to use `fresh` to have a unique name here.

travisbrown:
:+1: to @ngthanhtrung's first two comments. I think it probably makes sense just to have the annotation generate the code in the companion object whether or not it compiles, rather than try to perform the equivalent checks independently. As long as there's documentation for the annotation that makes it clear what it does, the error messages should highlight the problem fairly effectively.

@ngthanhtrung, off the top of my head I'm not seeing how the location definitions in the companion object could make a difference, but if there's any reason at all to prefer to put them at the end, I support that as well.

generic/shared/src/main/scala/io/circe/generic/JsonCodecMacros.scala
dwijnand: I'm don't think using `fresh` is a good idea here, because it's not stable between compiles.

Are you happy to leave this as is for now and revisit it in the future?

travisbrown: @dwijnand I agree that it's not ideal, but given that we're using export-hook that ship has already sailed. In this case I think just letting collisions collide is a reasonable approach, though, so let's leave it as is.

dwijnand:
I think I've addressed everything, if you'd like to review again please. Again, I've no problem re-working any minor detail.

Particularly I wasn't quite sure what to do test-wise, so I've gone for test-parity with the existing tests, but that might be too much duplication or it might be not exhaustive enough.. Also would you like to add some negative / illTyped tests? Let me know.

ngthanhtrung:
@travisbrown It's because I once had an issue with placing codecs at the beginning of the companion object in the past (something like the macro didn't know about all sub-classes of the sealed trait already). I've just checked it again and it works just fine so anywhere will do.

@dwijnand The PR looks neat now. You did a great job :+1:

travisbrown:
This looks great, @dwijnand, thanks much! (and thanks @ngthanhtrung for reviewing).

dwijnand:
Ah excellent. Thanks Travis.

PR #210, opened 2016-02-24
Overhaul Json API

travisbrown:
This is a fairly large but superficial change that brings some consistency to the `Json` API. Highlights:

* The `number*` methods that take a `Double` are now `fromDouble*`. It's not like `Double` is somehow more "number-y" than any other numeric types.
* `Json.int` is now `Json.fromInt`, in part because `int` is a keyword in Java.
* More generally, conversion methods are now `fromWhatever`, where `Whatever` is the full name of the type. For example `Json.bool` is now `Json.fromBoolean`.
* `Json.obj` was abbreviated (since `object` is a Scala keyword) but `Json.array` was not. They're now both abbreviated for the sake of consistency.
* I've replaced `JsonObject.fromIndexedSeq` with `JsonObject.fromIterable` since there was no real advantage to requiring an `IndexedSeq` there.
* I've added a `Json.fromBigInt` method.
* I've used public methods instead of package private constructors wherever possible.

No existing code will break, although anyone who's using `Json` methods is likely to see some deprecations. All of these deprecated methods will probably disappear in 0.5, but will definitely be available in the 0.4 releases.

### DIFF
--- a/core/shared/src/main/scala/io/circe/Encoder.scala
+++ b/core/shared/src/main/scala/io/circe/Encoder.scala
@@ -114,7 +114,7 @@ object Encoder extends TupleEncoders with LowPriorityEncoders {
   /**
    * @group Encoding
    */
-  implicit final val encodeString: Encoder[String] = instance(Json.string)
+  implicit final val encodeString: Encoder[String] = instance(Json.fromString)
 
   /**
    * @group Encoding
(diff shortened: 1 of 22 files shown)


notxcain:
`Json.fromString` sounds like it could be used to parse JSON from a string.

travisbrown:
@notxcain I've added some API docs to make the intention of the `from` methods clearer—what do you think?

notxcain:
@travisbrown looks much better now! I still got feeling that these changes decrease readability. It would be great to here from someone else.

easel:
I'm good with all these changes. It will be a bit of cut and paste, but the improved consistency will aid discoverability in the future. Also +1 to the use of public methods =p

travisbrown:
@notxcain Are you okay with us merging this for now, on the understanding that there's still room for improvement before 1.0?

notxcain:
Yep! I'm cool. :rocket:

PR #213, opened 2016-02-25
Key codecs for maps with non-string keys

travisbrown:
This PR includes and extends @pjan's work in #186 to support encoding _and_ decoding maps with non-string keys. Off-the-shelf support is provided for keys of type `String`, `Symbol`, and the primitive integral types, and other key type instances can easily be implemented by users.


easel:
Very pleased to see this!

travisbrown:
I'm going to go ahead and merge this for the sake of speeding things along toward 0.4.0, but I'm still interested in feedback.

danielyli:
I'm fairly new to Circe, and I'm not certain how to encode/decode a non-string key.

Specifically, I'm trying to encode/decode a `Map[Foo, Bar]`, with implicit codecs for `Foo` and `Bar` in scope.  The compiler still complains that no codec was found for type `Map[Foo, Bar]` — my guess is that I'm missing one last piece (probably addressed by this ticket).  What am I missing?

    import io.circe._, io.circe.generic.auto._, io.circe.parser._, io.circe.syntax._
    import my.package.models.{ Foo, Bar }
    import my.package.circecodecs._    // codecs for Foo and Bar

    val m: Map[Foo, Bar] = ???
    m.asJson    // compiler complains it cannot find implicit value here

easel:
Does your `circecodecs._` include something like this:

danielyli:
Yes, I believe that did it.  Thank you!

Can I suggest that this be added to the Readme, as it seems like a fairly common use case?

danielyli:
Cross-referencing with #341.

PR #214, opened 2016-02-25
Update to refined 0.3.6

fthomas:
This updates to refined 0.3.6 which depends on shapeless 2.3.0.


travisbrown:
Thanks @fthomas! Will merge when it's green.

PR #218, opened 2016-03-03
Fix #217

### LINKED ISSUE #217: decoding 'null' for a case class with only Option fields does not fail fast as expected
@ty707 and I found a bug with automatic decoding + derivation in which when decoding a `null` to a case class of only `Option` fields, it doesn't fail fast as expected - it succeeds and returns an instance of the case class where all of its fields are `None`.

Example error was repro'd on master @ dee431a5cf364f4ed4cd113f494bfe84cc826f11.

Pasting my REPL session:

```scala
import io.circe._
import io.circe.literal._
import io.circe.parser._
import io.circe.syntax._
import cats.data.Xor

scala> case class Container[T](data: T)
defined class Container

scala> case class Item(a: Option[Int], b: Option[String])
defined class Item

// I'd expect this to fail hard because data: T was improperly defined, since the "data" field is in fact nullable/optional
scala> decode[Container[Item]]("""{"data":null}""")
res0: cats.data.Xor[io.circe.Error,Container[Item]] = Right(Container(Item(None,None)))

// I'd also expect this to fail hard
scala> decode[Item]("""null""")
res1: cats.data.Xor[io.circe.Error,Item] = Right(Item(None,None))

scala> case class Qwer(a: Option[Int], b: String)
defined class Qwer

// This seems to behave as expected, having a single non-Option field causes it to fail hard
scala> decode[Qwer]("""null""")
res2: cats.data.Xor[io.circe.Error,Qwer] = Left(io.circe.DecodingFailure: Attempt to decode value on failed cursor: El(DownField(b),false))
```

travisbrown:
This is a quick draft of a fix for #217, which reports a bug (inherited from and also still present in Argonaut) where case classes with all optional fields can be successfully decoded from a JSON `null`:

```scala
scala> import io.circe._, io.circe.jawn._
import io.circe._
import io.circe.jawn._

scala> case class Foo(o: Option[String])
defined class Foo

scala> implicit val decodeFoo: Decoder[Foo] = Decoder.instance(c =>
     |   c.downField("o").as[Option[String]].map(Foo(_))
     | )
decodeFoo: io.circe.Decoder[Foo] = io.circe.Decoder$$anon$8@5f7b1ec9

scala> decode[Foo]("null")
res0: cats.data.Xor[io.circe.Error,Foo] = Right(Foo(None))
```

If you don't have any case classes with all optional members, or you don't mind that they can be decoded from a JSON `null` (as opposed to at least a `{}`), you're fine.

At a glance I think this is related to the issue reported in [this recent Argonaut pull request](https://github.com/argonaut-io/argonaut/pull/213/files), although the changes there don't fix this part of the problem:

```scala
scala> import argonaut._, Argonaut._
import argonaut._
import Argonaut._

scala> case class Foo(o: Option[String])
defined class Foo

scala> implicit val codec = jdecode1L[Option[String], Foo](Foo(_))("o")
codec: argonaut.DecodeJson[Foo] = argonaut.DecodeJson$$anon$4@4e71965f

scala> Parse.decodeOption[Foo]("null")
res0: Option[Foo] = Some(Foo(None))
```

This PR currently works in our simple example:

```scala
scala> import io.circe._, io.circe.jawn._
import io.circe._
import io.circe.jawn._

scala> case class Foo(o: Option[String])
defined class Foo

scala> implicit val decodeFoo: Decoder[Foo] = Decoder.instance(c =>
     |   c.downField("o").as[Option[String]].map(Foo(_))
     | )
decodeFoo: io.circe.Decoder[Foo] = io.circe.Decoder$$anon$8@251634

scala> decode[Foo]("null")
res0: cats.data.Xor[io.circe.Error,Foo] = Left(io.circe.DecodingFailure: [A]Option[A]: El(DownField(o),false))
```

But it's only a sketch (I'm at the Typelevel Summit today and haven't been able to take a closer look) and won't work in more interesting cases.

I'll wrap it up and publish an 0.3.1 as soon as possible (probably this evening or tomorrow).

### DIFF
--- a/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -527,7 +527,9 @@ final object Decoder extends TupleDecoders with LowPriorityDecoders {
           case Xor.Left(df) if df.history.isEmpty => rightNone
           case Xor.Left(df) => Xor.left(df)
         }
-      } else rightNone
+      } else if (c.any.focus.isObject) {
+        rightNone
+      } else Xor.left(DecodingFailure("[A]Option[A]", c.history))
     )
 
   /**


travisbrown:
Quick update: there are a couple of problems here. The first is that the behavior is underspecified—for example:

```scala
val c = json"""{ "a": 1 }""".hcursor

c.downField("a").downField("b").as[Option[Int]]
```

Should this succeed with a `None` or fail? My feeling is it probably shouldn't succeed, but that complicates things.

Another (much less important) concern is that if we start policing the structure of the failure that the `Option` decoder starts with, that's going to introduce a significant amount of overhead into some code that gets executed _a lot_.

I'm thinking we should go with a fairly simple solution for now: we check the failure history and if there's a single failed operation and it's either a `downField` or a `downN` (and the focus is either an object or array as appropriate), we succeed with a `None`.

longcao:
@travisbrown For what it's worth, in the perspective of myself as a user in the `c.downField("a").downField("b").as[Option[Int]]` example, I'd expect it to return `None`, since in that situation I would think that I'm trying to read a field where any undefined key or `null` value signals a `None`. At least, I _think_ this is what makes sense in my head.

travisbrown:
Superseded by #231.

PR #219, opened 2016-03-04
Included netvl/picopickle in the benchmark

mgttlinger:


travisbrown:
Thanks, @mgttlinger! Running this locally now.

Issue #188, opened 2016-02-03
Pretty-printing of decoding failure history


ArtemPyanykh:
Consider the following example:

```scala
scala> val json = parse("""[1, 2, "hello", 4, "world", 6]""").getOrElse(Json.empty)
scala> Decoder[List[Int]].apply(json.hcursor)
res5: io.circe.Decoder.Result[List[Int]] = Left(io.circe.DecodingFailure: Int: El(MoveRight,true),El(MoveRight,true),El(DownArray,true))
```

This part might be hard to parse for a human eye:
```scala
io.circe.DecodingFailure: Int: El(MoveRight,true),El(MoveRight,true),El(DownArray,true)
```

I guess we could improve the error message by reversing a history, making it a bit more visual, and maybe adding the element we failed to parse. Something along these lines:
```
[↴ array]
  →|→ × Failed to decode hello as Int
```

I'm not sure about arrow symbols though (on github they are very tiny), so maybe just printing
```
[down array]
  [right], [right] × Failed to decode hello as Int
```

ArtemPyanykh:
I can take a look into it in a couple weeks after I get back from vacation.

travisbrown:
:+1:, this sounds great.

tobias-johansson:
**WIP**
I've done some work on a `Show[DecodingFailure]` that tries to make simple, compact messages by representing the cursor history as javascript style selection, i.e. `"DecodingFailure at .foo.bar[2].baz: some message"`.
I believe this is a good idea if you want to return error messages to http clients, etc.

WIP can be seen here: https://github.com/travisbrown/circe/compare/master...tobias-johansson:Show-DecodingFailure
It deals with normal navigation operations, and if there are other kinds of operations it falls back to `Show[CursorOp]` for those.

**Problem**
The error messages could still be improved though, and I would like to discuss if it's possible to encode the reason for failure in `DecodingFailure`s in a better way.
At the moment if I do
```scala
scala> decode[Int](""""a"""")
res8: cats.data.Xor[io.circe.Error,Int] = Left(io.circe.DecodingFailure: Int)
```
the message in the `DecodingFailure` is simply the string `"Int"` which isn't of much help when trying to create a nice error message.

**Suggestion**
I'm suggesting an explicit encoding of the reason to replace or complement the `message` field in `DecodingFailure`, something like:
```scala
sealed trait Reason
case class TypeExpectation(expected: String) extends Reason
case class FailedValidation(message: String) extends Reason
[...]

PR #222, opened 2016-03-07
Instance of Show[Error]

tobias-johansson:
Ref #188 

Adds instances of `Show[ParsingFailure]`, `Show[DecodingFailure]` and `Show[Error]`

- `ParsingFailure`:s are shown as `"ParsingFailure: the message"`
- `DecodingFailure`:s are shown as `"DecodingFailure at .foo.bar[2].baz: the message"` where the cursor op history is represented as js style selection (falls back to `Show[CursorOp]` for ops that are not simple navigation ops)
- `Show[Error]` simply delegates to the two other instances

### DIFF
--- a/core/shared/src/main/scala/io/circe/Error.scala
+++ b/core/shared/src/main/scala/io/circe/Error.scala
@@ -1,7 +1,9 @@
 package io.circe
 
 import algebra.Eq
+import cats.Show
 import cats.std.list._
+import io.circe.CursorOp._
 
 sealed trait Error extends Exception
 
@@ -20,6 +22,10 @@ final object ParsingFailure {
   implicit final val eqParsingFailure: Eq[ParsingFailure] = Eq.instance {
     case (ParsingFailure(m1, t1), ParsingFailure(m2, t2)) => m1 == m2 && t1 == t2
   }
+
+  implicit final val showParsingFailure: Show[ParsingFailure] = Show.show { failure =>
+    s"ParsingFailure: ${failure.message}"
+  }
 }
 
 final object DecodingFailure {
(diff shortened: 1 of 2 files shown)


travisbrown:
@tobias-johansson This looks great, thanks! I've got a couple of questions / comments that I'll make inline.

core/shared/src/main/scala/io/circe/Error.scala
```diff
@@ -27,6 +34,44 @@ final object DecodingFailure {
...
   }
+
+  /**
+    * Creates compact, human readable string representations for DecodingFailure
+    * Cursor history is represented as JS style selections, i.e. ".foo.bar[3]"
+    */
+  implicit final val showDecodingFailure: Show[DecodingFailure] = Show.show { failure =>
+
+    /** Represents JS style selections into JSON structure */
+    sealed trait Selection
```
travisbrown: This is a relatively minor issue, but I think I'd like to try to avoid defining types in methods whenever possible—these get stuck in `DecodingFailure` and leak into its public API. We could either bump them up into `DecodingFailure` and make them `private[this]` there, or we could define `Show` as an anonymous class instance (I don't have a strong opinion either way).

core/shared/src/main/scala/io/circe/Error.scala
```diff
@@ -20,6 +22,11 @@ final object ParsingFailure {
   implicit final val eqParsingFailure: Eq[ParsingFailure] = Eq.instance {
     case (ParsingFailure(m1, t1), ParsingFailure(m2, t2)) => m1 == m2 && t1 == t2
   }
+
+  implicit final val showParsingFailure: Show[ParsingFailure] = Show.show { failure =>
+    val name = classOf[ParsingFailure].getSimpleName
```
travisbrown: What's the argument for getting the name like this? I guess it could help in refactoring, but these names are likely to be pretty stable, and I'm inclined to prefer a blanket prohibition on runtime reflection in circe-core (even though in this case it's clearly not unsafe). I could be convinced the other way, though.

tobias-johansson: Yeah for staying safe if ever refactoring. But this is easily caught by tests instead, so let's go with hard coded names.

core/shared/src/main/scala/io/circe/Error.scala
tobias-johansson: Ok thanks, didn't realize this. Fix coming up

tobias-johansson:
Any further comments?
If not, I am to squash these commits and push -f to this branch right?

travisbrown:
@tobias-johansson Looks good to me! Squashing isn't strictly necessary, but if you don't mind let's do it here, at least for the last commit. Thanks again!

tobias-johansson:
I just went ahead and squashed the whole thing.
Please merge at your leisure.

travisbrown:
@tobias-johansson Perfect, thanks!

travisbrown:
We should document this feature more completely, but for now here's an example:

```scala
import cats.std.list._
import cats.syntax.foldable._, cats.syntax.functor._, cats.syntax.show._
import io.circe.Decoder
import io.circe.jawn.decode
import io.circe.literal._

case class Foo(bar: Bar)
case class Bar(baz: Baz)
case class Baz(qux: Qux, i: List[Int])
case class Qux(a: String, b: Boolean)

val doc = json"""{
  "bar": {
    "baz": {
      "i": [1, 2, true],
      "qux": { "a": 10, "b": "TRUE" }
    }
  }
}"""

val result = Decoder[Foo].accumulating(doc.hcursor)
val errors = result.swap.toList.flatMap(_.map(_.show).toList)
```

And then:

```scala
scala> errors.foreach(println)
DecodingFailure at .bar.baz.qux.a: String
DecodingFailure at .bar.baz.qux.b: Boolean
DecodingFailure at .bar.baz.i[2]: Int
```

devnoname120:
@travisbrown Under circle `0.10.0`, your example only returns `DecodingFailure at : CNil`. Has the way to accumulate errors changed?

Issue #228, opened 2016-03-11
Proposed design changes for decoding


travisbrown:
I've discussed the idea outlined below with a few people (in person at the Typelevel summit last week, etc.), but it's somewhat relevant to the plans in #226, so I decided it's time to go ahead open an issue for it.

Paul Chiusano has a [blog post](https://pchiusano.github.io/2014-08-12/zippers-not-useful.html) where he argues that cursors aren't the right abstraction for things like JSON decoding. They're too powerful—they provide modification operations that don't make sense at all in the context of decoding and navigation operations (moving _up_ the tree) that are probably always a bad idea there.

So far I've stuck with Argonaut's cursor-based model for decoding because it's a simple solution if you want to be able to track history during decoding and already have a nice cursor implementation lying around. The full power of cursors is useful in non-decoding contexts, so we don't want to strip out the parts that are unnecessary for decoding, which means we'd need a new mechanism that looks a lot like a cursor without those parts. So far that hasn't seemed worth the effort.

I think that I'd like to make this change in circe 0.5. You'll still be able to construct decoders from `HCursor => Decoder.Result[A]`, but the preferred route will be to use a new algebra that provides a more limited set of navigation operations. You build up your decoder using these operations and then interpret it into one of several contexts:
- `Option` if you want fail-fast decoding with no history or position information.
- `Xor[DecodingFailure, ?]` for fail-fast decoding with history / positions.
- `ValidatedNel[DecodingFailure, ?]` for error accumulation with history / positions.

This has a handful of advantages. The first is just that as an abstraction it fits the problem more elegantly—it's less possible to construct nonsensical decoders (e.g. decoders that make modifications that will just be thrown away or decoders that snoop around higher up the document than the current focus).

The second advantage is that it supports potential optimizations that the current model doesn't. Most obviously, if you don't care about having position information to put in your error messages, you don't have to pay the cost of tracking history as your decoder moves around the document. It also may be possible to rearrange navigation operations to avoid unnecessary work (that's probably further down the line, though).

The goal would be to make this switch without breaking existing code (at least for a deprecation cycle), but it would be a major change, and some decoders you can write right now simply wouldn't be possible under the new model (only uselessly strange ones, I hope).

Does this seem reasonable for circe 0.6? Any objections or concerns?

longcao:
We spoke about this offline at Typelevel Summit last week and I know I would be interested (hopefully others too) in what kind of optimizations would be possible here! Do I smell a new blog post?

[...]

Issue #226, opened 2016-03-11
Proposed design changes for optional decoders


travisbrown:
This issue is closely related to #217 and #218, but there are a couple of unresolved underlying questions about what the "correct" behavior should be that I think deserve their own discussion.

## Setup

Suppose we have a case class like this:

```scala
case class Foo(ab: Option[Int])
```

And we want to decode a document like this into `Foo(Some(1))`:

```json
{ "a": { "b": 1 }}
```

## Chains of operations

We can currently write a decoder like this:

```scala
implicit val decodeFoo: Decoder[Foo] =
  Decoder.instance(_.downField("a").downField("b").as[Option[Int]].map(Foo(_)))
```

And confirm that it works on valid input:

```scala
scala> decode[Foo]("""{ "a": { "b": 1 }}""")
res0: cats.data.Xor[io.circe.Error,Foo] = Right(Foo(Some(1)))
```

If there's no `b` field in the inner JSON object we get a `None`, which I think is pretty reasonable:

```scala
scala> decode[Foo]("""{ "a": { "x": 1 }}""")
res1: cats.data.Xor[io.circe.Error,Foo] = Right(Foo(None))
```

If there is a `b` field but it's not an integral number, decoding fails (also reasonable to me):

```scala
scala> decode[Foo]("""{ "a": { "b": true }}""")
res3: cats.data.Xor[io.circe.Error,Foo] = Left(io.circe.DecodingFailure: Int: El(DownField(b),true),El(DownField(a),true))
```

If there's no `a` field in the outer object, we also get a `None` (somewhat more questionable to my eye, but still I think the right choice):

```scala
scala> decode[Foo]("""{ "x": { "b": 1 }}""")
res5: cats.data.Xor[io.circe.Error,Foo] = Right(Foo(None))
```

The next part is where it gets weird. Suppose we do have an `a` field, but its value isn't an object:

```scala
scala> decode[Foo]("""{ "a": true}""")
res6: cats.data.Xor[io.circe.Error,Foo] = Right(Foo(None))
```

Or that we don't have an `a` field at all (this is more or less what #217 reports):

```scala
scala> decode[Foo]("true")
res7: cats.data.Xor[io.circe.Error,Foo] = Right(Foo(None))
```

I think both of these cases should result in decoding failures.

## Optional decoders and JSON arrays

A similar batch of issues comes up when our optional decoder uses positional selectors:

```scala
implicit val decodeFoo: Decoder[Foo] =
[...]

PR #231, opened 2016-03-13
Clearer failure behavior for optional decoders

travisbrown:
## Update

Okay, tests are in place and I've confirmed that they fail as expected on the current master, so this is ready for review.

## Original description

This PR still needs tests but I think it's a reasonable solution for #226 and #217. The underlying abstractions (see #228) will change a bit in 0.5.0, so I don't want to spend too much time worrying about the implementations, but I do want to be sure we get the right behavior now.


travisbrown:
I'm going to go ahead and merge this, but I'm still interested in feedback.

PR #233, opened 2016-03-15
Performance improvements from profiling common use cases

travisbrown:
Results for two runs of 0.3.0 (the first of each pair) and this PR:

```scala
[info] DecodingBenchmark.decodeFoosC              thrpt   40     2562.882 ±    23.607  ops/s
[info] DecodingBenchmark.decodeFoosC              thrpt   40     3590.287 ±    19.077  ops/s

[info] DecodingBenchmark.decodeIntsC              thrpt   40    11709.801 ±    54.986  ops/s
[info] DecodingBenchmark.decodeIntsC              thrpt   40    17154.936 ±    95.222  ops/s

[info] EncodingBenchmark.encodeFoosC              thrpt   40     6757.780 ±    32.703  ops/s
[info] EncodingBenchmark.encodeFoosC              thrpt   40     7450.555 ±    43.893  ops/s

[info] EncodingBenchmark.encodeIntsC              thrpt   40    94377.750 ±   814.382  ops/s
[info] EncodingBenchmark.encodeIntsC              thrpt   40   121730.641 ±   710.014  ops/s

[info] ParsingBenchmark.parseFoosC                thrpt   40     3051.392 ±    21.343  ops/s
[info] ParsingBenchmark.parseFoosC                thrpt   40     2926.140 ±    46.165  ops/s

[info] ParsingBenchmark.parseIntsC                thrpt   40    14202.027 ±   283.446  ops/s
[info] ParsingBenchmark.parseIntsC                thrpt   40    12964.576 ±   141.045  ops/s

[info] PrintingBenchmark.printFoosC               thrpt   40     3736.044 ±    56.407  ops/s
[info] PrintingBenchmark.printFoosC               thrpt   40     4134.720 ±    20.008  ops/s

[info] PrintingBenchmark.printIntsC               thrpt   40    22717.970 ±   142.034  ops/s
[info] PrintingBenchmark.printIntsC               thrpt   40    24389.645 ±   381.590  ops/s
```

The parsing benchmarks are just a bit worse, and I'm not sure why—they're close enough that it probably could just be noise.

PR #234, opened 2016-03-16
Codecs for Some and None

travisbrown:
This PR builds on #233 (for the sake of convenience only) and I'll rebase it once that's merged.

The issue is that while in general I don't think many people do or should encode or decode `Some` and `None` directly, it's possible and in some cases reasonable, and the current story is pretty confusing:

```scala
scala> Some(1).asJson.noSpaces
res0: String = [1]

scala> decode[Some[Int]]("[1]")
res1: cats.data.Xor[io.circe.Error,Some[Int]] = Left(io.circe.DecodingFailure: Attempt to decode value on failed cursor: El(DownField(x),false))

scala> None.asJson.noSpaces
res2: String = {}

scala> decode[None.type]("{}")
res3: cats.data.Xor[io.circe.Error,None.type] = Right(None)
```

These codecs are coming from all over the place—generic derivation, collection encoders, etc.

This PR provides explicit encoders and decoders for `Some` and `None` that behave more or less as you'd expect—decoding into `Some[A]` means you should be able to decode an `A`, and `None` is always encoded and decoded as a JSON null.

I've also snuck in a small fix for a test that was recently added as part of the optional decoding overhaul.

### DIFF
--- a/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -552,6 +552,21 @@ final object Decoder extends TupleDecoders with LowPriorityDecoders {
   /**
    * @group Decoding
    */
+  implicit final def decodeSome[A](implicit d: Decoder[A]): Decoder[Some[A]] = d.map(Some(_))
+
+  /**
+   * @group Decoding
+   */
+  implicit final val decodeNone: Decoder[None.type] = new Decoder[None.type] {
+    final def apply(c: HCursor): Result[None.type] = if (c.focus.isNull) Xor.right(None) else {
+      Xor.left(DecodingFailure("None", c.history))
+    }
+  }
+
+
+  /**
+   * @group Decoding
+   */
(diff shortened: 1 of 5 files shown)


easel:
Looks good! Interestingly I haven't ran into issues with these, I'm guessing because all my `Some` and `None` cases are embedded as `Option[A]` inside a `semiauto` codec.

It looks like this will make it difficult / impossible to encode None as a missing key, is that correct? In other words `None !== undefined`.

travisbrown:
@easel That's right about `None`—the idea is that if you want the more complex `null`-as-empty behavior you can and should be using `Option`. If you're working with `None` statically typed as `None.type` (which is pretty uncommon in the first place), we assume you want a mapping to `null`.

easel:
I think that works. It seems pretty spurious to demand that `None.type` encode to empty, if you're operating at that level you should be able to drop the key yourself.

travisbrown:
Just to clarify one point: this PR doesn't change the behavior of working with `Option` in any way, including `None`s that are statically typed as `Option[A]` (rather than `None.type`).

Issue #26, opened 2015-08-11
Add support for convenient non-generic codec definition


travisbrown:
circe currently doesn't include anything like Argonaut's `jdecodeNL`, `casecodecN`, etc.—you can define instances entirely by hand, semi-automatically with `io.circe.generic.semiauto._`, or fully automatically with `io.circe.generic.auto._`, but that's it.

There's currently no arity-level boilerplate in the `core` project, and I'm not sure how I want to handle this, but I want to provide something comparable to `casecodecN` in 0.3.0.

travisbrown:
This will be easier now that the basic machinery for boilerplate is set up in #42.

travisbrown:
Added in #237.

PR #237, opened 2016-03-17
Add forProduct helper methods

travisbrown:
This PR finally addresses #26 (currently our oldest open issue).

What's new here is some code-gen-ed helper methods that are similar to Argonaut's `jdecodeNL` and `jencodeNL`. These methods (named `forProductN` here) support relatively convenient codec definitions for case class-like types:

```scala
import io.circe._, io.circe.jawn._, io.circe.syntax._

case class User(id: Long, firstName: String, lastName: String)

object User {
  implicit val decodeUser: Decoder[User] =
    Decoder.forProduct3("id", "first_name", "last_name")(User.apply)

  implicit val encodeUser: Encoder[User] =
    Encoder.forProduct3("id", "first_name", "last_name")(u =>
      (u.id, u.firstName, u.lastName)
    )
}
```

And then:

```scala
scala> val json = User(101, "Foo", "McBar").asJson
json: io.circe.Json =
{
  "id" : 101,
  "first_name" : "Foo",
  "last_name" : "McBar"
}

scala> decode[User](json.noSpaces)
res1: cats.data.Xor[io.circe.Error,User] = Right(User(101,Foo,McBar))
```

It's not as clean or as maintainable as generic derivation, but it's less magical and requires nothing but circe-core.

This PR still needs tests, and I'd appreciate feedback on the `forProductN` name (or any other details) before this gets merged.

### DIFF
--- a/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -172,9 +172,12 @@ trait Decoder[A] extends Serializable { self =>
  * @groupname Tuple Tuple instances
  * @groupprio Tuple 5
  *
+ * @groupname Product Case class and other product instances
+ * @groupprio Product 6
+ *
  * @author Travis Brown
  */
-final object Decoder extends TupleDecoders with LowPriorityDecoders {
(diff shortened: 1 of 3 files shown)


lvicentesanchez:
:+1: I was missing this Argonaut feature :)

PR #164, opened 2016-01-14
Generic derivation overhaul (with configurability)

travisbrown:
Supersedes #163 and #79. Also fixes #69, although I'm not entirely sure why. Still needs tests before it gets merged.

PR #242, opened 2016-03-20
Replace export-hook

travisbrown:
This PR is a proof-of-concept that removes the dependency on [export-hook](https://github.com/milessabin/export-hook) and replaces it with a simpler (but similar) mechanism.

I decided to try this for two reasons. The first is that I don't like the macro-generated method names that `@imports` introduces into the `Encoder` and `Decoder` companion objects (`fresh$macro$1` in the case of `Decoder` and `fresh$macro$4` for `Encoder`). These seem like a binary compatibility-breaking liability, so I decided to replace them with non-macro versions, and it turned out that once I'd done that export-hook wasn't providing all that much additional value.

The second reason is that as I've been trying to wrap up #164, I keep running into export-hook limitations (most along the lines of https://github.com/milessabin/export-hook/issues/24). If it turns out that this approach makes it easier to finish up one of my experiments for #164, I'll be pretty tempted to merge it.

It might also be possible to do this in an even simpler way (as @alexarchambault notes [on Gitter](https://gitter.im/travisbrown/circe?at=56ef004a8b806f6b7a18cfb9)).

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,9 +92,8 @@ circe is a fork of Argonaut with a few important differences.
 
 ### Dependencies and modularity
 
-circe depends on [cats][cats] instead of [Scalaz][scalaz], and the `core` project has only two
-dependencies: cats-core and [export-hook][export-hook] (a lightweight mechanism for cleaner generic
-type class instance derivation).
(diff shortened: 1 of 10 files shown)


travisbrown:
Superseded by #247.

PR #249, opened 2016-03-28
Fix method name in ScalaDoc

fthomas:


travisbrown:
Thanks @fthomas! We really need to get sbt-doctest running on these eventually.

fthomas:
Agreed. Adding sbt-doctest would have been my first option if I had more time right now.

PR #261, opened 2016-04-10
fixes #259: use node.js to build scala.js in Travis

### LINKED ISSUE #259: Modernize Scala.js build
@ngbinh has done [some work](https://github.com/typelevel/cats/pull/954) on the Scala.js build for Cats that looks like it'd be valuable here as well (maybe we could even have Travis CI running the Scala.js builds!) but I know nothing about Scala.js build config and don't want to just cargo cult it.

If someone who knows / cares about Scala.js wants to help, that'd be awesome.

ngbinh:


travisbrown:
Thanks, @ngbinh!

One strange thing—I pushed your commit to my `scalajs` branch, which is specially configured to run the Scala.js tests on Travis CI (by default they don't run), and now it's showing up here as a push build, which is still failing because of timeouts.

Are these settings reasonable to have in place even if we're not running Scala.js builds on Travis CI right now? Can I expect them to have an impact (either way) on how long and how much memory it takes to run the Scala.js tests locally?

ngbinh:
Let me take a closer look at `JS` tests to make sure there are not any crazy thing I missed.

This PR should make local build a bit faster because it uses `nodeJsEnv` instead of `Rhino`

ngbinh:
sorry for the delay. I was pretty swarm at work so will take a look at this some time next week.

travisbrown:
I believe this is the default for Scala.js now. Feel free to re-open if the changes here are still relevant.

Issue #224, opened 2016-03-08
Consider providing a history-to-selection-path printing helper


travisbrown:
In #222 @tobias-johansson added a `Show` instance for `DecodingFailure` that provides a handy path representation of the history:

``` scala
DecodingFailure at .bar.baz.qux.a: String
DecodingFailure at .bar.baz.qux.b: Boolean
DecodingFailure at .bar.baz.i[2]: Int
```

I was just wondering whether it might make sense to factor out the path-printing part to something like a `def opsToPath(history: List[HistoryOp]): String` method on the `HistoryOp` companion object that would be usable in other contexts. For example, I could imagine Finch users wanting to prepare a custom error message from a `DecodingFailure` that would use the nice path part without the `DecodingFailure at…` context.

What do people think?

tobias-johansson:
@travisbrown I think breaking this out is a good idea.
Originally I created a `Show[List[HistoryOp]]` in the `HistoryOp` companion object, but it felt a bit weird defining a Show for a specific type parametrization of List, and also potentially confusing having both that and the default `Show[List[A]]` from Cats available.
So a function `def opsToPath(history: List[HistoryOp]): String` sounds good

hamnis:
Closing in cleanup run, If anyone cares about this, please comment and I'll reopen.

Issue #265, opened 2016-04-11
Fix Arbitrary instances for java.time types


travisbrown:
In the circe-java8 package, the `Arbitrary` instances for types in `java.time` are derived from ScalaCheck's provided instance for `java.util.Date`, which results in nonsensical values like `-26558675-11-05T14:56:40.198Z[GMT0]` that are causing tests to fail occasionally.

I'm not too worried about this since it seems fairly rare and doesn't represent a real problem, but we should have better `java.time` instances.

PR #267, opened 2016-04-12
Move ops-to-path printing into HistoryOp

tobias-johansson:
Ref: #224

### DIFF
--- a/core/shared/src/main/scala/io/circe/Error.scala
+++ b/core/shared/src/main/scala/io/circe/Error.scala
@@ -4,7 +4,6 @@ import algebra.Eq
 import cats.Show
 import cats.data.NonEmptyList
 import cats.std.list._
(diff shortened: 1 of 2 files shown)


tobias-johansson:
Um, build failure seems unrelated to my changes...?

travisbrown:
Oh, damn—yeah, that's #265, which I reported yesterday, after seeing it once out of a few dozen jobs. I was hoping it wouldn't pop up again so soon. I'll restart the job…

Thanks for doing this! I probably won't be able to review until this afternoon, but will get it in asap.

tobias-johansson:
Hey, no rush

travisbrown:
Thanks, @tobias-johansson! I think we should probably add some tests (now that this is a little more easily testable), but I don't think that should block merging this.

PR #268, opened 2016-04-12
Added extra parameter of type LocalDateTime to showcase its encoding-decoding

rahulsinghai:
Added extra parameter of type LocalDateTime to showcase its encoding-decoding


travisbrown:
Nice, thanks! Maybe it would make sense to use the `java.time` instances already defined in circe-java8, though?

If we're making this example Java 8-specific (which I'm on board with), we should also add a note making that clear, since circe in general still supports Java 7.

travisbrown:
Thanks again, @rahulsinghai! This is included in #276.

PR #284, opened 2016-05-05
info on macro paradise plugin inclusion updated in the README

Mullefa:


travisbrown:
Thanks, @Mullefa! Sorry for missing this until now—I've been traveling. One question: can we change this to "or if you need macro annotation support for `@JsonCodec`"? That's the only feature that uses macro annotations, but I don't think we should assume people know that.

Mullefa:
@travisbrown good point - I've edited the README accordingly.

travisbrown:
Thanks!

PR #289, opened 2016-05-17
Update to Cats 0.6.0

travisbrown:
This isn't ready to merge (we should probably wait for Cats 0.6.0 final), but I wanted to have a place to track the update.


travisbrown:
I've just rebased this and I'm planning to go ahead and merge it after CI says that's okay. It depends on a couple of snapshot versions, but they're for libraries I'm planning to release soon, and on the latest Cats 0.6.0 milestone, but the 0.6.0 release is also in the works.

I'll publish a milestone if somehow circe 0.5.0 is ready before Cats 0.6.0 final is out.

PR #290, opened 2016-05-18
compile java8 for JS

reactormonk:
There's a scalajs implementation for java.time.


reactormonk:
Not all the required classes are implemented.

PR #293, opened 2016-05-20
fix the order of JsonObject.traverse

julien-truffaut:


travisbrown:
Thanks @julien-truffaut!

PR #299, opened 2016-05-30
Improve macro hygiene

### LINKED ISSUE #298: Automatic encoders fail when compiling with -Yno-predef
Decoders, however work fine. 

Work-around is importing Predef.ArrowAssoc wherever you need the encoder implicit to be available.

```
scala> import io.circe._, io.circe.generic.auto._, io.circe.parser._, io.circe.syntax._
import io.circe._
import io.circe.parser._
import io.circe.syntax._

scala> case class Foo(s: String)
defined class Foo

scala> val dec = Decoder[Foo]
dec: io.circe.Decoder[Foo] = io.circe.generic.decoding.DerivedDecoder$$anon$1@7be4f5bf

scala> val enc = Encoder[Foo]
<console>:25: error: could not find implicit value for parameter instance: io.circe.Encoder[Foo]
       val enc = Encoder[Foo]
                        ^

scala> import Predef.ArrowAssoc
import Predef.ArrowAssoc

scala> val enc = Encoder[Foo]
enc: io.circe.Encoder[Foo] = io.circe.generic.encoding.DerivedObjectEncoder$$anon$1@3e6b2787
```

fthomas:
Probably fixes #298.

### DIFF
--- a/generic/shared/src/main/scala/io/circe/generic/DerivationMacros.scala
+++ b/generic/shared/src/main/scala/io/circe/generic/DerivationMacros.scala
@@ -213,7 +213,7 @@ class DerivationMacros(val c: whitebox.Context) {
         case (Member(name, _, tpe, _), instanceName, (patternAcc, fieldsAcc)) =>
         val currentName = TermName(c.freshName)
 
-        (pq"_root_.shapeless.::($currentName, $patternAcc)", q"$name -> $instanceName.apply($currentName)" :: fieldsAcc)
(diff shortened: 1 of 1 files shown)


fthomas:
Ok, this definitely fixes #298. The following compiles on this branch but not on master: 
```scala
scala> import Predef.{ArrowAssoc => _}
import Predef.{ArrowAssoc=>_}

scala> case class Foo(s: String)
defined class Foo

scala> val enc = Encoder[Foo]
enc: io.circe.Encoder[Foo] = io.circe.generic.encoding.DerivedObjectEncoder$$anon$1@11bdde4b
```

travisbrown:
Thanks again, @fthomas!

PR #300, opened 2016-05-30
Add hygiene compilation test sub-project

travisbrown:
This PR adds a new `hygiene` sub-project that provides tests for #298 (fixed by @fthomas in #299). I'm made `hygiene` a separate project so that we don't have to turn on `-Yno-predef` for the `test` project generally. The fake definitions for common type and package names were suggested by @retronym [here](https://github.com/scala/pickling/issues/329).

I've confirmed that both `Encoder` derivation lines fail without the fix in #299.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ script:
     then sbt ++$TRAVIS_SCALA_VERSION clean validateJS;
     else
       sbt ++$TRAVIS_SCALA_VERSION clean coverage validateJVM &&
+      sbt hygiene/compile &&
       sbt ++$TRAVIS_SCALA_VERSION coverageReport && codecov;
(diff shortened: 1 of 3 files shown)


fthomas:
The build error on 2.10 is interesting. Does that mean `Lazy` has hygiene issues on 2.10?

travisbrown:
@fthomas Yeah, I just spent half an hour investigating that and have no idea (I've also asked on [Gitter](https://gitter.im/milessabin/shapeless?at=5766f943f0528c4c5bba816d)). For now I'm just not going to run the compilation tests on 2.10.

fthomas:
I've also looked for obvious hygiene issues in shapeless, but didn't find anything.

Issue #133, opened 2015-12-01
Add a type class combining encoding and decoding


travisbrown:
Today [on Gitter](https://gitter.im/travisbrown/circe?at=565e013c5993bcb005d2cd9a) @julienrf asked the following question:

> @travisbrown can you remind me why you decided to remove `CodecJson`?

And [my response](https://gitter.im/travisbrown/circe?at=565e062122df37d14f936065):

> @julienrf Primarily because I think the role it plays in Argonaut is confusing. It can be handy for definitions, but you generally don't want to use it as a context bound anywhere except tests, since there aren't `CodecJson` instances for lots of things that have `DecodeJson` and `EncodeJson` instances…
> @julienrf …and that's the case because you can't define e.g. `CodecJson[String]` in the `CodecJson` companion object, since then it wouldn't be found when you ask for `DecodeJson[String]`.
> @julienrf So you could either put all your `CodecJson` instances for these basic types in some object that you expect users to import (ugh), or have some kind of implicit that automatically combines `DecodeJson` and `EncodeJson` into a `CodecJson` (which was removed in Argonaut for reasons I don't exactly remember, although I can imagine how that might get messy), or do what Argonaut does and just provide `CodecJson` for convenient definitions and let users figure out why it's pretty much useless as a requirement.
> None of those options are very nice, so at the beginning I decided to leave it out of circe entirely.
> It's possible that it's possible to do it right (or at least dramatically better) with export-hook, and I'd definitely be open to that possibility.

My wishlist for a `Codec` type class in circe would look something like this:
1. A `Codec[A]` should be available (without any imports) for any `A` that has a `Decoder` and `Encoder`—i.e. if `io.circe.Decoder[A]` and `io.circe.Encoder[A]` compile, then `io.circe.Codec[A]` must compile as well.
2. All tests should pass as currently written.
3. If a type has a `Codec` instance, then asking for a `Decoder` (or `Encoder`) for that type shouldn't require additional allocations.
4. We shouldn't have to change the name of the `apply` methods on `Decoder` or `Encoder`.
5. It should be possible to define `Codec` instances for standard library and circe types in the `Codec` companion object and have `Encoder` and `Decoder` instances available with no imports.

Only the first two are hard requirements, and the third and fourth are probably incompatible.

tpolecat:
FWIW I just went through this with doobie (and had a good discussion with @mpilquist who went through a similar exercise with scodec). My approach is:

[...]

PR #301, opened 2016-06-11
Add Codec type class (#133)

OlivierBlanvillain:
This implementation follows what is done in https://github.com/jto/validation with [`Format`](https://github.com/jto/validation/blob/master/validation-core/src/main/scala/play/api/data/mapping/Format.scala#L7). I think is fulfills the first 4 requirements listed in #133, 5. does not quite work unfortunately: 96a57230a903bc7778aed49e3bbaf32fc7ec78d0.

If you think that's still acceptable I would be happy to continue the PR with tests & co :)

### DIFF
--- a/core/shared/src/main/scala/io/circe/Codec.scala
+++ b/core/shared/src/main/scala/io/circe/Codec.scala
@@ -1,0 +1,17 @@
+import io.circe._
+
+/**
+ * A type class that provides back and forth conversion between values of type `A`
+ * and the [[Json]] format. Must obey the laws defined in [[io.circe.tests.CodecLaws]].
+ */
+trait Codec[A] extends Encoder[A] with Decoder[A]
+
+object Codec {
+  def apply[A](implicit instance: Codec[A]): Codec[A] = instance
+
+  implicit def fromEncoderDecoder[A](implicit e: Encoder[A], d: Decoder[A]): Codec[A] =
+    new Codec[A] {
+      def apply(c: HCursor): Decoder.Result[A] = d(c)
+      def apply(a: A): Json = e(a)
+    }
+}
--- a/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -8,6 +8,9 @@ import java.util.UUID
 import scala.collection.generic.CanBuildFrom
 import scala.collection.mutable.Builder
 
+/**
+ * A type class that provides a way to materialize a value of type `A` from a [[Json]] value.
+ */
 trait Decoder[A] extends Serializable { self =>
   /**
    * Decode the given hcursor.


julienrf:
@travisbrown Any thought on this PR?

travisbrown:
Sorry for the long delay in responding, @OlivierBlanvillain and @julienrf. I'm still not 100% sure I like this, but there does seem to be demand for it, and would like to make a decision before 0.7.0.

I probably won't have time to play around with it until later this week, but will do my best to follow up in more detail before next week.

rcavalcanti:
Wouldn't it be much simpler to have just the following?

```scala
  type Codec[A] = Encoder[A] with Decoder[A]

  def deriveCodec[A](implicit encode: Lazy[DerivedObjectEncoder[A]], 
                     decode: Lazy[DerivedDecoder[A]]): Codec[A] =
    new Encoder[A] with Decoder[A] {
      val enc = deriveEncoder[A]
      val dec = deriveDecoder[A]
      def apply(a: A): Json = enc(a)
      def apply(c: HCursor): Result[A] = dec(c)
    }

// and then just derive both Encoder and Decoder altogether
implicit val foo: Codec[Foo]  = deriveCodec
```

Codec is an Encoder and a Decoder and it's generate in one shot. Nothing more than that.

And this technique could be applied to any other variable, like: 

```scala
// for enumerations
def deriveEnumCodec[A](implicit encode: Lazy[EnumerationEncoder[A]], 
                       decode: Lazy[EnumerationDecoder[A]]): Codec[A] =
    new Encoder[A] with Decoder[A] {
      val enc = deriveEnumerationEncoder[A]
      val dec = deriveEnumerationDecoder[A]
      def apply(a: A): Json = enc(a)
      def apply(c: HCursor): Result[A] = dec(c)
    }
```

```scala
// for configured encoders and decoders
def deriveCodec[A](implicit encode: Lazy[ConfiguredObjectEncoder[A]], 
                   decode: Lazy[ConfiguredDecoder[A]]): Codec[A] =
    new Encoder[A] with Decoder[A] {
      val enc = deriveEncoder[A]
      val dec = deriveDecoder[A]
      def apply(a: A): Json = enc(a)
      def apply(c: HCursor): Result[A] = dec(c)
    }
```

core/shared/src/main/scala/io/circe/Codec.scala
```diff
@@ -0,0 +1,17 @@
+import io.circe._
+
+/**
+ * A type class that provides back and forth conversion between values of type `A`
+ * and the [[Json]] format. Must obey the laws defined in [[io.circe.tests.CodecLaws]].
+ */
+trait Codec[A] extends Encoder[A] with Decoder[A]
```
kevinmeredith: Why does `Codec` extend `Encoder` and `Decoder`?

OlivierBlanvillain: > A type class that provides back and forth conversion between values of type `A` and the [[Json]] format.

A => Json is `Encoder` and Json => Option[A] is `Decode`.

travisbrown:
Included in #1151.

PR #303, opened 2016-06-14
Add some utilities to Decoder

johnynek:
As I was working with circe, I hit two annoying pain points:

1) I didn't see an easy way to make a constant decoder, which is useful in many combinator applications.
2) a lot of existing code returns Try, and each time we work with it we need to jump through a bunch of hoops.

### DIFF
--- a/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -7,6 +7,7 @@ import io.circe.export.Exported
 import java.util.UUID
 import scala.collection.generic.CanBuildFrom
 import scala.collection.mutable.Builder
+import scala.util.{Failure, Success, Try}
 
 trait Decoder[A] extends Serializable { self =>
   /**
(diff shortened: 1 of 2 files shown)


travisbrown:
Thanks, @johnynek! I'll review in more detail this evening, but one quick note: right now I'd consider the following the "idiomatic" way to create a constant decoder:

```scala
scala> import cats.syntax.applicative._
import cats.syntax.applicative._

scala> import io.circe.Decoder
import io.circe.Decoder

scala> "foo".pure[Decoder]
res0: io.circe.Decoder[String] = io.circe.Decoder$$anon$1$$anon$36@4dcad16c
```

Or an equivalent (`Applicative[Decoder].pure("foo")`). But I think it'd be reasonable to add `const` and write `pure` in terms of it.

johnynek:
I hear you on the above, but I am trying to help cats and circe get adoption, and honestly there is a lot of pushback on any implicit tricks. As much as people can use methods on the companions and maybe only get the encoders and decoders implicitly, that will help.

travisbrown:
@johnynek Thanks again—I'm going to merge this now but will make a few small changes (like defining `pure` in terms of `const`) and add tests before the 0.5.0 release.

johnynek:
Thank you sir (sorry for lack of tests... that was lame of me).

PR #308, opened 2016-06-19
Minor changes and tests for #303

travisbrown:
@johnynek I renamed `trymap` to `emapTry` to emphasize that it does more or less the same thing as `emap`.

### DIFF
--- a/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -7,7 +7,7 @@ import io.circe.export.Exported
 import java.util.UUID
 import scala.collection.generic.CanBuildFrom
 import scala.collection.mutable.Builder
-import scala.util.{Failure, Success, Try}
+import scala.util.{ Failure, Success, Try }
 
 trait Decoder[A] extends Serializable { self =>
   /**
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/io/circe/Decoder.scala
```diff
@@ -771,10 +775,7 @@ final object Decoder extends TupleDecoders with ProductDecoders with LowPriority
   implicit final val monadInstances: SemigroupK[Decoder] with MonadError[Decoder, DecodingFailure] =
     new SemigroupK[Decoder] with MonadError[Decoder, DecodingFailure] {
       final def combineK[A](x: Decoder[A], y: Decoder[A]): Decoder[A] = x.or(y)
-      final def pure[A](a: A): Decoder[A] = new Decoder[A] {
-        final def apply(c: HCursor): Result[A] = Xor.right(a)
-        override def decodeAccumulating(c: HCursor): AccumulatingDecoder.Result[A] = Validated.valid(a)
-      }
+      final def pure[A](a: A): Decoder[A] = const(a)
```
johnynek: you could add `pureEval` right, and only evaluate on `apply`? That laziness might be useful for some folks.

core/shared/src/main/scala/io/circe/Decoder.scala
```diff
@@ -158,7 +158,7 @@ trait Decoder[A] extends Serializable { self =>
    *
    * @param f a function returning either a value or an error message
    */
-  final def trymap[B](f: A => Try[B]): Decoder[B] = new Decoder[B] {
+  final def emapTry[B](f: A => Try[B]): Decoder[B] = new Decoder[B] {
```
johnynek: nice. I had an inconsistency between `trymap` and `instanceTry` with capitalization. This is better.

core/shared/src/main/scala/io/circe/Decoder.scala
travisbrown: @johnynek Yeah, I guess we should do that…

PR #310, opened 2016-06-19
Add pureEval implementation for Decoder

travisbrown:
As [suggested](https://github.com/travisbrown/circe/pull/308#discussion_r67623059) by @johnynek.

### DIFF
--- a/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -1,6 +1,6 @@
 package io.circe
 
-import cats.{ MonadError, SemigroupK }
+import cats.{ Eval, MonadError, SemigroupK }
 import cats.data.{ Kleisli, NonEmptyList, OneAnd, Validated, Xor }
 import cats.std.list._
 import io.circe.export.Exported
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/io/circe/Decoder.scala
```diff
@@ -776,6 +776,11 @@ final object Decoder extends TupleDecoders with ProductDecoders with LowPriority
     new SemigroupK[Decoder] with MonadError[Decoder, DecodingFailure] {
       final def combineK[A](x: Decoder[A], y: Decoder[A]): Decoder[A] = x.or(y)
       final def pure[A](a: A): Decoder[A] = const(a)
+      final override def pureEval[A](a: Eval[A]): Decoder[A] = new Decoder[A] {
+        final def apply(c: HCursor): Result[A] = Xor.right(a.value)
```
johnynek: actually, I guess `a.value` can throw if execution throws an exception. Should this throw or should we use `Xor.catchNonFatal` (or whatever it is).

travisbrown: @johnynek I'm inclined to think that if there's a possibility that the expression you're pure-ing might not evaluate to anything, that's your responsibility, whether you're using `pure` or `pureEval` (and you should be using `instance` or whatever instead).

johnynek: Sounds good, (or make an `Eval[Try[A]]` or something if you have an `Eval` that can throw).

PR #314, opened 2016-06-20
Propose TabMo in adopters list

studiodev:
We (at TabMo) are using Circe in many of your projects, and we are very satisfied of it.

The main use case is fast parsing in real-time applications (Akka-Streams & Spark Jobs). We are also using Circe with Akka HTTP backends.


travisbrown:
Thanks @studiodev!

travisbrown:
Are you using akka-http-json?

studiodev:
Yes, we are using `akka-http-json` in our Akka HTTP backends (using both high-level and low-level API).

Our first experiments are here https://github.com/studiodev/microservices-benchmark-scala/tree/reboot

PR #315, opened 2016-06-21
Avoid boxing on Double operations

johnynek:
just noticed this and figured I'd fix it.


travisbrown:
Thanks for catching this!

PR #316, opened 2016-06-22
add latin pronunciation to readme

joprice:


joprice:
No problem. Just trying to equally represent all dead languages.

PR #322, opened 2016-07-03
Add http4s as a community project

bryce-anderson:


travisbrown:
Thanks @bryce-anderson!

PR #325, opened 2016-07-08
Added encoder and decoder for Scala Enumeration

### LINKED ISSUE #297: Case class with Enumeration is not converted
I'm trying to convert a case class that has a Enumeration as a argument

```scala
import io.circe._, io.circe.generic.auto._, io.circe.jawn._, io.circe.syntax._
object WeekDay extends Enumeration {
  val Mon, Tue, Wed, Thu, Fri, Sat, Sun = Value
}
case class test1(day: WeekDay.Value)
case class test2(day: String)
```

While test2("Mon").asJson works as expected:
```json
{
  "day" : "Mon"
}
```

But: test1(WeekDay.Mon).asJson results in:
> <console>:27: error: could not find implicit value for parameter encoder: io.circe.Encoder[test]
>       test(WeekDay.Mon).asJson

Is there a way to work around this or is this a bug?

mariussoutier:
Added an encoder and a decoder for Scala Enumerations.

Parsing an enumeration value that doesn't exist will result in a `DecodingFailure` that contains a `NoSuchElementException`.

To summarize why:
- In some cases they are actually useful (convenience methods for example)
- Sometimes you don't have control over how the domain model looks
- They are a part of the Scala standard lib

Fixes #297

### DIFF
--- a/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -788,6 +788,17 @@ final object Decoder extends TupleDecoders with ProductDecoders with LowPriority
       final def raiseError[A](e: DecodingFailure): Decoder[A] = Decoder.failed(e)
       final def handleErrorWith[A](fa: Decoder[A])(f: DecodingFailure => Decoder[A]): Decoder[A] = fa.handleErrorWith(f)
     }
+
+  /**
+    * @group Enumeration
+    * {{{
+    *   object WeekDay extends Enumeration { ... }
+    *   implicit val weekDayDecoder = Decoder.enumDecoder(WeekDay)
+    * }}}
+    */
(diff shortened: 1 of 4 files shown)


travisbrown:
Thanks, @mariussoutier. The one change I think we should make is using the `Xor` to represent parsing failures instead of throwing an exception (which should just take `flatMap`-ing on the string decoder and using either `try` or `Xor.catchOnly`).

jpablo:
This is what I'm doing:
```scala
  def enumDecoder[E <: Enumeration](enum: E) = Decoder.instance { hc =>
    Try(hc.as[String] map (enum.withName(_))) match {
      case Failure(ex) => Xor.left(DecodingFailure(ex.getMessage, hc.history))
      case Success(eVal) => eVal
    }
  }
```

mariussoutier:
Alright updated the PR.

PR #328, opened 2016-07-15
Update README: correct reference to circe-scalajs

damirv:
I believe the `parser` subproject does not exist (anymore)?


travisbrown:
circe-parser is still available—if you check Maven Central you'll find 0.5.0-M2 releases for both the JVM and Scala.js. In many cases you'll want to use circe-jawn or circe-scalajs directly, but if you want 100% shared code and only need the lowest common denominator of the two, circe-parser works.

damirv:
Ah okay, clear, thanks!

PR #335, opened 2016-07-29
Add Context.fold method

durban:
A `Context` is either an `ObjectContext` or an `ArrayContext`. However, distinguishing them by pattern matching is not possible, since these subclasses are private. The existing `field` and `index` methods give access to the data in these subclasses, but they both return `Option`s: exactly one of these will be `Some`, but this is not visible in the types.

This PR adds a `fold` method to `Context`, which makes it possible to process `Context`s in a pattern match-like way, without resorting to `Option.get` or similar things.

### DIFF
--- a/core/shared/src/main/scala/io/circe/Context.scala
+++ b/core/shared/src/main/scala/io/circe/Context.scala
@@ -4,9 +4,9 @@ import cats.{ Eq, Show }
 
 sealed abstract class Context extends Serializable {
   def json: Json
+  def field: Option[String]
+  def index: Option[Int]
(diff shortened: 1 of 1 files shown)


travisbrown:
Thanks, @durban, this looks reasonable. My one concern with the implementation is that defining `field` and `index` in terms of `fold` now creates a couple of function objects every time either is used. This probably doesn't matter much in most situations, but since leaving the old `field` and `index` definitions is trivial I think I'd prefer to see us go that route.

durban:
Fair enough, I've pushed a commit restoring the old definitions.

PR #338, opened 2016-08-02
Add getOrElse to *Cursor.

tixxit:
I've found myself often saying, essentially, `cursor.get[Option[A]](k).map(_.getOrElse(default))`. This just adds a nice convenience method to do this instead.

We could put the default implementation on `GenericCursor` instead, but I saw that the trait was empty and had assumed there were reasons for this, so I just duplicated the method on each cursor for now.

### DIFF
--- a/core/shared/src/main/scala/io/circe/GenericCursor.scala
+++ b/core/shared/src/main/scala/io/circe/GenericCursor.scala
@@ -1,6 +1,7 @@
 package io.circe
 
 import cats.Functor
+import cats.data.Xor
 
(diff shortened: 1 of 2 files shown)


travisbrown:
Thanks, @tixxit, this looks like a reasonable addition to me.

In this case since `getOrElse` can be written in terms of `get`, the implementation probably should go in the base class.

tixxit:
OK, great! I've updated the PR and put the implementation in `GenericCursor`.

PR #340, opened 2016-08-05
Optimization for sequence and map decoders (do not merge)

travisbrown:
The changes here have two motivations: to clean up `Decoder.scala` by defining helper classes for map and sequence decoders (which also allows us to avoid duplicating some logic), and to improve performance by replacing a bunch of unpleasant functional code with unpleasant imperative code.

The API and behavior are unchanged (apart from one minor detail about the order of errors when using error accumulation with maps, in which case the previous behavior was arguably less intuitive), and none of the newly introduced mutability or imperativity faces the outside world.

The performance gains surprised me a bit—throughput is around 35% higher for our map-focused benchmark, and 8% for arrays of integers:

```
decodeFoosC     thrpt   20   3312.674 ±   27.015  ops/s (before)
decodeFoosC     thrpt   20   4513.516 ±  163.341  ops/s (after)

decodeIntsC     thrpt   20  19639.295 ±  164.289  ops/s (before)
decodeIntsC     thrpt   20  21266.008 ±  109.671  ops/s (after)
```

And allocation rates are around 25% lower:

```
decodeFoosC:gc.alloc.rate.norm  thrpt  5  1193152.347 ±   0.005  B/op (before)
decodeFoosC:gc.alloc.rate.norm  thrpt  5   888800.267 ±   0.003  B/op (after)

decodeIntsC:gc.alloc.rate.norm  thrpt  5   275359.804 ±  61.666  B/op (before)
decodeIntsC:gc.alloc.rate.norm  thrpt  5   203237.521 ±  47.057  B/op (after)
```

Most of these gains are due to the fact that we're not constructing extra `Xor` or `Option` values for every element we successfully decode.

Note that I'm deprecating `traverseDecode` and `traverseDecodeAccumulating` on `HCursor`, but I may remove them entirely before the 0.5.0 release—I don't think anyone else is likely to be using them.

This PR depends on #337—I'll rebase it and get rid of the snapshot dependencies once the new versions of Cats and iteratee.io are published.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -21,7 +21,7 @@ lazy val compilerOptions = Seq(
   "-Xfuture"
 )
 
-lazy val catsVersion = "0.6.1"
+lazy val catsVersion = "0.7.0-SNAPSHOT"
 lazy val jawnVersion = "0.8.4"
 lazy val shapelessVersion = "2.3.1"
 lazy val refinedVersion = "0.5.0"
@@ -371,7 +371,7 @@ lazy val streaming = project
   )
   .settings(allSettings)
   .settings(
-    libraryDependencies += "io.iteratee" %% "iteratee-core" % "0.5.0"
+    libraryDependencies += "io.iteratee" %% "iteratee-core" % "0.6.0-SNAPSHOT"
   )
   .dependsOn(core, jawn)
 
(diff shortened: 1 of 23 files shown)


build.sbt
```diff
@@ -21,7 +21,7 @@ lazy val compilerOptions = Seq(
   "-Xfuture"
 )
 
-lazy val catsVersion = "0.6.1"
+lazy val catsVersion = "0.7.0-SNAPSHOT"
```
viktorklang: Deps on snapshot?

travisbrown: The two releases will be coordinated—this PR won't be merged until the Cats release is out.

(I mention that in the last paragraph of the note, but I guess I should add it to the title as well. :smile:)

travisbrown:
Superseded by #353.

PR #342, opened 2016-08-08
Add circe-scodec subproject

fthomas:
circe-scodec provides `Decoder` and `Encoder` instances for
scodec-bits' `BitVector` and `ByteVecor` using Base64 for
serialization.

Example:
```scala
scala> import io.circe._, io.circe.generic.auto._, io.circe.parser._,
io.circe.syntax._, io.circe.scodec._
import io.circe._
import io.circe.parser._
import io.circe.syntax._
import io.circe.scodec._

scala> import _root_.scodec.bits.ByteVector
import _root_.scodec.bits.ByteVector

scala> case class Foo(bytes: ByteVector)
defined class Foo

scala> val foo = Foo(ByteVector.view("Hello world!".getBytes))
foo: Foo = Foo(ByteVector(12 bytes, 0x48656c6c6f20776f726c6421))

scala> foo.asJson.noSpaces
res0: String = {"bytes":"SGVsbG8gd29ybGQh"}

scala> decode[Foo](res0)
res1: cats.data.Xor[io.circe.Error,Foo] = Right(Foo(ByteVector(12 bytes,
0x48656c6c6f20776f726c6421)))
```
/cc @mpilquist

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -80,6 +80,7 @@ def noDocProjects(sv: String): Seq[ProjectReference] = Seq[ProjectReference](
   numbersJS,
   parserJS,
   refinedJS,
+  scodecJS,
   tests,
   testsJS
 ) ++ (
@@ -109,6 +110,7 @@ lazy val aggregatedProjects: Seq[ProjectReference] = Seq[ProjectReference](
   literal, literalJS,
   refined, refinedJS,
   parser, parserJS,
+  scodec, scodecJS,
   tests, testsJS,
   jawn,
   jackson,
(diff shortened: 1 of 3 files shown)


travisbrown:
Thanks, @fthomas! I'm inclined to suggest naming the instances e.g. `encodeBitVector` for consistency (with e.g. the instances in `Encoder` and `Decoder`), but I don't feel too strongly about that.

fthomas:
Good idea, @travisbrown. I've renamed the instances accordingly.

PR #343, opened 2016-08-08
Replace Xor

travisbrown:
This is in response to the [proposal in Cats](https://github.com/typelevel/cats/issues/1192) to get rid of `Xor` now that the standard library's `Either` will be slightly less terrible in 2.12.

I'm still opposed to that proposal, but it seems like I'm the only one, so I wanted to put together a quick proof-of-concept of what this would look like in Cats. This PR still needs a lot of polish, but all JVM tests pass.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/NonEmptySeqDecoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/NonEmptySeqDecoder.scala
@@ -12,10 +12,13 @@ private[circe] abstract class NonEmptySeqDecoder[A, C[_], S](implicit
   final def apply(c: HCursor): Decoder.Result[S] = {
     val arr = c.downArray
 
-    decodeA.tryDecode(arr).flatMap { head =>
-      decodeCA.tryDecode(arr.delete).map { tail =>
-        create(head, tail)
-      }
+    decodeA.tryDecode(arr) match {
+      case Right(head) =>
+        decodeCA.tryDecode(arr.delete) match {
+          case Right(tail) => Right(create(head, tail))
+          case l @ Left(_) => l.asInstanceOf[Decoder.Result[S]]
+        }
+      case l @ Left(_) => l.asInstanceOf[Decoder.Result[S]]
     }
   }
 
--- a/modules/benchmark/src/main/scala/io/circe/benchmark/Benchmark.scala
+++ b/modules/benchmark/src/main/scala/io/circe/benchmark/Benchmark.scala
@@ -122,7 +122,7 @@ class EncodingBenchmark extends ExampleData {
 @OutputTimeUnit(TimeUnit.SECONDS)
 class DecodingBenchmark extends ExampleData {
   @Benchmark
-  def decodeIntsC: List[Int] = intsC.as[List[Int]].getOrElse(throw new Exception)
+  def decodeIntsC: List[Int] = intsC.as[List[Int]].right.getOrElse(throw new Exception)
 
   @Benchmark
   def decodeIntsA: List[Int] = intsA.as[List[Int]].result.getOrElse(throw new Exception)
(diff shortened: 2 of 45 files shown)


mpilquist:
@travisbrown Would this mean circe-core would lose its existing cats dependency?

mpilquist:
Ah no, disregard me. :)

travisbrown:
@mpilquist Right, there's enough other Cats stuff (especially `Validated`) that that would be impractical.

penland365:
I'm opposed to the other proposal as well, so I would vote for removing it iff it gets removed in cats. Based on the Scala 2.12 Either, I don't feel like it's adequate for what I want out of a `DecodingResult` either.

travisbrown:
As a side note, moving to `Either` and losing `map` and `flatMap` means you have to do one of the following:

* Call `.right` to get the right projection and then `map`, etc.
* Use Cats's functor and monad syntax to get `map`, etc. enrichment methods.
* Just do a bunch of pattern matching instead.

In this PR I've taken the last approach, which means more noise in the code, but it also gives a slight performance boost (e.g. normalized allocation rate is about 6% lower for the integer JSON array decoding benchmark).

rossabaker:
Are you avoiding defining `.map` and `.flatMap` syntax for performance reasons, SI-2712 reasons, or other reasons?  The inlining of those operations contributes much of the dreariness of this PR.

travisbrown:
@rossabaker The performance hit that comes with using Cats's current `map` and `flatMap` syntax for `Either` is [more than I want to pay](https://gist.github.com/travisbrown/cf9697d33bd4ed8fa473ab46d4942e31#gistcomment-1843314). Using `.right` everywhere might not be quite so bad, but it's deprecated in 2.12.

The inlining does have advantages (see my previous comment), isn't _that_ terrible, and fits with the "library code doesn't need to model usage" idea from the design docs.

hseeberger:
Go for `Either`.

rossabaker:
I added [benchmarks with EitherSyntax](https://gist.github.com/rossabaker/ae49d2716e47b2f2cb8542a6186dd638).  I'm unsure why my numbers are not more proportional to yours for the same tests.

I share some of your misgivings.  `Either` with syntax can perform as well as `Xor`, but the typeclass syntax is a subtle trap.  `EitherSyntax` wins when both are in scope, so a simple `import cats.implicits._` could do the right thing.  Except when you have have a `_[_]`-kinded alias, as does circe, and you hit SI-2712.  These problems have easy solutions, but they are awkward conversations to have with the new user who just wants to parse some JSON.

All that said, converting between `Xor` and `Either` because we couldn't all agree comes with its own runtime and cognitive overhead.  While a move to `Either` will irritate us library cross-builders for a long time, application developers get a good experience starting in 2.12.  `Either` is only going to look better over time, and increasing circe uptake is going to make this migration more expensive over time.  My crystal ball reluctantly puts me on Team Either.

adelbertc:
@travisbrown Have you benchmarked performance when `flatMap` and `map` are done through an implicit value class as opposed to through the `Monad` syntax?

modules/core/shared/src/main/scala/io/circe/NonEmptySeqDecoder.scala
```diff
@@ -12,10 +12,13 @@ private[circe] abstract class NonEmptySeqDecoder[A, C[_], S](implicit
...
 
-    decodeA.tryDecode(arr).flatMap { head =>
-      decodeCA.tryDecode(arr.delete).map { tail =>
-        create(head, tail)
-      }
+    decodeA.tryDecode(arr) match {
+      case Right(head) =>
+        decodeCA.tryDecode(arr.delete) match {
+          case Right(tail) => Right(create(head, tail))
+          case l @ Left(_) => l.asInstanceOf[Decoder.Result[S]]
```
adelbertc: We have syntax on `Left` and `Right` that gives a more type safe way of casting (the irony) https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/syntax/either.scala#L301

PR #354, opened 2016-08-20
Update README.md

aishfenton:
Added Argus link.


travisbrown:
NIce, thanks @aishfenton!

PR #355, opened 2016-08-29
Adding circe-akka-http project to provide circe support in akka-http

menzus:
Hi,

I'm using circe with akka-http and I thought it would be useful to have akka-http support in circe project. I've forked your project and created a circe-akka-http module. 

I've followed the spray project's JsonSupport interface, however I'm not using RootEncoder/RootDecoder. 

What do you think?

Regards,
Tamas

### DIFF
--- a/akkaHttp/README.md
+++ b/akkaHttp/README.md
@@ -1,0 +1,99 @@
+# circe-akka-http
+
+*circe-akka-http* provides support for using *circe* library with
+[akka-http](http://doc.akka.io/docs/akka/current/scala/http/).
+
+## Usage
+By importing `io.circe.akkahttp.JsonSupport._` or `io.circe.akkahttp.ErrorAccumulatingJsonSupport._`
+you can add circe support to your routes.
+
+## Example
+
+### Code
+
+```scala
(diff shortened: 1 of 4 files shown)


ktoso:
FYI the support trait exists already, and is maintained here: https://github.com/hseeberger/akka-http-json/blob/master/akka-http-circe/src/main/scala/de/heikoseeberger/akkahttpcirce/CirceSupport.scala
It's likely that we'll include it in the Akka HTTP repo soon.

So unless you want to maintain it here, I think we have a good place and are taking good care of that integration 😄

menzus:
Yeah, I'm aware of that project.
At my company we're using third project libs only if it's not "really" needed. For these few lines of code it was decided it's not needed :/ It's stupid, but we ended up to write it for ourself and maintain it.. 

I guess few other companies have policies like that, so I thought it would be useful for others.

But I get your point.

travisbrown:
Thanks, @menzus, but I'd prefer to keep effort focused on a single implementation (in this case akka-http-json). Mind if I close this PR?

menzus:
Sure, I've closed it!

PR #366, opened 2016-09-09
Bad searches fail faster during generic derivation

travisbrown:
I've been working on compilation time benchmarks for the past week, and one of the things I noticed was that while the #247 overhaul improves compile times significantly for small case classes / ADTs, it actually very rapidly gets worse than the non-macro version once you get to around twenty members in a case class (or about the same number of case classes in an ADT).

At least part of the issue is that because the materializers are whitebox macros, a search for an `HList` representation of a case class will invoke `decodeCoproduct` (and vice versa).

This PR essentially just enforces the constraint by immediately crashing the materialization if the type doesn't match, instead of letting it trigger other searches, etc.

I'll be looking at the performance differences briefly in my Scala World talk in a couple of days.

This is binary compatible, so I'll probably backport it to for 0.5.2 release in the upcoming week.

### DIFF
--- a/generic/shared/src/main/scala/io/circe/generic/DerivationMacros.scala
+++ b/generic/shared/src/main/scala/io/circe/generic/DerivationMacros.scala
@@ -62,6 +62,8 @@ class DerivationMacros(val c: whitebox.Context) {
     private[this] val KeyTagSym = typeOf[KeyTag[_, _]].typeSymbol
     private[this] val ShapelessTagType = typeOf[shapeless.tag.type]
     private[this] val ScalaSymbolType = typeOf[scala.Symbol]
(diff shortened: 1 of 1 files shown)


cb372:
I don't quite understand this. e.g. `def decodeHList` already has a bound `[R <: HList]`, so why do you need to check the subtype again in `fromHListType`?

travisbrown:
@cb372 You would think the constraints would be enough to keep these from being considered in the wrong cases, right? But for whatever reason they're not, and failing explicitly has a pretty big impact on compile time, especially for large types.

cb372:
Interesting! Well done for discovering that 🔍

PR #369, opened 2016-09-16
Add notes on how to encode and decode data

liff:
I felt this sort of explicit information about encoding/decoding was missing from the documentation.


travisbrown:
Thanks, @liff!

Issue #374, opened 2016-09-18
Validate invalid fields


olafurpg:
``` scala
@ import io.circe.generic.auto._
@ import io.circe.parser._
@ case class Foo(col: Option[Int])
@ val result = parse("""{ "typo": 2 }""").map(_.as[Foo])
result: cats.data.Xor[io.circe.ParsingFailure, io.circe.Decoder.Result[Foo]] = Right(Right(Foo(None)))
```

I'm curious to know, is there's a way for a decoder to validate that there's no invalid key like `typo` in this example? Phrased alternatively, is there a way to know after decoding whether there was an unused key/value in the json?

I'm hoping to use [circe-yaml](https://github.com/jeremyrsmith/circe-yaml) to read configuration for scalafmt. All the fields are optional (`Option[T]`) but I would like the decoder to fail if the user has entered a field that contains a typo.

I noticed there's a `validate` method on `Decoder[T]` but I don't see how it could help me in this scenario.

I'm using circe 0.5.2.

travisbrown:
By default decoders don't consume fields as they use them, so this isn't trivial, but you can do something that works with `validate` and some boilerplate:

```scala
import io.circe._, io.circe.generic.semiauto.deriveDecoder

def validateKeys(allowed: Set[String])(c: HCursor): Boolean =
  c.focus.asObject.exists(_.fieldSet.subsetOf(allowed))

case class Foo(a: Int, b: String, c: Boolean)

implicit val decodeFoo: Decoder[Foo] =
  deriveDecoder[Foo].validate(validateKeys(Set("a", "b", "c")), "Foo: invalid fields")
```

And then:

```scala
scala> import io.circe.jawn.decode
import io.circe.jawn.decode

scala> decode[Foo]("""{ "a": 1, "b": "xyz", "c": false, "d": null }""")
res0: cats.data.Xor[io.circe.Error,Foo] = Left(DecodingFailure(Foo: invalid fields, List()))
```

The error message isn't great, though, and I agree that this is a use case that deserves better support. I'll think about what this should look like this week (and any ideas would be appreciated).

olafurpg:
I was afraid this wouldn't be so trivial.

The `validate` workaround gets a lot tricker once you have 30 keys of which some are nested, contain arrays and so forth. Also, I think it's quite important that the error message points to the culprit field.

[...]

PR #375, opened 2016-09-19
StateT decoders

travisbrown:
This PR includes a couple of first steps toward the functionality requested by @olafurpg in #374. The goal is to support the construction of decoders that consume fields from JSON objects so that it's possible to have decoding fail if there are leftover fields.

For example:

```scala
import io.circe.Decoder, io.circe.jawn.decode
import cats.instances.either._

case class Foo(a: String, b: Int, c: Boolean)

implicit val decodeFoo: Decoder[Foo] = Decoder.fromState(
  for {
    a <- Decoder.state.decodeField[String]("a")
    b <- Decoder.state.decodeField[Int]("b")
    c <- Decoder.state.decodeField[Boolean]("c")
    _ <- Decoder.state.requireEmpty
  } yield Foo(a, b, c)
)
```

And then:

```scala
scala> import cats.syntax.show._
import cats.syntax.show._

scala> val good = """{ "a": "xyz", "b": 1, "c": false }"""
good: String = { "a": "xyz", "b": 1, "c": false }

scala> val bad = """{ "a": "xyz", "b": 1, "c": false, "d": null, "e": -1 }"""
bad: String = { "a": "xyz", "b": 1, "c": false, "d": null, "e": -1 }

scala> decode[Foo](good)
res9: Either[io.circe.Error,Foo] = Right(Foo(xyz,1,false))

scala> decode[Foo](bad).left.foreach(e => println(e.show))
DecodingFailure at .a{!_/}.b{!_/}.c{!_/}: Leftovers: d, e
res0: Any = ()
```

These methods still need documentation and tests, but I wanted to run the basic API past people. One alternative to the `State` suffixes would be to have a `state` object inside `Decoder` with `read`, `requireEmpty`, etc. methods, so that you could more easily import just these methods when you want to build decoders in this way.

This PR currently depends on #343—once that's merged I'll rebase.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -273,26 +273,14 @@ final object Decoder extends TupleDecoders with ProductDecoders with LowPriority
     final def apply(c: HCursor): Result[A] = f(c)
   }
 
+  /**
+   * Construct an instance from a [[cats.data.StateT]] value.
+   *
+   * @group Utilities
+   */
   def fromState[A](s: StateT[Result, ACursor, A]): Decoder[A] = new Decoder[A] {
     final def apply(c: HCursor): Result[A] = s.runA(c.acursor)
   }
-
-  def readState[A: Decoder](k: String): StateT[Result, ACursor, A] = StateT[Result, ACursor, A] { c =>
-    val field = c.downField(k)
-
-    field.as[A] match {
-      case Right(a) => Right((field.delete, a))
-      case l @ Left(_) => l.asInstanceOf[Result[(ACursor, A)]]
-    }
-  }
-
-  def requireEmptyState(createMessage: List[String] => String): StateT[Result, ACursor, Unit] =
-    StateT[Result, ACursor, Unit] { c =>
-      val fields = c.focus.flatMap(_.asObject).toList.flatMap(_.fields)
-
-      if (fields.isEmpty) Right((c, ())) else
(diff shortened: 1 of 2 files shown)


non:
@travisbrown This design seems OK to me, but I just wanted to propose an alternate way to think about this.

You could imagine taking an existing decoder and updating it with extra assertions that must be true. For example, maybe you read three optional keys but you want exactly two of them to be true (no more no less). Either before or after decoding you could check that the condition is valid, and if it isn't you'd fail the decoding.

Under this model instead of maintaining read state in this way, you'd just look for a key in the object that isn't in your set of allowed keys.

One advantage of this approach is that it makes certain kinds of invariants a bit easier to enforce. On the other hand it's also a bit fiddly. I just wanted to put it out there for you to consider.

andyscott:
Just wanted to mention that I'm very interested in using this. It will be particularly useful if we can get a generic derivation mechanism.

I may be able to lend a hand once a solution is chosen.

travisbrown:
@andyscott What do you think of the current API (setting aside generically derived instances for the moment)? I'd like to include this in 0.6.0, so any feedback would be much appreciated.

travisbrown:
@non Sorry about the long delay in responding, but I agree that that approach (validating the keys or other properties as a separate initial step) works well in many cases—it's what I've been recommending that people do in this situation up until this PR.

This PR isn't a replacement for that approach, but more an alternative for cases where it helps to have something more compositional (the way we currently build up generically derived instances, for example).

PR #376, opened 2016-09-20
Convert `JNull` to `null` instead of `undefined` in convertJsonToJs

jimmydivvy:
When using `convertJsonToJs` to convert a JSON structure into native JS, it currently converts `JNull` to `undefined`.

This differs from expected behaviour as used in other libraries and as defined in the ECMAScript standard, which specifies that `null` in json should be decoded to `null` [1]

This can be problematic when interacting with existing JS libraries that may be expecting certain values to be passed specifically as `null`, not `undefined. 

(I'm unsure if this is expected circe behaviour and/or if this function was intended to be internal only?)

[1] http://www.ecma-international.org/ecma-262/7.0/index.html#sec-json-object


travisbrown:
Thanks, @jimmydivvy—this looks reasonable to me. @chandu0101, do you remember the intention behind the current behavior?

PR #377, opened 2016-09-20
Mention scala-jsonapi in related projects section

emanresusername:


travisbrown:
Thanks @emanresusername!

PR #381, opened 2016-09-25
Add Json#\\.

kevinmeredith:
Aims to address https://github.com/travisbrown/circe/issues/256.

```
scala> :paste
// Entering paste mode (ctrl-D to finish)

import io.circe._, io.circe.parser._

val json =
"""{
  "sha": "some sha",
  "parents": [{
    "url": "some url",
    "sha": "some parent sha"
  }]
}"""

val obj = parse(json).right.get


// Exiting paste mode, now interpreting.

// omitted

scala> Json.\\(obj, "sha")
res0: List[io.circe.Json] = List("some sha", "some parent sha")
```

TODO: 
* add tests
* verify that this PR's `Json#\\` aligns with Play's `\\` on multiple tests

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Json.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Json.scala
@@ -161,6 +161,28 @@ sealed abstract class Json extends Product with Serializable {
    * Use implementations provided by case classes.
    */
   override def hashCode(): Int
+
+  // Alias for `findAllByKey`.
+  final def \\(key: String): List[Json] = findAllByKey(key)
+
+  /**
+    * Recursively return all values matching the specified `key`.
+    *
+    * The Play docs, from which this method was inspired, reads:
+    *   "Lookup for fieldName in the current object and all descendants."
+    */
+  final def findAllByKey(key: String): List[Json] = keyValues(this).collect {
+    case (k, v) if (k == key) => v
+  }
+
+  private def keyValues(json: Json): List[(String, Json)] = json match {
+    case JObject(obj)  => obj.toList.flatMap { case (k, v) => keyValuesHelper(k, v) }
+    case JArray(elems) => elems.toList.flatMap(keyValues)
+    case _             => Nil
+  }
+
+  private def keyValuesHelper(key: String, value: Json): List[(String, Json)] =
+    (key, value) :: keyValues(value)
 }
 
 final object Json {
--- a/modules/tests/shared/src/test/scala/io/circe/JsonSuite.scala
+++ b/modules/tests/shared/src/test/scala/io/circe/JsonSuite.scala
@@ -1,5 +1,6 @@
 package io.circe
 
+import io.circe.Json.{JString, JArray, JNumber, JBoolean, JObject, JNull}
 import io.circe.tests.CirceSuite
 
 class JsonSuite extends CirceSuite {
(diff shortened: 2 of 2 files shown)


kevinmeredith:
I added some example-based tests. Next I'll look for an `Arbitrary[Json]` to use property-based testing.

Perhaps it would be reasonable to state the property?

> circe's `Json#\\` must produce the same result as Play's `\\` for every JSON input

kevinmeredith:
Hi @travisbrown - can you please take a look?

travisbrown:
Yes, thanks much @kevinmeredith—I should be able to review today or tomorrow.

kevinmeredith:
Respectful bump

modules/tests/shared/src/test/scala/io/circe/JsonSuite.scala
```diff
@@ -13,4 +14,51 @@ class JsonSuite extends CirceSuite {
 
     assert(merged.asObject.map(_.toList) === Some(fields.reverse))
   }
+
+  val Key    = "x"
+  val Value1 = "fizz"
+  val Value2 = "foobar"
+
+  "\\\\"  should "return all values matching the given key with key-value pairs at heights 0 and 1." in {
```
travisbrown: Nitpick, but could we triple-quote these names instead of escaping?

modules/tests/shared/src/test/scala/io/circe/JsonSuite.scala
```diff
@@ -13,4 +14,51 @@ class JsonSuite extends CirceSuite {
...
+    val expected  = List(JString(Value1), JString(Value2), JNull)
+    val `0`       = (Key, JString(Value1))
```
travisbrown: Can we use something like `at0`, `at1`, etc. here?

modules/tests/shared/src/test/scala/io/circe/JsonSuite.scala
```diff
@@ -13,4 +14,51 @@ class JsonSuite extends CirceSuite {
 
     assert(merged.asObject.map(_.toList) === Some(fields.reverse))
   }
+
+  val Key    = "x"
```
travisbrown: Is there a reason for these to be capitalized?

modules/core/shared/src/main/scala/io/circe/Json.scala
```diff
@@ -161,6 +161,26 @@ sealed abstract class Json extends Product with Serializable {
...
+    *   "Lookup for fieldName in the current object and all descendants."
+    */
+  final def \\(key: String): List[Json] = keyValues(this).collect {
+    case (k, v) if (k == key) => v
+  }
+
+private def keyValues(json: Json): List[(String, Json)] = json match {
+    case JObject(obj)  => obj.toList.flatMap { case (k, v) => keyValuesHelper(k, v) }
+    case JArray(elems) => elems.toList.flatMap(keyValues)
+    case _             => List.empty
```
travisbrown: This is just a matter of nitpicky consistency with the rest of the codebase, but could we use `Nil` here.

modules/core/shared/src/main/scala/io/circe/Json.scala
```diff
@@ -161,6 +161,26 @@ sealed abstract class Json extends Product with Serializable {
...
   override def hashCode(): Int
+
+
+  /**
+    * Recursively return all values matching the specified `key`.
+    *
+    * The Play docs, from which this method was inspired, reads:
+    *   "Lookup for fieldName in the current object and all descendants."
+    */
+  final def \\(key: String): List[Json] = keyValues(this).collect {
```
travisbrown: I've been really resistant to symbolic operators, and though I think I can live with this one, I'd still like it to be an alias for something like `deepFind` or `findAllByKey`. What do you think?

kevinmeredith:
I believe that I've addressed all of your comments, Travis.

modules/core/shared/src/main/scala/io/circe/Json.scala
kevinmeredith: Does `alias` mean that `findAllByKey` (or the other name) would have `public` access?

kevinmeredith:
I just made `findAllByKey` public available. I suppose it's obvious that `alias` means that both `\\` and the method that it aliases should be `public`.

kevinmeredith:
OK I'm confident all issues are addressed.

travisbrown:
Thanks, @kevinmeredith.

Issue #320, opened 2016-07-02
Consider going Predef-less


travisbrown:
In the spirit of #315, I know there are other places where we're doing stuff that involves unnecessary wrapping (like `string.size` as opposed to `string.length`), and I'd like to see what it would look like to turn on `-Yno-predef` for everything (excluding tests).

I've pushed [a branch](https://github.com/travisbrown/circe/tree/starter/320) with the configuration that's needed.

travisbrown:
Done in #526.

PR #383, opened 2016-09-25
No Predef

travisbrown:
These changes address #320. In all cases except `round` I'm pretty sure the syntax provided by `Predef` doesn't involve extra allocations, but I'm tempted to make this change anyway, for the sake of explicitness and future safety.


travisbrown:
Superseded by #526.

PR #384, opened 2016-09-26
Create circe-testing

travisbrown:
I've moved some of the pieces that are likely to be useful to users who want to test their own codecs to a new publishable circe-testing module. This means you can add a circe-testing test dependency to your project and then do something like this:

```scala
import cats.Eq

case class Foo(i: Int, s: String)

object Foo {
  implicit val decodeFoo: Decoder[Foo] =
    Decoder.forProduct2("someNumber", "someString")(Foo.apply)

  implicit val encodeFoo: Encoder[Foo] =
    Encoder.forProduct2("someNumber", "someString")(foo => (foo.i, foo.s))

  implicit val eqFoo: Eq[Foo] = Eq.fromUniversalEquals
}
```

And then in your tests:

```scala
import io.circe.testing.CodecTests
import org.scalacheck.Arbitrary

implicit val arbitraryFoo: Arbitrary[Foo] = Arbitrary(
  for {
    i <- Arbitrary.arbitrary[Int]
    s <- Arbitrary.arbitrary[String]
  } yield Foo(i, s)
)
```

And then finally:

```scala
scala> CodecTests[Foo].codec.all.check
+ codec.consistency with accumulating: OK, passed 100 tests.
+ codec.roundTrip: OK, passed 100 tests.
```

There are also laws for printers and parsers, and a couple of other useful things (arbitrary JSON number strings, etc.).

Update: one note about naming and arrangement: the general idea is that anything that's useful for testing circe types generally should go in circe-testing, and anything that's specific to testing circe's own modules (or that requires a ScalaTest dependency) goes in circe-tests. I know this naming is going to be a little confusing, but I think it makes sense.


andyscott:
👏 👏 👏 👏 
`CodecTests` was exactly what I was hoping to use as a lib!

Issue #390, opened 2016-09-26
Another divergence checking nightmare


travisbrown:
This issue is a minimization of #386—thanks much to @dzikowski for the original report.

The problem shows up in situations like this:

``` scala
import io.circe.Encoder
import io.circe.syntax._
import io.circe.generic.auto._

object Foo {
  case class Bar(i: Int)
  object Bar {
    implicit val encodeBar: Encoder[Bar] = Encoder[Int].contramap(_.i)
  }
}

case class Bars(values: Seq[Foo.Bar])
```

And then:

``` scala
scala> List(Bars(List(Foo.Bar(1)))).asJson.noSpaces
res0: String = {"::":[{"values":[1]}]}
```

This doesn't happen if `Bars` is defined in the same object as `Bar`, or if `Bar` is moved out of `Foo`, or if `Bars` is defined separately (e.g. two `:paste`s in the REPL). It also doesn't happen if `Bars` is explicitly given a derived instance with `deriveEncoder`.

The problem is that the divergence checker is bailing when trying to find the `Bars` encoder in `encodeTraversableOnce[Bars, List]`, because the derivation mechanism needs to use `encodeTraversableOnce` again for `Bars#values`. I have no idea why this only happens in the nested case, but the divergence checker's notions of "bigger" and "smaller" have always been a mystery to me.

If we had a Shapeless dependency in circe-core this would be trivial—we'd just put a `Lazy` in `encodeTraversableOnce` and we're done. That's not really an option, though.

The only workaround I can think of is to copy-paste all generic instances where this could happen from both `Encoder` and `Decoder` into `auto` (user-defined generic contexts _should_ be fine, I think—at least they are in the simple cases I've tried). This works, but it's really, really unsatisfying, and any other suggestions would be greatly appreciated.

milessabin:
FYI, you could have a "soft" dependency on `shapeless.Lazy` using the technique employed in [export-hook](https://github.com/milessabin/export-hook/blob/master/core/src/main/scala/export/export.scala#L499-L507).

travisbrown:
@milessabin But I've held out against macros in circe-core for soooo long. 😦

milessabin:
Ahh ... sorry, I thought this was an issue relating to the new macros in recent circe releases.

travisbrown:
[...]

Issue #386, opened 2016-09-26
Embedded collection of Enumeration values causes errors in serialization of collection of objects


dzikowski:
I have observed weird behavior of serialization of objects with collection of Enumeration values (for example `case class User(roles: Seq[Roles.Value])`) . Serialization of collection of this objects (for example `Seq[User]`) produces JSON Object with double-colon attribute: `{"::":[obj1, obj2, ...]}`, when correct output should be JSON Array: `[obj1, obj2, ...]`. 

This error **does not** occur when instead of Enumeration values we have other objects (even with custom Encoders).

Workaround: Implement your own `Encoder` for collection of Enumeration values.

Minimal example:

``` scala
import io.circe._
import io.circe.syntax._
import io.circe.generic.auto._

object Roles extends Enumeration {
  val Admin, Guest = Value
}

implicit object RoleEncoder extends Encoder[Roles.Value] {
  override def apply(r: Roles.Value): Json = r.toString.asJson
}

case class User(roles: Seq[Roles.Value])

val u = User(Seq(Roles.Admin))

Seq(u).asJson.noSpaces
// ERROR: res0: String = {"::":[{"roles":["Admin"]}]}

implicit object RolesSeqEncoder extends Encoder[Seq[Roles.Value]] {
  override def apply(rs: Seq[Roles.Value]): Json = rs.map(_.asJson).asJson
}

Seq(u).asJson.noSpaces
// OK: res1: String = [{"roles":["Admin"]}]
```

travisbrown:
@dzikowski Hmm, that's bad. I'm taking a look now. What version of circe are you using?

dzikowski:
I have checked `0.4.1`, `0.5.1` and current head of master branch. It seems to apply to all of these versions.

travisbrown:
@dzikowski Can you tell me more about your build setup? I can't reproduce this exact error (for me no encoder is found, which is also a bug, but not as bad of one).

dzikowski:
OK. I dug a bit deeper, tried to reproduce it in isolated environment (only Circe dependencies) and... I failed :-) It seems that is error is related also with Akka HTTP and the glue I use to connect Circe with Akka HTTP. Compilation errors (no encoder found) were missing, and - I guess - in runtime some kind of default Encoder/Marshaller was chosen what led to this weird output. If you want, you can see the minimal project, where this issue is reproducible: [https://github.com/dzikowski/circe-akka-http-test](https://github.com/dzikowski/circe-akka-http-test).

[...]

PR #394, opened 2016-09-30
Reinstate export-hook

travisbrown:
This is a fix for #386 / #390 that basically lets [export-hook](https://github.com/milessabin/export-hook) handle the messy parts.

We've had an export-hook dependency before (see #242), and the fact that it helps in corner cases like this makes me really tempted to switch back.


travisbrown:
This is definitely not going to happen.

PR #395, opened 2016-10-03
Semi-automatic codec derivation with monocle.Iso

andrelfpinto:
I don't know if generic derivation is going to cover this, but sometimes one happens to already have an Iso that converts from one class to another.


travisbrown:
Thanks, @andrelfpinto—this is a good idea. I have a couple of questions:

* I think I'd like to try to think of a different name than `semiauto` to avoid confusion with the _generic_ derivation package.
* I know this is a matter of personal taste, but I'm not a big fan of the overloading-via-dummy-implicit trick—adding a `From` or `Reverse` suffix seems preferable to noise in the API and the cognitive overhead for readers.

What do you think?

andrelfpinto:
Well naming is, for me at least, a very difficult problem. No wonder the two issues were kind of related to that.

1. I thought about putting these methods in the `semiauto` object in the `generic` package, but then the project would have to depend on `optics` project and I decided to just put it the `optics`.  Then I didn't have a better name for the object. Feel free to suggest a name that suits better the design of the library.

2. Again, I'm just a user of the library and I don't have the global picture of the choices you have made to design this (awesome, by the way) library. Also, I do not make a living by programming and may not have the best taste. What would you prefer?

Summing up, it's your call.

andrelfpinto:
Also, I have read the `CONTRIBUTING.md` and I'm sorry to have skipped the first bullet "Pick (or report) an issue". Maybe these issues would have been discussed there first.

travisbrown:
@andrelfpinto Okay, I've thought a little more about this, and have the following suggestions:

* Let's put these directly into the optics package object. Since they're not implicit and have distinctive names, it doesn't really pollute the scope for people who import `io.circe.optics._` wholesale.
* Let's put a `Reverse` suffix on the methods that currently have a dummy implicit. It's a little more work for (some) users, but in general the project tends to avoid overloading in favor of workarounds like `DummyImplicit`, and I think it'd be best to be consistent.

If you don't mind making these changes we should be able to get this into 0.6.0. Thanks again!

andrelfpinto:
Hey @travisbrown, if the PR is ok for you, I will rebase it for you to merge it.

travisbrown:
@andrelfpinto Yes, looks great, but there's one last thing I just noticed: the `all` object has moved from the package to the package object, which affects the way it's represented in bytecode and breaks binary compatibility (which is already broken because of the Monocle version update, but we might as well minimize it).

Do you mind changing that part back? I.e. this should work:
```scala
package optics {
  final object all extends JsonNumberOptics with JsonObjectOptics with JsonOptics
}

package object optics {
  // new stuff here
}
```

The compiler will merge them appropriately from the user's point of view.

andrelfpinto:
Sorry about the `package object`: I didn't know it would break the binary compatibility.

travisbrown:
No worries, thanks again! Is this ready to go?

andrelfpinto:
Yes, I just need to rebase it now, right (as in the CONTRIBUTING.md)?

travisbrown:
@andrelfpinto I can just do that when I merge—it's not a big deal here since this doesn't touch many files, etc.

andrelfpinto:
OK, then. Thank you for helping me through the process and thanks again for the library.

travisbrown:
@andrelfpinto Sure! It's a nice addition.

Issue #396, opened 2016-10-06
Decoder[A].validate is subject to infinite recursion


sirocchj:
```
import io.circe._
implicit val intDecoder = Decoder[Int].validate(_ => true, "")
Json.fromInt(42).as[Int]
```

Issue is easy-peasy to fix, apply() is recursively calling itself instead of self type'd apply, submitting PR

crakjie:
Tanks to open a pr 👍

travisbrown:
Resolved in #396.

PR #397, opened 2016-10-06
Fix for Decoder[A].validate infinite recursion (#396)

sirocchj:
Should be pretty clear what's been changed 😄


travisbrown:
Oh, wow—good catch. This is why paying more attention to code coverage would be a good thing.

sirocchj:
Just really stumbled across it :) but yeah, it would certainly be a Good Thing™️

PR #399, opened 2016-10-07
Make `BiggerDecimal` extend `Serializable`

jeremyrsmith:
This PR makes `BiggerDecimal` extend `Serializable`. This should ensure that all ADT elements are fully `Serializable`, which is necessary for Spark friendliness (among other things).


travisbrown:
Thanks @jeremyrsmith! Mind adding a test for `BiggerDecimal` [here](https://github.com/travisbrown/circe/blob/master/modules/tests/shared/src/test/scala/io/circe/SerializableSuite.scala)?

travisbrown:
@jeremyrsmith I'm going to go ahead and merge this—we can add the test later.

PR #400, opened 2016-10-07
Make `BiggerDecimal` covered by `Json` generator

jeremyrsmith:
This commit adds a generator for `BiggerDecimal` and includes it in the generator for `Json`-numbers. This way it will be included in `Arbitrary[Json]` and covered by all associated property checks.

This causes some existing tests to fail, because they don't pass for `BiggerDecimal` cases.


travisbrown:
Now included in #408. Thanks again, @jeremyrsmith!

PR #402, opened 2016-10-11
build circe-opticsjs package

ericbmerritt:
This adds a crossProject definition for circe-optics so that the dervied
scalajs package can be built.

Two things to note. I ran out of heap space trying to run the tests and this is a more complex build.sbt than any I have worked with before, so please pay close attention before merging.

---
This change is [Reviewable](https://reviewable.io/reviews/travisbrown/circe/402)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -512,12 +512,12 @@ lazy val opticsBase = crossProject.in(file("modules/optics"))
   .settings(allSettings: _*)
   .settings(
     libraryDependencies ++= Seq(
-      "com.github.julien-truffaut" %% "monocle-core" % "1.2.2",
-      "com.github.julien-truffaut" %% "monocle-law" % "1.2.2" % "test",
-      compilerPlugin("org.scalamacros" % "paradise" % "2.1.0" cross CrossVersion.full)
-    ),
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -504,12 +504,12 @@ lazy val spray = project.in(file("modules/spray"))
...
+lazy val opticsBase = crossProject.in(file("modules/optics"))
   .settings(
     description := "circe optics",
     moduleName := "circe-optics"
   )
-  .settings(allSettings)
+  .settings(allSettings: _*)
   .settings(
     libraryDependencies ++= Seq(
       "com.github.julien-truffaut" %% "monocle-core" % "1.2.2",
```
travisbrown: I think you'll need a `%%%` instead of `%%` here?

build.sbt
```diff
@@ -518,7 +518,13 @@ lazy val optics = project.in(file("modules/optics"))
     ),
     mimaPreviousArtifacts := Set("io.circe" %% "circe-optics" % previousCirceVersion)
```
travisbrown: I believe this will need to be moved to a new `.jvmSettings` section.

PR #403, opened 2016-10-13
tutorial: Added comment about needing macro paradise plugin when using JsonCodec macro

jtcwang:
It took me a while to figure out why it wasn't working for me.


travisbrown:
Thanks, @jtcwang. This used to be mentioned in the FAQ, but it looks like that got lost in the move to the new site.

PR #404, opened 2016-10-13
A few small build fixes for optics

travisbrown:
This is a follow-up to #402 with a few fixes to the build.


ericbmerritt:
Nice. Sorry for missing these.

travisbrown:
@ericbmerritt No worries—it's a complicated build file. :)

PR #413, opened 2016-10-18
Adding 'The Guardian' to list of orgs using Circe

LATaylor-guardian:


travisbrown:
Thanks, @LATaylor-guardian!

PR #414, opened 2016-10-19
Update to Cats 0.8.0 (do not merge)

travisbrown:
Still needs locally published Cats and Refined, but otherwise I think it's ready to go.


travisbrown:
Superseded by #417.

PR #421, opened 2016-10-30
Add semi-automatic derivation for ADT enumerations

travisbrown:
Currently the derived codecs for case objects result in either an empty object, or a field with an empty object value, depending on whether the value is statically typed as an ADT leaf or root:

``` scala
scala> import io.circe.generic.auto._, io.circe.syntax._
import io.circe.syntax._

scala> sealed trait Base; case object Foo extends Base
defined trait Base
defined object Foo

scala> Foo.asJson.noSpaces
res0: String = {}

scala> (Foo: Base).asJson.noSpaces
res1: String = {"Foo":{}}
```

This will probably always be the default behavior, since it's always safe and unambiguous, and since having derived encoders that return JSON values that aren't objects complicates things significantly.

There is a common use case for case objects that can be pretty easily supported, though. If we're defining an enumeration as a sealed trait of case objects, we often want the values to be represented as strings. This PR adds support for this representation to `io.circe.semiauto`:

``` scala
import io.circe.generic.semiauto.{ deriveEnumerationDecoder, deriveEnumerationEncoder }
import io.circe.jawn._, io.circe.syntax._

sealed trait Suit extends Product with Serializable
case object Club extends Suit
case object Heart extends Suit
case object Spade extends Suit
case object Diamond extends Suit

object Suit {
  implicit val decodeSuit: Decoder[Suit] = deriveEnumerationDecoder
  implicit val encodeSuit: Encoder[Suit] = deriveEnumerationEncoder
}
```

And then:

``` scala
scala> decode[Suit]("\"Club\"")
res0: Either[io.circe.Error,Suit] = Right(Club)

scala> List(Heart, Club, Diamond).asJson
res1: io.circe.Json =
[
  "Heart",
  "Club",
  "Diamond"
]
```

If we had any case classes that extended `Suit`, the derivation would fail to compile.

Note that this must be explicitly invoked via `deriveEnumerationX`, and will not collide with any automatically derived decoders.

### DIFF
--- a/modules/generic/shared/src/main/scala/io/circe/generic/decoding/EnumerationDecoder.scala
+++ b/modules/generic/shared/src/main/scala/io/circe/generic/decoding/EnumerationDecoder.scala
@@ -1,0 +1,34 @@
+package io.circe.generic.decoding
+
+import io.circe.{ Decoder, DecodingFailure, HCursor }
+import shapeless.{ :+:, CNil, Coproduct, HNil, Inl, Inr, LabelledGeneric, Witness }
+import shapeless.labelled.{ field, FieldType }
+
+abstract class EnumerationDecoder[A] extends Decoder[A]
+
+final object EnumerationDecoder {
+  implicit val decodeEnumerationCNil: EnumerationDecoder[CNil] = new EnumerationDecoder[CNil] {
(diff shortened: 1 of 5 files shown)


travisbrown:
Superseded by #429.

PR #427, opened 2016-11-02
Adds LocalTime enconder/decoder to io.circe.java8.time and tests

darienmt:


travisbrown:
Looks great, thanks @darienmt!

PR #433, opened 2016-11-03
Update to Scala 2.12.0 (do not merge)

travisbrown:
~~Still waiting on a few dependencies.~~

Okay, now confirmed that it's just Cats that we're waiting on.


travisbrown:
Hmm, PR isn't updating.

PR #440, opened 2016-11-07
Decoder#or should accumulate failures when using decodeAccumulating

travisbrown:
Inspired by the discussion [here](https://github.com/typelevel/cats/pull/1448), this change makes decoders built with `or` accumulate errors if error accumulation is requested.


travisbrown:
Ugh, now I'm having second thoughts about this…

PR #447, opened 2016-11-10
Correct scaladoc on KeyDecoder#apply.

tixxit:
I think the scaladoc had gotten copied from `KeyEncoder`. This just uses a more correct phrasing.


travisbrown:
Thanks, @tixxit!

PR #448, opened 2016-11-10
add diffson to related projects

emanresusername:


travisbrown:
Thanks @emanresusername!

PR #450, opened 2016-11-14
Added custom key string in generic extra

aparo:
During the wrapping external API and general in real world there are a lot of key values in JSOn that must be contains custom values such as "@timestamp", "my-wonderful-field" and so on.
This pull request add an annotation *Key* that allows to define a string that will be used in de/serialize an objects.
For example
```
case class ConfigExampleFoo(thisIsAField: String, a: Int = 0, @Key("myField") b: Double)
```


travisbrown:
Thanks, @aparo! I'll review this weekend.

aparo:
I saw that it requires that case class Key needs to extend Annotation.
For the second error, I don't know probably it's fired by the first one.

aparo:
Travis, I run the full test suite on my branch without problems. Now I see codecov issues in files that I never changed.
This feature is very common in business applications.

travisbrown:
Ugh, this totally slipped off my radar. I'm at NE Scala today and tomorrow but should be able to get this done this weekend.

aparo:
Don't worry. I fixed the scalacheck errors. I'll add some docs examples.

From ConfiguredJsonCodecWithKeySuite:

```scala
@ConfiguredJsonCodec
sealed trait ConfigExampleBase
case class ConfigExampleFoo(thisIsAField: String, a: Int = 0, @Key("myField") b: Double) extends ConfigExampleBase

  "ConfiguredJsonCodec" should "support key annotation and configuration" in forAll { (f: String, b: Double) =>
    val foo: ConfigExampleBase = ConfigExampleFoo(f, 0, b)
    val json = json"""{ "type": "ConfigExampleFoo", "this_is_a_field": $f, "myField": $b}"""
    val expected = json"""{ "type": "ConfigExampleFoo", "this_is_a_field": $f, "a": 0, "myField": $b}"""

    assert(Encoder[ConfigExampleBase].apply(foo) === expected)
    assert(Decoder[ConfigExampleBase].decodeJson(json) === Right(foo))
  }
```

aparo:
Do you have any news about this pulll request?
Do you need something else that I can figure out?

travisbrown:
Okay, this has languished a long, long time, but I want it in 0.8.0 and I've got a few little changes that I think should go in (e.g. deriving the labelling from `LabelledGeneric` instead of `DefaultSymbolicLabelling` so that people who provide their own `LabelledGeneric` don't get inconsistent results, etc.). I'll be opening a PR against your branch shortly, @aparo.

travisbrown:
Included in #616.

PR #451, opened 2016-11-14
add ravel law to adopters list

emanresusername:


travisbrown:
Thanks @emanresusername!

PR #453, opened 2016-11-15
Use friendly name in POM

adelbertc:
Similar fix (https://github.com/typelevel/cats/pull/1426) and
issue (https://github.com/typelevel/cats/issues/1396) to Cats.


travisbrown:
Thanks, @adelbertc!

PR #461, opened 2016-11-20
Introduce decodeByteBuffer, decodeFile, decodeByteArray

vkostyukov:
Circe supported parsers (Jawn, Jackson) provide meanings for parsing from byte arrays and byte buffers (which is super nice for things like https://github.com/finagle/finch/pull/671). Although, the `decodeX` API doesn't support anything but `String` as an input. While it's totally possible to drop down to the `parseX` variant and do decoding as a separate step, having a single entry point method (i.e., `decodeX`) for may both simplify the call-site and reduce some unnecessary allocations.

This PR introduces `decodeX` variants into both parsers for decoding from `ByteBuffer`s, `File`s, and `Array[Byte]`s.

Please, let me know if this requires some testing - I'm happy to add those respectively.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Parser.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Parser.scala
@@ -5,17 +5,25 @@ import cats.data.{ NonEmptyList, Validated, ValidatedNel }
 trait Parser extends Serializable {
   def parse(input: String): Either[ParsingFailure, Json]
 
-  final def decode[A](input: String)(implicit decoder: Decoder[A]): Either[Error, A] =
-    parse(input) match {
-      case Right(json) => decoder.decodeJson(json)
-      case l @ Left(_) => l.asInstanceOf[Either[Error, A]]
-    }
+  protected[this] final def finishDecode[A](input: Either[ParsingFailure, Json])(implicit
+    decoder: Decoder[A]
+  ): Either[Error, A] = input match {
+    case Right(json) => decoder.decodeJson(json)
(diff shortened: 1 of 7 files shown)


travisbrown:
Thanks @vkostyukov! This looks good to me, but I do think we should add test coverage.

vkostyukov:
Thanks @travisbrown! I restructured `ParserLaws` and `ParserTest` so we can test`decodeX` variations. In its current version, I added round-trip tests for `decode` (from string, from byte array, and from byte buffer) and `parse` (from byte array, from byte buffer). Let me know what do you think about this.

vkostyukov:
I think I should invest in testing more to make codecov happy. Let me think about that.

travisbrown:
@vkostyukov I might have time to take a look in an hour or so as well.

travisbrown:
It's a shame `Parser` is a trait instead of an abstract class, since otherwise this would be a binary compatible change. I wonder if it's too late to change it for future versions…

vkostyukov:
Added more tests. Thinking about testing `decode*Accumulating` variants as well.

vkostyukov:
Both codecov and I are happy now. 

@travisbrown let me know if you have any concerns about the current tests structure for parsers - I'm happy to change them. Also, happy to `s/trait/absrtact class` if you feel we have to do that right now.

travisbrown:
Thanks @vkostyukov! The tests look good to me. And no, I think the abstract class move should happen separately—I want to be sure it doesn't break anyone else's code, and don't want to hold up this PR for that.

Issue #271, opened 2016-04-14
Modify deepMerge behavior on arrays


RomanIakovlev:
At present, when one runs `deepMerge` on 2 arrays, only the argument contents remains in the resulting array, the contents of calling array is lost:
```
import io.circe.Json, io.circe.literal._

val a1: Json = json"""[{"id": 1}, {"id": 2}, {"id": 3}]"""
val a2: Json = json"""[{"id": 4}, {"id": 5}, {"id": 6}]"""

a1.deepMerge(a2)
res4: Json = [{"id" : 4}, {"id" : 5}, {"id" : 6}]
```
This behavior is [documented](https://github.com/travisbrown/circe/blob/master/core/shared/src/main/scala/io/circe/Json.scala#L132). However, could we consider changing it for arrays, and make it do the simple concatenation, like `[{"id": 1}, {"id": 2}, {"id": 3}, {"id" : 4}, {"id" : 5}, {"id" : 6}]`? I'd say, the current behavior is not completely intuitive.

travisbrown:
I'm sympathetic but seem to remember some motivation for the current behavior. Let's wait a bit for some more feedback—

jatcwang:
Concatenation is what I assume it does but it seems like Lodash is performing merging by array index as well.

https://codepen.io/anon/pen/eVXPXX

In the OP's example, if `a2` only has 1 element, then you'll end up with `[{"id": 4}, {"id": 2}, {"id": 3}]`

redarqas:
A concatenation mode could be useful

eirirlar:
I think both concat and replace by index mode has merits. Maybe we could add concat mode by a Boolean param to the deepMerge method, could be default false? Or just another method deepMergeConcat?

jcallin:
Any reason why #1275 was closed and not merged? This would be super useful

PR #466, opened 2016-11-22
Move AST and other core types to new circe-ast module [do not merge]

travisbrown:
As suggested by @adelbertc [on Gitter this evening](https://gitter.im/travisbrown/circe?at=583388cec0a2732923ffeca6).

I'm not 100% sold on this, but even apart from the fact that you can use the AST without bringing in the rest of the project, I like the fact that it gets rid of some of the circularity and multiple-ways-of-doing-things in the API.

The fact that we need to move core types to a new package is kind of a drag, but we could provide aliases in `io.circe` to make the transition less painful.

I'm interested in any feedback. If we want to move forward with this, the one additional thing that would need to be done for this PR is some test rearrangement.

### DIFF
--- a/modules/ast/src/main/scala/io/circe/ast/Json.scala
+++ b/modules/ast/src/main/scala/io/circe/ast/Json.scala
@@ -1,4 +1,4 @@
-package io.circe
+package io.circe.ast
 
 import cats.{ Eq, Show }
 import io.circe.numbers.BiggerDecimal
@@ -48,16 +48,6 @@ sealed abstract class Json extends Product with Serializable {
     case JObject(o)  => jsonObject(o)
   }
 
-  /**
-   * Construct a cursor from this JSON value.
-   */
-  final def cursor: Cursor = Cursor(this)
-
-  /**
-   * Construct a cursor with history from this JSON value.
-   */
-  final def hcursor: HCursor = HCursor.fromCursor(Cursor(this))
-
   def isNull: Boolean
   def isBoolean: Boolean
   def isNumber: Boolean
@@ -95,11 +85,6 @@ sealed abstract class Json extends Product with Serializable {
       case JArray(_)   => "Array"
       case JObject(_)  => "Object"
     }
-
-  /**
-   * Attempts to decode this JSON value to another data type.
-   */
-  final def as[A](implicit d: Decoder[A]): Decoder.Result[A] = d(HCursor.fromCursor(cursor))
 
   /**
    * Pretty-print this JSON value to a string using the given pretty-printer.
--- a/build.sbt
+++ b/build.sbt
@@ -78,6 +78,7 @@ def noDocProjects(sv: String): Seq[ProjectReference] = Seq[ProjectReference](
   genericExtrasJS,
   shapesJS,
   numbersJS,
+  astJS,
   opticsJS,
   parserJS,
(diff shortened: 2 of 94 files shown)


adelbertc:
Wow you're fast

For context, I found myself wanting just a JSON AST and parser when working on https://github.com/adelbertc/circular which would also be for parsing/printing JSON, but through an optics-like mechanism instead of type classes. I've had success with [scodec](https://github.com/scodec) using just the `scodec-bits` project without `scodec` itself (which has the analogous binary encode/decode type classes).

To add to @travisbrown 's point, I also like to see the separation of the core data types from the ways-you-can-do-stuff such as type classes, optics, etc.

RomanIakovlev:
If you want to touch AST, maybe it's a good opportunity to reevaluate https://github.com/circe/circe/issues/271?

modules/ast/src/main/scala/io/circe/ast/Json.scala
```diff
@@ -97,11 +87,6 @@ sealed abstract class Json extends Product with Serializable {
     }
 
   /**
-   * Attempts to decode this JSON value to another data type.
-   */
-  final def as[A](implicit d: Decoder[A]): Decoder.Result[A] = d(HCursor.fromCursor(cursor))
```
cb372: I'm a bit sad to see this method go. I know it's redundant but it's also quite convenient. Maybe it could be added back as an extension method using the syntax._ import?

notxcain: Yes, moving it to `syntax` would be very handy

nicolasdalsass:
The main benefit I understand is allowing people to depend only on circe-ast, without pulling circe-core. For what I can tell, this is an economy of a few kilobytes top when importing only the ast (and probably less, the current core becoming ~ 30 scala files).

Otoh, it's yet another import to drag in and remember when developing use cases beyond auto derivation with circe, which are already a pain (I've just discovered by reading the PR than by importing io.circe.syntax._ I could have written much more concise code ^^' ). 

So, while I appreciate the fact it encapsulates even more tightly each functional part, I think this is detrimental to newcomers to circe, by making again that tiny bit harder to dive in.

travisbrown:
@nicolasdalsass Would providing aliases in `io.circe` make the situation better? This is what Cats does for e.g. `cats.Eq`, which lives in cats-kernel, and the user experience seems okay so far.

The biggest downside I can see with the aliases (apart from just the extra complexity) is the compiler warnings for people who import both `io.circe._` and the `io.circe.ast` types in the same scope.

More generally I agree that it's not clear yet that the benefits are worth the costs—I just want to get a clear sense of how bad the costs would be for people given the things we can do to mitigate them.

nicolasdalsass:
Yes, I think aliases help a lot !

Anoter example about imports simplification is that being able to read code snippets (e.g on gitter or SO) and use it simply is nice to have. More often than not, this kind of ressource doesn't include imports, and I've often been frustrated (not only using circe ^^) by not being able to know how to access the function that solves my need.

So, +1 for anything keeping imports relatively contained when using the lib !

travisbrown:
This may happen after #228 but if it does the motivation will be slightly different—most importantly, circe-core wouldn't necessarily even have a circe-ast dependency.

PR #472, opened 2016-11-25
Add a contravariant functor instance to ObjectEncoder

backuitist:


travisbrown:
Thanks @backuitist!

modules/core/shared/src/main/scala/io/circe/ObjectEncoder.scala
```diff
@@ -41,6 +50,13 @@ final object ObjectEncoder extends LowPriorityObjectEncoders {
   final def instance[A](f: A => JsonObject): ObjectEncoder[A] = new ObjectEncoder[A] {
     final def encodeObject(a: A): JsonObject = f(a)
   }
+
+  /**
+    * @group Instances
+    */
+  implicit final val encoderContravariant: Contravariant[ObjectEncoder] = new Contravariant[ObjectEncoder] {
```
backuitist: Done

backuitist:
Thanks for the quick review @travisbrown

travisbrown:
Thanks, @backuitist! The test failure looks like an unrelated fluke (the string `꩘` apparently parses as a valid Java integer)—I'll fix it in a separate PR, restart the test for now, and merge when green.

PR #473, opened 2016-11-26
add filter and unsafeFilter to JsonPath and rename arrFilter and objFilter

julien-truffaut:
I am planning to add some methods on `Optional`, `Prism` and so on to simplify the syntax of `filter` see https://github.com/julien-truffaut/Monocle/pull/425

```
// currently
root.cars.each.filter(root.maxSpeed.int.asFold.exist(_ > 100)(_)).model.string.getAll(john)

// after monocle PR
root.cars.each.filter(root.maxSpeed.int.exist(_ > 100)).model.string.getAll(john)
```


travisbrown:
Thanks, @julien-truffaut!

PR #474, opened 2016-11-27
Build sbt diet

jonas:
Trims 200+ lines off `build.sbt` in an attempt to make it easier to maintain.
```
 48 files changed, 115 insertions(+), 340 deletions(-)
```

- Most project boilerplate is now configured via `circeCrossModule` and `circeModule`.
- Make several cross projects pure: core, generic, numbers, refined, scodec
- Improve management of projects by declaring several groups and use that to define aliases etc.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -192,7 +192,7 @@ lazy val aggregatedProjects: Seq[ProjectReference] = Seq[ProjectReference](
   if (sys.props("java.specification.version") == "1.8") Seq[ProjectReference](java8) else Nil
 )
 
-lazy val macroDependencies: Seq[Setting[_]] = Seq(
+def macroSettings(scaladocFor210: Boolean): Seq[Setting[_]] = Seq(
   libraryDependencies ++= Seq(
     "org.scala-lang" % "scala-compiler" % scalaVersion.value % "provided",
     "org.scala-lang" % "scala-reflect" % scalaVersion.value % "provided",
@@ -205,6 +205,12 @@ lazy val macroDependencies: Seq[Setting[_]] = Seq(
       case Some((2, scalaMajor)) if scalaMajor >= 11 => Nil
       // in Scala 2.10, quasiquotes are provided by macro paradise.
       case Some((2, 10)) => Seq("org.scalamacros" %% "quasiquotes" % "2.1.0" cross CrossVersion.binary)
+    }
+  },
+  sources in (Compile, doc) := {
+    CrossVersion.partialVersion(scalaVersion.value) match {
+      case Some((2, 10)) if !scaladocFor210 => Nil
+      case _ => (sources in (Compile, doc)).value
     }
   }
 )
@@ -249,15 +255,9 @@ lazy val core = coreBase.jvm
 lazy val coreJS = coreBase.js
 
 lazy val genericBase = circeCrossModule("generic", mima = previousCirceVersion)
-  .settings(macroDependencies: _*)
+  .settings(macroSettings(scaladocFor210 = false))
   .settings(
-    libraryDependencies += "com.chuusai" %%% "shapeless" % shapelessVersion,
-    sources in (Compile, doc) := (
-      CrossVersion.partialVersion(scalaVersion.value) match {
-        case Some((2, 11)) => (sources in (Compile, doc)).value
-        case _ => Nil
-      }
-    )
+    libraryDependencies += "com.chuusai" %%% "shapeless" % shapelessVersion
   )
   .dependsOn(coreBase)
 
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -140,31 +149,53 @@ lazy val docs = project.dependsOn(core, generic, parser, optics)
...
   spray,
   hygiene,
   benchmark
-) ++ (
-  if (sys.props("java.specification.version") == "1.8") Seq[ProjectReference](java8) else Nil
 )
 
+def jvmVersionFilter: Set[Project] = sys.props("java.specification.version") match {
+  case "1.8"         => Set.empty
+  case "1.6" | "1.7" => Set(java8, optics)
```
jonas: I am not sure if `optics` is 1.8 only given it was always aggregated.

build.sbt
```diff
@@ -200,236 +231,92 @@ lazy val circe = project.in(file("."))
...
-    moduleName := "circe-generic-extras",
-    name := "Circe generic extras",
-    crossScalaVersions := scalaVersions
-  )
-  .settings(allSettings: _*)
-  .settings(macroDependencies: _*)
-  .settings(
-    sources in (Compile, doc) := (
-      CrossVersion.partialVersion(scalaVersion.value) match {
-        case Some((2, 11)) => (sources in (Compile, doc)).value
```
jonas: I removed this (also in other projects) because unidoc has its own filter and this was not updated to 2.12 so most likely it is not used.

jonas:
There are several modules which doesn't show up in the scaladoc, e.g. parser and java8 because they only expose a package object. One way to fix this is to move the implicits and defs into a trait.

build.sbt
jonas: OK, I think I found the answer. It compiles on 1.7 but doesn't run!

build.sbt
```diff
@@ -140,31 +153,53 @@ lazy val docs = project.dependsOn(core, generic, parser, optics)
...
   spray,
   hygiene,
   benchmark
-) ++ (
-  if (sys.props("java.specification.version") == "1.8") Seq[ProjectReference](java8) else Nil
 )
 
+def jvmVersionFilter: Set[Project] = sys.props("java.specification.version") match {
+  case "1.8"         => Set.empty
+  case "1.6" | "1.7" => Set(java8, optics)
```
travisbrown: The latest version of Monocle supports Java 7 for the Scala 2.10 build, but not for 2.11 or 2.12.

build.sbt
travisbrown: If I remember correctly this is about the non-Unidoc doc generation that happens automatically during publishing.

The problem is that it chokes on some of the macros here (I believe the annotation macros) on 2.10, and I was avoiding this by excluding the sources from the Javadoc artifact generation on 2.10.

You're right that 2.12 should have been added here, but the fact that it wasn't just resulted in Javadoc artifacts being empty for 2.10 _and_ 2.12:

* https://repo1.maven.org/maven2/io/circe/circe-generic_2.10/0.6.1/
* https://repo1.maven.org/maven2/io/circe/circe-generic_2.11/0.6.1/
* https://repo1.maven.org/maven2/io/circe/circe-generic_2.12/0.6.1/

This isn't a big deal, but removing the removal entirely will cause 2.10 publishing to break.

Ideally we should reinstate the removal for 2.10 only, so that 2.11 and 2.12 have proper Javadoc artifacts.

jonas: OK, thanks for clarifying. I reinstated the filter.

jonas:
@travisbrown OK, reverted the change to make more projects pure as well as the workaround to remove sources from certain macro projects for 2.10.

build.sbt
```diff
@@ -712,48 +545,14 @@ credentials ++= (
...
-  "genericExtrasJS",
-  "shapesJS",
-  "opticsJS",
-  "parserJS",
-  "refinedJS",
-  "scalajs",
-  "scodecJS",
-  "testsJS"
-)
+/* Only run optics tests on Java 8 since Scala 2.11 on Java 7 breaks. */
```
jonas: Documented here what I believe is the reason to only run optics tests on JVM 8 and made sure optics is built regardless of the JVM version.

travisbrown:
Thanks again, @jonas! I want to run through a few more checks (publishing locally and looking at artifacts, etc.) this morning, but I'll try to get this merged today.

travisbrown:
Okay, I've just spent some time poking at this, and it looks perfect except for one tiny thing—we're no longer publishing real Javadoc artifacts for circe-shapes on 2.10 even though we can (and used to). I'm happy to fix that in a follow-up PR, @jonas, unless you want to take care of it here.

jonas:
OK, fixed it in the additional commit instead of rebasing.

travisbrown:
@jonas Ah, great, thanks! Will merge when green.

dwijnand:
Lean sbt builds FTW. Jonas is a legend.

PR #478, opened 2016-11-28
fix missing bracket

xuwei-k:


travisbrown:
Thanks @xuwei-k! I wish Scaladoc would warn in situations like this.

PR #481, opened 2016-11-29
Make logo in header scale down

travisbrown:
The background logo image isn't scaled in all the browsers I've tried on my Linux desktop. This seems to fix the issue, and looks the same on OS X. /cc @jonas


jonas:
LGTM

I am considering also making a PR to sbt-microsite that will make it branding easier to allow for example SVG icons.

Issue #393, opened 2016-09-29
Long value incorrectly decoded on Scala.js


tindzk:
The following test case fails on Scala.js:

``` scala
import cats.data.Xor
import io.circe.generic.auto._
import io.circe.parser._
import io.circe.syntax._

val num = 767946224062369796L
val json = num.asJson.noSpaces
assert(decode[Long](json) == Xor.Right(num))
```

travisbrown:
Thanks @tindzk—I'll take a closer look at this, but at a glance I think [this](https://travisbrown.github.io/circe/api/index.html#io.circe.Decoder$@decodeBigDecimal:io.circe.Decoder[BigDecimal]) is what you're seeing:

> Also note that because `scala.scalajs.js.JSON` parses JSON numbers into a floating point representation, decoding a JSON number into a `BigDecimal` on Scala.js may lose precision.

Could you try `decode[JsonNumber]` and then `toLong` as a workaround?

tindzk:
Thanks! I tried the following code:

```scala
import io.circe._
import io.circe.parser._

println(767946224062369796L)
println(decode[Long]("767946224062369796"))
println(decode[JsonNumber]("767946224062369796"))
println(decode[JsonNumber]("767946224062369796").map(_.toLong))
```

Here is the JVM output:
```
767946224062369796
Right(767946224062369796)
Right(767946224062369796)
Right(Some(767946224062369796))
```

Unfortunately, it behaves differently on Scala.js:
```
767946224062369796
Right(767946224062369792)
Right(767946224062369800)
Right(Some(767946224062369792))
```

jonas:
The problem is that circe uses [`JSON.parse`](https://github.com/circe/circe/blob/991aaf596693acd9c35d400a845577232030b797/modules/parser/js/src/main/scala/io/circe/parser/package.scala#L9) parsing and decoding in Scala.js. When circe gets to [convert to its AST](https://github.com/circe/circe/blob/4fe187a123183cac09b0a7edf060c6eda689c979/modules/scalajs/src/main/scala/io/circe/scalajs/package.scala#L14) the damage is already done.

```javascript
> JSON.parse("767946224062369796")
767946224062369800
> JSON.parse("767946224062369796").toFixed()
"767946224062369792"
> JSON.parse("767946224062369796").toPrecision()
"767946224062369800"
```

I will make a PR to document the current behaviour.

The only way to fix this would be to include a "custom" JSON parser for Scala.js and let users choose between performance or correctness.

[...]

PR #488, opened 2016-12-01
Doc improvements

jonas:
As discussed on Gitter a few minor tweaks to the Scaladoc and document limitation regarding parsing of numerical values on Scala.js.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ target/
 .project
 .classpath
 tmp/
+docs/src/main/tut/contributing.md
--- a/docs/src/main/tut/parsing.md
+++ b/docs/src/main/tut/parsing.md
@@ -6,7 +6,7 @@ position: 1
 
 # Parsing JSON
 
-Circe includes a parsing module, which is a wrapper around the [Jawn][jawn] JSON parser.
+Circe includes a parsing module, which on the JVM is a wrapper around the [Jawn][jawn] JSON parser and for JavaScript uses the built-in [`JSON.parse`][json.parse].
 
 Parsing is not part of the `circe-core` module, so you will need to include a dependency on the `circe-parser` module in your build:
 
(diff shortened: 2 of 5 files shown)


docs/src/main/tut/parsing.md
```diff
@@ -6,7 +6,7 @@ position: 1
 
 # Parsing JSON
 
-Circe includes a parsing module, which is a wrapper around the [Jawn][jawn] JSON parser.
+Circe includes a parsing module, which is on the JVM is a wrapper around the [Jawn][jawn] JSON parser and for JavaScript uses the built-in [`JSON.parse`][json.parse].
```
travisbrown: Nitpick: "which ~~is~~ on the JVM is".

docs/src/main/tut/parsing.md
```diff
@@ -60,4 +60,16 @@ import cats.syntax.either._
 val json: Json = parse(rawJson).getOrElse(Json.Null)
 ```
 
+## Warnings and known issues
+
+For Scala.js, the result of parsing JSON is restricted to the expressiveness of
+the JavaScript types. In particular numerical values may [loose precision][#393]
```
travisbrown: s/loose/lose/

docs/src/main/tut/parsing.md
```diff
@@ -60,4 +60,16 @@ import cats.syntax.either._
...
+when decoded by [`JSON.parse`][json.parse] to a JavaScript number. For example:
```
travisbrown: I'm not sure "JavaScript number" is exactly the right phrasing here—it seems more like the problem is that you're trying to decode to a Scala type like `Long` that has its own semantics, even if it's implemented using JavaScript numbers on Scala.js. I'm happy to work on details like this in a follow-up PR, though.

.gitignore
```diff
@@ -8,3 +8,4 @@ target/
 .project
 .classpath
 tmp/
+docs/src/main/tut/contributing.md
```
travisbrown: Thanks—I keep meaning to open a PR for this but hadn't yet.

docs/src/main/tut/parsing.md
jonas: Thanks

docs/src/main/tut/parsing.md
jonas: OK, I've tried to clarify further. On Scala.js `Long`s are represented with [high/low pair](https://www.scala-js.org/api/scalajs-library/latest/#scala.scalajs.runtime.RuntimeLong) to ensure the Scala semantics and `StringLike.toLong` and `java.lang.long.parseLong` works correctly for the issue mentioned in #393. My understanding is that JSON.parse coerces the value into a double at which point then circe's JSON AST converter cannot do anything about it.

travisbrown:
:+1:

PR #495, opened 2016-12-07
Decoder ops

jeremyrsmith:
This PR adds two operations that I found useful:

* `Decoder#widen[B]` is just a nicer syntax for `decoder.map(a => a:B)`, to ease the fact that `Decoder` is not covariant.
* `Decoder.singletonCases` is a convenience function for constructing a `Decoder[A]` similar to how a sealed trait is encoded, except that the  descriminator strings are explicit. For example:

  ```scala
  sealed trait A
  case class B(b: String) extends A
  case class C(c: Int) extends A

  object A {
    implicit val decoder: Decoder[A] = Decoder.singletonCases {
      case "b" | "B" | "foob" => Decoder.decodeString.map(B.apply).widen[A]
      case "c" | "C" | "fooc" => Decoder.decodeInt.map(C.apply).widen[A]
    }
  }
  ```

  It just provides an easier way of defining decoders of this variety.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -3,6 +3,7 @@ package io.circe
 import cats.{ MonadError, SemigroupK }
 import cats.data.{ Kleisli, NonEmptyList, NonEmptyVector, OneAnd, StateT, Validated }
 import cats.instances.either.{ catsDataSemigroupKForEither, catsStdInstancesForEither }
+import cats.syntax.either._
 import io.circe.export.Exported
 import java.util.UUID
 import scala.annotation.tailrec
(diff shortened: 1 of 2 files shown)


travisbrown:
I've been reluctant to add an explicit `widen` given that it's available already via `cats.instances.functor._`, but at this point I'm almost convinced we should have it explicitly.

One implementation note: I'd prefer `Decoder.decoderInstances.widen(this)`, which makes the relationship clearer and also avoids a couple of allocations.

I like the `singletonCases` idea, but I'm not sure about the name, and there are a couple of points of behavior that I think we should refine:

* It definitely should preserve error accumulation from whichever decoder is used. This should be a fairly straightforward matter of writing the instance out more explicitly.
* All other decoders (except the new-ish state ones) ignore unused fields, and I wonder whether it would work to do that here—you'd just iterate over the keys until you get a match. This introduces the possibility of ambiguity and the way it's resolved may not be intuitive, but I think it's worth considering.

jeremyrsmith:
This is way out of date, so I'm going to close it. If I ever miss these methods, I'll revisit.

PR #496, opened 2016-12-07
Simplify match statement

zaneli:


travisbrown:
Thanks for catching this, @zaneli!

PR #497, opened 2016-12-09
Use applyDynamic in JsonPaths to allow more terse array indexing

paulbrimicombe:
We can use `applyDynamic` to allow simple array indexing. Given the following JSON

```json
{
  "array": [0, 1, 2, 3, 4]
}
```

this change allows us to access an element of `array` like this: `root.array(1)` rather than like this: `root.array.index(1)`

I could only find a relevant test for `JsonPath` and not for `JsonFoldPath` and `JsonTraversalPath` which I found a little worrying. Have I missed something somewhere?

### DIFF
--- a/modules/optics/src/main/scala/io/circe/optics/JsonPath.scala
+++ b/modules/optics/src/main/scala/io/circe/optics/JsonPath.scala
@@ -29,6 +29,8 @@ final case class JsonPath(json: Optional[Json, Json]) extends Dynamic {
   final def selectDynamic(field: String): JsonPath =
     JsonPath(json.composePrism(jsonObject).composeOptional(Index.index(field)))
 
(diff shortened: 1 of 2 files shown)


travisbrown:
That was fast, @paulbrimicombe!

The test situation isn't so dire since `JsonFoldPath` and `JsonTraversalPath` get exercised in `JsonPathSuite`, but it could be better (I was just talking to @julien-truffaut about this earlier this afternoon).

julien-truffaut:
I think we should also add `apply` as an alias for `index` such as we get the same syntax for a top level array

paulbrimicombe:
I absolutely agree @julien-truffaut - I can probably do that today. My kids have been ill which is why I've been offline for a couple of days...

julien-truffaut:
take your time, there is no rush!

PR #499, opened 2016-12-10
Use scala versions defined in .travis.yml

dwijnand:
Using sbt-travisci 'crossScalaVersions' and 'scalaVersion' are defined
in terms of the scala versions defined in .travis.yml.

Specifically, for 'scalaVersion' the last scala version is used
(therefore either re-order them or override it in build.sbt if it's not
the default you want). Therefore, this also changes the value of
'scalaVersion' to 2.12.1.


travisbrown:
Thanks again @dwijnand!

PR #501, opened 2016-12-10
Update Jackson to 2.8.5

jensraaby:
There are 2 code changes:
- use non generic factory method 'writerWithPrettyPrinter'
- replace deprecated exception method

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -417,8 +417,8 @@ lazy val streaming = circeModule("streaming", mima = previousCirceVersion)
 lazy val jackson = circeModule("jackson", mima = previousCirceVersion)
   .settings(
     libraryDependencies ++= Seq(
-      "com.fasterxml.jackson.core" % "jackson-core" % "2.5.3",
-      "com.fasterxml.jackson.core" % "jackson-databind" % "2.5.3"
+      "com.fasterxml.jackson.core" % "jackson-core" % "2.8.5",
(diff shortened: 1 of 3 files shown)


travisbrown:
Thanks for doing this, @jensraaby. I'd like to leave this open for a while (probably at least a week or two) to make sure that anyone who needs the older Jackson version is able to upgrade.

We should probably ask around in the Finch Gitter room in particular, since finch-circe provides support for using Jackson for parsing and printing, and this change will force all finch-circe users to be on the newer Jackson release.

travisbrown:
@jensraaby An alternative approach would be to split circe-jackson into its own repo in the new circe organization and cross-publish it for multiple Jackson releases. I don't really want to have circe-jackson25, circe-jackson26, etc. in the circe build itself, but it seems totally reasonable to promote this to a new repo and have those modules there.

jensraaby:
Thanks for the information. I've had issues with multiple Jackson versions before (like this https://github.com/twitter/twitter-server/issues/4)

jensraaby:
Sorry that link should have been  #43

travisbrown:
I'm going to go ahead and merge this, filter-branch the circe-jackson source into the new circe-jackson repo I've got set up, and then we can delete all the circe-jackson stuff in this repo in a follow-up PR.

PR #504, opened 2016-12-12
Provide Auto trait

mxl:
Sometimes it's more useful to bring `io.circe.generic.auto._` implicits into scope by not importing them but by extending package object, object or class from a trait.
For example:
```scala
package app

import io.circe.generic.Auto

package object api extends Auto {
}
```
makes implicits for automatic encoders and decoders generation visible for all code in package `app.api` and after that there is no need to import `io.circe.generic.auto._` in each source file in this package.

### DIFF
--- a/modules/generic-extras/src/main/scala/io/circe/generic/extras/auto/package.scala
+++ b/modules/generic-extras/src/main/scala/io/circe/generic/extras/auto/package.scala
@@ -1,0 +1,10 @@
+package io.circe.generic.extras
+
+/**
+ * Fully automatic configurable codec derivation.
+ *
+ * Importing the contents of this package object provides [[io.circe.Decoder]] and [[io.circe.Encoder]]
+ * instances for case classes (if all members have instances), "incomplete" case classes, sealed
+ * trait hierarchies, etc.
+ */
+package object auto extends AutoDerivation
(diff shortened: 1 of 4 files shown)


modules/generic-extras/src/main/scala/io/circe/generic/extras/auto/package.scala
```diff
@@ -0,0 +1,11 @@
+package io.circe.generic.extras
+
+/**
+ * Fully automatic configurable codec derivation.
+ *
+ * Importing the contents of this package object provides [[io.circe.Decoder]] and [[io.circe.Encoder]]
+ * instances for case classes (if all members have instances), "incomplete" case classes, sealed
+ * trait hierarchies, etc.
+ */
+package object auto extends Auto {
```
travisbrown: Nitpick but do you mind removing the empty `{}`?

modules/generic-extras/src/main/scala/io/circe/generic/extras/Auto.scala
```diff
@@ -10,11 +10,11 @@ import scala.language.experimental.macros
 /**
  * Fully automatic configurable codec derivation.
  *
- * Importing the contents of this object provides [[io.circe.Decoder]] and [[io.circe.Encoder]]
+ * Extending this trait provides [[io.circe.Decoder]] and [[io.circe.Encoder]]
  * instances for case classes (if all members have instances), "incomplete" case classes, sealed
  * trait hierarchies, etc.
  */
-final object auto {
+trait Auto {
```
travisbrown: What do you think about calling this `AutoDerivation` instead? I'd find it a bit clearer in cases where it's imported and then mixed into something like a protocol object that might have a number of other traits involved.

mxl:
@travisbrown Done! I squashed commits.

travisbrown:
@mxl Great, thanks! I'll probably publish a second 0.7 milestone in the next couple of days, and then 0.7 next week.

PR #510, opened 2016-12-15
Remove circe-jackson, which is now a stand-alone project

travisbrown:
I've split circe-jackson out into [its own project](https://github.com/circe/circe-jackson) and have cross-published it against Jackson versions 2.5, 2.6, 2.7, and 2.8 (for circe 0.6.x), so if you want to use circe with Jackson 2.8 in your project, for example, you can just add this dependency to your build:

```scala
libraryDependencies += "io.circe" %% "circe-jackson28" % "0.6.2"
```

This PR just removes all of the pieces of circe-jackson from this repo, since they now have a new home.

/cc @jensraaby


jensraaby:
Thanks Travis - this should make dependency management a bit less painful.

PR #513, opened 2016-12-16
upgrade monocle to 1.4.0-M1

julien-truffaut:


travisbrown:
How far out do you think 1.4.0 is, @julien-truffaut? I'm hoping to release 0.7.0 within the next week.

julien-truffaut:
I would like to get some feedback on the new typeclass encoding first, so it might take more than a week.

travisbrown:
I'm now thinking 0.7.0 might not happen until the beginning of January, and it wouldn't really be terrible to release it with a milestone dependency for circe-optics, anyway. Thanks @julien-truffaut!

PR #515, opened 2016-12-18
Fix broken links

n4to4:
`DESIGN.md` link works on github, but breaks on https://circe.github.io/circe/contributing.html


travisbrown:
Nice catch(es)!

PR #516, opened 2016-12-18
Remove twitter util reference in doc

n4to4:
`Twitter Util` seems to be removed since 252f2944f4de318297a2511b8d01498412819430 ?


travisbrown:
Right, thanks.

PR #517, opened 2016-12-18
zenith not built on unfiltered

xuwei-k:
/cc @sungiant


sungiant:
Correct, Zenith doesn't use Unfiltered.

travisbrown:
Ah, thanks—not sure how this ended up wrong.

PR #520, opened 2016-12-19
Remove unused references

n4to4:
`[api]` reference should be kept for future reuse?

```
$ for r in $(cut -d: -f1 ./docs/src/main/resources/microsite/includes/references.md); do lines=$(fgrep $r -r . | wc -l); if [ $lines = 1 ]; then echo $r; fi; done
[api]
[cats-xor]
[spool]
[unfiltered]
```


travisbrown:
If `[api]` isn't being used we can always add it back, so this looks good to me, @n4to4.

Issue #503, opened 2016-12-12
Fix or document `sbt test`


mxl:
Following the [discussion](https://gitter.im/circe/circe?at=584e6d0ebe9d43bc634cad21) in Gitter.

Currently it fails with:
```
sbt.ResolveException: unresolved dependency: com.typesafe.play#play-json_2.12;2.3.10: not found
unresolved dependency: io.argonaut#argonaut_2.12;6.1: not found
unresolved dependency: io.github.netvl.picopickle#picopickle-core_2.12;0.2.1: not found
unresolved dependency: io.github.netvl.picopickle#picopickle-backend-jawn_2.12;0.2.1: not found
...
sbt.ResolveException: unresolved dependency: com.typesafe.akka#akka-actor_2.12;2.3.9: not found
unresolved dependency: io.spray#spray-httpx_2.12;1.3.3: not found
unresolved dependency: io.spray#spray-routing-shapeless2_2.12;1.3.3: not found
unresolved dependency: io.spray#spray-testkit_2.12;1.3.3: not found
```

jensraaby:
I've got the same issue at the moment with IntelliJ.

IntelliJ and sbt (only for the benchmark module) appear to be looking for 2.12 artefacts that don't exist.

I actually get different errors to the above:

```
[info] Resolving jline#jline;2.14.1 ...
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	::          UNRESOLVED DEPENDENCIES         ::
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	:: com.typesafe.play#play-json_2.12;2.3.10: not found
[warn] 	:: io.argonaut#argonaut_2.12;6.1: not found
[warn] 	:: io.github.netvl.picopickle#picopickle-core_2.12;0.2.1: not found
[warn] 	:: io.github.netvl.picopickle#picopickle-backend-jawn_2.12;0.2.1: not found
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn]
[warn] 	Note: Unresolved dependencies path:
[warn] 		com.typesafe.play:play-json_2.12:2.3.10 (/Users/raabyj01/Projects/circe/circe/build.sbt#L425)
[warn] 		  +- io.circe:circe-benchmark_2.12:0.7.0-SNAPSHOT ()
[warn] 		io.argonaut:argonaut_2.12:6.1 (/Users/raabyj01/Projects/circe/circe/build.sbt#L425)
[...]

PR #521, opened 2016-12-19
Make the benchmark module work against Scala 2.12

jensraaby:
This change removes picopickle as a dependency for the benchmark module, as it is not published for Scala 2.12.

At the same time, Play JSON and Argonaut have been updated to the latest Scala 2.12 published releases.

This change resolves the IntelliJ issue with importing mentioned in https://github.com/circe/circe/issues/503

It also allows running `sbt benchmark/test` locally out of the box (i.e. with Scala 2.12).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -423,11 +423,9 @@ lazy val benchmark = circeModule("benchmark", mima = None)
   .settings(
     crossScalaVersions := crossScalaVersions.value.init,
     libraryDependencies ++= Seq(
-      "com.typesafe.play" %% "play-json" % "2.3.10",
-      "io.argonaut" %% "argonaut" % "6.1",
+      "com.typesafe.play" %% "play-json" % "2.6.0-M1",
+      "io.argonaut" %% "argonaut" % "6.2-RC2",
       "io.spray" %% "spray-json" % "1.3.2",
-      "io.github.netvl.picopickle" %% "picopickle-core" % "0.2.1",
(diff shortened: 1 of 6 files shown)


jensraaby:
I need to understand how TravisCI is running the benchmarks across versions. It looks like the new PlayJSON version is on 2.11 and 2.12 but not 2.10. D'oh.
What's the best option for the benchmarks?

travisbrown:
Hmm, I'm starting to think that the benchmarks against other libraries should live in their own project. This would be nice in a lot of ways: it's easier for people to contribute new benchmarks representing libraries or use cases they care about, it lets us keep our most dependency-heavy module out of the core repo, it puts a nicer boundary between the internal (e.g. generically derived vs. explicit instances) and external benchmarks, etc.

jensraaby:
I think you're probably right - it doesn't feel like part of the main library repo.

Would you ever want to fail a build based on the benchmark results? I'm just wondering whether TravisCI can pull from another Github repo to run the benchmarks and use the result in the main Circe build.

travisbrown:
@jensraaby I've wanted a way to do performance regression testing for a long time, but that would be against a previous version of circe itself (or the current master or whatever), not against other libraries, so if somehow that became practical it could live in the internal benchmark module, which would still be around.

jensraaby:
OK, this is made redundant by https://github.com/circe/circe-benchmarks and https://github.com/circe/circe/pull/528 so I'll close.

PR #523, opened 2016-12-20
Fix README in example

n4to4:
I thought It might be good to replace the relative path to `streaming` with the corresponding URL, so that it would be easier to grep and fix next time we change the directory structure or something.
For now, I just fixed the path anyway.


travisbrown:
:+1:, thanks!

PR #524, opened 2016-12-20
Fix incorrect links

n4to4:
Those links seem incorrect/broken


travisbrown:
Ugh, I didn't realize that the new 2.12 Scaladocs broke old links. The circe-parser one should probably point [here](https://github.com/circe/circe/tree/master/modules/parser), since it's specifically about the cross-platform parsing module, and we can't link to the API docs for that thanks to SI-8124.

n4to4:
@travisbrown Fixed 😸

travisbrown:
Thanks!

PR #526, opened 2016-12-21
No Predef

n4to4:
I'm not sure what's the status of #320 (on-hold?), but I've got this (based on #383) working.


travisbrown:
Thanks, @n4to4! This actually isn't too bad at all—I'm tempted to merge it. What do other people think?

Issue #525, opened 2016-12-20
Move benchmarks against other libraries to a separate project


travisbrown:
See #521 for context:

> Hmm, I'm starting to think that the benchmarks against other libraries should live in their own project. This would be nice in a lot of ways: it's easier for people to contribute new benchmarks representing libraries or use cases they care about, it lets us keep our most dependency-heavy module out of the core repo, it puts a nicer boundary between the internal (e.g. generically derived vs. explicit instances) and external benchmarks, etc.

So the idea is that all benchmarks against other JSON libraries would live in a new circe/circe-benchmarks repository, and only the internal benchmarks would stay in this repo.

jensraaby:
I don't know how timely I would be, but I'm willing to start working on this. 

Initially I suppose there needs to be a new repo, then a lift-and-shift of the original benchmark module and associated SBT & Travis config.

travisbrown:
@jensraaby Yep, it'd look pretty similar to what @n4to4 did for #514.

jensraaby:
I've started a new repo with just the benchmark project here: https://github.com/jensraaby/circe-benchmark

It's not ready for prime time yet, as I can't figure out why the benchmarks don't run on "sbt test" on my box. Perhaps it's another Scala version issue.

travisbrown:
Thanks, @jensraaby, looks great! I'll clone it to the circe org and add you as an owner.

travisbrown:
Okay, it's up: https://github.com/circe/circe-benchmarks

I've called it circe-benchmarks because that seems better for a stand-alone project (and is probably what I'd call the module in circe if I were naming it today—to be honest I'm not sure I remember why I went with the singular). I'm happy to reconsider this if something thinks it's really bad.

travisbrown:
Okay, with #528 this is done. Thanks @jensraaby!

PR #528, opened 2016-12-21
Remove benchmarks of 3rd party libraries

jensraaby:
See #525.

This removes the benchmarks against 3rd party JSON libraries from the main repository.


jensraaby:
@travisbrown if you prefer, I can keep the benchmarks of Circe itself in the repo and do this as a simple 'remove 3rd party benchmarks' PR?

travisbrown:
@jensraaby Yeah, right now I think `CirceDerivationBenchmark` is the only one that needs to stay, but we definitely want to keep it around in this repo (and will probably add other internal benchmarks in the future).

jensraaby:
I've restored the `CirceDerivationBenchmark` benchmark and just removed the 3rd party comparisons. Obviously I need to remove the other commit - just let me know if its OK to rebase.

travisbrown:
:+1:, looks good to me.

n4to4:
We should probably update 'docs' (eg. referece to the Benchmark.scala) in this PR or another.

travisbrown:
Sorry, @jensraaby—one last thing. Do you mind reinstating the deleted tests for the derivation suite?

@n4to4 We'll also need to add a reference in the list of circe org projects in the README, but I'm happy to have the doc updates happen in a follow-up PR.

jensraaby:
@travisbrown I've hopefully covered everything now, including updating the link in the microsite documentation. Also removed the 2.12 branching logic in .travis.yml, as that was caused by the removed dependencies.

travisbrown:
Great, thanks again, @jensraaby!

PR #534, opened 2017-01-02
String concatenation

jonas:
Some minor cleanups based on the no-Predef PR removing use of `String.format` a couple of places. 

Even though `Printer.toString` is not performance critical I made [a small benchmark](https://gist.github.com/jonas/f45c2dce32297f442084d7ff60db3896) for fun and profit. If `String.concat` is not appropriate I suggest to at least switch to use `StringContext`.


travisbrown:
👍, thanks! Are the `Test` and `Provided` symbols new? They're definitely new to me. 😄

jonas:
They have been there for some time, see: https://github.com/sbt/website/pull/306#discussion_r94271718

Issue #536, opened 2017-01-07
Byte Array Printer


ImLiar:
There is a discussion in finch https://github.com/finagle/finch/issues/676 about printing JSON straight into byte array, skipping string transformation.
I assume this performance-oriented feature could be useful in many projects besides finch.

Should it be a part of circe ecosystem? Just as another subproject printer.

travisbrown:
👍, I [definitely think we should add this](https://github.com/finagle/finch/issues/676#issuecomment-271087926). I guess the simplest approach would be to add a method that accepts a `PrintWriter` and then define `pretty` in terms of that. I'll take a look this morning.

ImLiar:
Glad to see this as a part of M2! I think it's okay to close this issue now.

PR #537, opened 2017-01-07
Support printing to arbitrary Writers

travisbrown:
Addresses #536. In addition to the new `prettyBytes` I've made a few other small changes that simplify things a bit and seem to give us a few percent more throughput in the benchmarks.

I'm not currently exposing the ability for users to print to their own `Appendable`s. I'm not totally opposed to the idea, but I'd prefer not to add methods that are fundamentally about mutation unless there's a really clear need.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Printer.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Printer.scala
@@ -2,6 +2,7 @@ package io.circe
 
 import java.io.{ BufferedWriter, ByteArrayOutputStream, OutputStreamWriter }
 import java.lang.StringBuilder
+import java.nio.ByteBuffer
 import scala.annotation.switch
 
 /**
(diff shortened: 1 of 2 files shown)


modules/core/shared/src/main/scala/io/circe/Printer.scala
```diff
@@ -118,89 +117,106 @@ final case class Printer(
...
+  }
+
+  final def prettyBytes(json: Json): Array[Byte] = {
+    val bytes = new ByteArrayOutputStream()
+    val writer = new BufferedWriter(new OutputStreamWriter(bytes, "UTF-8"))
+
+    printJsonAtDepth(writer)(json, 0)
+
+    writer.close()
+    bytes.toByteArray
```
vkostyukov: I wish we could avoid [copying here](https://docs.oracle.com/javase/7/docs/api/java/io/ByteArrayOutputStream.html#toByteArray()). Even though there is no API to retrieve the underlying buffer directly, people seem to use [this simple trick](http://stackoverflow.com/a/15686667) to do that. Please, note that this means we'd need to change the return type to `ByteBuffer` since the underlying `Array[Byte]` may contain more bytes than we need so we have to crop it.

I think we can also expose that as a standalone method (`printByteBuffer`) if we're worried about this extra complexity. What do you think @travisbrown?

travisbrown:
Ah, good point, @vkostyukov. I've just pushed an update—what do you think?

vkostyukov:
Looks great! I really like how simple `EnhancedByteArrayOutputStream` could be.

travisbrown:
Ugh, was just running Scala.js tests for the 0.7.0-M2 release and get this:

```scala
[error] Referring to non-existent method java.io.BufferedWriter.<init>(java.io.Writer)
[error]   called from io.circe.Printer.prettyByteBuffer(io.circe.Json)java.nio.ByteBuffer
```

I guess we'll have to make this a JVM-only method—I'll get started on that asap.

Issue #542, opened 2017-01-12
Printing int arrays is slower when targeting byte buffers


vkostyukov:
While working on https://github.com/finagle/finch/pull/717 I figured that printing int arrays is 25% slower in Circe when targeting byte buffers instead of strings. Intuitively it should be the opposite. Interesting enough, the allocation profile is improving significantly with byte buffers, but it's the throughput that degrades.

vkostyukov:
I think this hurt Finch's results pretty dramatic in pre r14: https://www.techempower.com/benchmarks/previews/round14/r13-vs-r14p3.html

I was trying to look into that and I can certainly confirm that printing to `ByteBuffer`s is always better for more-or-less "big" JSON objects. TechEmpower's `{ "message": "Hello, World!"}` isn't that big.

I'm going to spend some time digging into that but in the meantime, I want to do change our benchmark in a way that string-based encoding is used. Presumably passing a non UTF-8 charset would be enough as a workaround.

vkostyukov:
I modified the Printing benchmark to also take this "Hello World" object as an input (to more-or-less simulate what TE does) and the results aren't surprising:

```
[info] Benchmark                                       Mode  Cnt        Score        Error  Units
[info] PrintingBenchmark.printBooleansToByteBuffer    thrpt   20    29596.013 ±   2995.222  ops/s
[info] PrintingBenchmark.printBooleansToString        thrpt   20    32970.890 ±   3993.489  ops/s
[info] PrintingBenchmark.printFoosToByteBuffer        thrpt   20     3455.862 ±     79.134  ops/s
[info] PrintingBenchmark.printFoosToString            thrpt   20     3756.336 ±    302.617  ops/s
[info] PrintingBenchmark.printHelloWorldToByteBuffer  thrpt   20   384342.251 ±  22680.948  ops/s
[info] PrintingBenchmark.printHelloWorldToString      thrpt   20  5747159.050 ± 295483.892  ops/s
[info] PrintingBenchmark.printIntsToByteBuffer        thrpt   20    20184.448 ±    657.650  ops/s
[info] PrintingBenchmark.printIntsToString            thrpt   20    22450.036 ±    478.925  ops/s
```

PR #544, opened 2017-01-12
Add printing benchmarks and rearrange some things

travisbrown:
First stab at investigating #542, but the results aren't very surprising / helpful:

```
Benchmark                                     Mode  Cnt      Score      Error  Units
PrintingBenchmark.printBooleansToByteBuffer  thrpt   20  31892.871 ± 2119.524  ops/s
PrintingBenchmark.printBooleansToString      thrpt   20  37532.772 ± 3550.116  ops/s
PrintingBenchmark.printFoosToByteBuffer      thrpt   20   2839.652 ±  221.515  ops/s
PrintingBenchmark.printFoosToString          thrpt   20   3496.935 ±   39.529  ops/s
PrintingBenchmark.printIntsToByteBuffer      thrpt   20  17816.349 ± 1988.353  ops/s
PrintingBenchmark.printIntsToString          thrpt   20  17039.468 ± 4612.298  ops/s
```

### DIFF
--- a/modules/benchmark/src/main/scala/io/circe/benchmark/ExampleData.scala
+++ b/modules/benchmark/src/main/scala/io/circe/benchmark/ExampleData.scala
@@ -1,0 +1,17 @@
+package io.circe.benchmark
+
+import io.circe.Json
+import io.circe.syntax._
+
+class ExampleData {
+  val ints: List[Int] = (0 to 1000).toList
+  val booleans: List[Boolean] = ints.map(_ % 2 == 0)
+
+  val foos: Map[String, Foo] = List.tabulate(100) { i =>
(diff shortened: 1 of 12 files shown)


vkostyukov:
It's quite interesting that string-based print is nearly always better. My guess is byte-buffers are faster when the data set is big enough.

travisbrown:
Added a small optimization for JSON arrays (`Vector#iterator` seems a little faster than iterating over indices by hand).

jonas:
@travisbrown I was playing with this yesterday and although it doesn't change anything for the benchmarks I wonder if it would make sense to change `Printer.MemoizedPieces` to always return the same instance when `indent.isEmpty` or somehow inject the `pieces` member so that only one is ever used for the `noSpaces` case.

travisbrown:
@jonas Makes sense—I've just pushed a new commit with that and a few other little changes.

PR #545, opened 2017-01-15
Move time-related instances to trait in io.circe.java8.time

ImLiar:
Same logic as `AutoDerivation`


travisbrown:
Looks reasonable to me. My one reservation is that eventually we'll drop Java 7 and these instances will move into the `Encoder` and `Decoder` companion objects, and I'd prefer not to maintain a separate trait forever. Since there's no real timeline for dropping Java 7 I'm happy to see this go in now, on the understanding that it may be deprecated in the future.

PR #550, opened 2017-01-18
Add java.time.Period encoder/decoder

cakper:


travisbrown:
👍, thanks @cakper! Unfortunately one of the drawbacks of the new `TimeInstances` trait is that we can't add methods without breaking binary compatibility, so this will have to wait for 0.8.0.

cakper:
@travisbrown sure, no worries - I've included this code in my project for now anyway, so it's not a blocker :)

PR #553, opened 2017-01-21
Update versions in doc

n4to4:


travisbrown:
Thanks much @n4to4! I've been meaning to do this but it's been a busy week…

Issue #484, opened 2016-11-30
Add an fs2 module


travisbrown:
This would probably look similar to circe-streaming but would use [fs2](https://github.com/functional-streams-for-scala/fs2) instead of [iteratee.io](http://iteratee.io).

fsvehla:
Is there also interest in a version for Monix observables?

fommil:
looks like I may have need of a `Pipe[F, Byte, T]` and (maybe) a `Pipe[F, T, Byte]` which would be constructed if there is an implicit (circe equivalent of) `JsonReader[T]` around. Or would it be better to provide `Pipe[F, Byte, JsonValue]` and then a separate `Pipe[F, JsonValue, T]`?

fommil:
ah, looks like @fsvehla already did the latter! `Pipe[F, Chunk[Byte], Json]` (although I think fs2 may need to chunk up the bytes in that case)

travisbrown:
This is now available [in a separate repo](https://github.com/circe/circe-fs2) in the circe org.

PR #554, opened 2017-01-21
Add fs2 module

n4to4:
#484

Tried to port the iteratee.io version and it seems to be working.
Cross compiling for 2.10.x is failing because fs2 doesn't support 2.10.x ?.
And I couldn't find a way to fix it (maybe sbt-doge?), so just left it for now.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -179,7 +179,8 @@ lazy val circeJsModules = Seq[Project](scalajs)
 lazy val circeJvmModules = Seq[Project](
   jawn,
   java8,
-  streaming
+  streaming,
+  fs2
 )
 
 lazy val circeDocsModules = Seq[Project](docs)
@@ -375,7 +376,7 @@ lazy val testsBase = circeCrossModule("tests", mima = None)
     coverageExcludedPackages := "io\\.circe\\.tests\\..*"
   )
   .jvmSettings(fork := true)
-  .jvmConfigure(_.dependsOn(jawn, streaming))
+  .jvmConfigure(_.dependsOn(jawn, streaming, fs2))
   .jsConfigure(
     _.settings(
       libraryDependencies += "org.spire-math" %% "jawn-parser" % jawnVersion % "compile-time"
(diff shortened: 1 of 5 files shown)


travisbrown:
Thanks—I'll review in detail tomorrow.

I was really, really happy to be able to get rid of sbt-doge last month, and I'm not excited about bringing it back, but I guess we'll have to.

n4to4:
@travisbrown what's the status? should I have to do something (sbt-doge settings)? I will close this if it's not useful.

travisbrown:
No, it's definitely useful, @n4to4, but we will either need to reinstate sbt-doge or make this a separate circe organization repository. I don't have a strong preference, but I think it fits better here, even if it means we have to go back to sbt-doge.

travisbrown:
@n4to4 I'd like to go ahead and publish a circe 0.8.0 release in the next few days, and I definitely want to include this. Would you have time to handle the doge-ification? If not I should be able to get to it this weekend. /cc @fommil

n4to4:
@travisbrown I'll give it a try

fommil:
the one thing I'd say about this impl is that it can be a bit of a pain to have to supply the type params (often the compiler fails to infer the types), can you think of a way that this could be simplified? (not the end of the world if not), e.g. in `fs2-http`

```scala
    clientTask.flatMap { client =>
      client.request(backlogRequest).flatMap { resp =>
        resp.body.chunks.through(byteParser).through(decoder[Task, Backlog])
      }.runLast.map(_.get)
    }
```

it would be nice to be able to just say `decoder` or `decoder[Backlog]` instead of having to mention the effect. But at least there are no other type params, so maybe I'm just being pedantic.

travisbrown:
@n4to4 I think I've decided to do this as a new repo, if that's okay with you. I've `filter-branch`-ed your commits and added the build stuff [here](https://github.com/circe/circe-fs2).

n4to4:
@travisbrown No problem at all 👍

travisbrown:
Okay, closing since this is live at [circe/circe-fs2](https://github.com/circe/circe-fs2).

PR #557, opened 2017-01-26
Add durban/seals to the list of related projects

durban:
As discussed in https://github.com/durban/seals/issues/41, this commit adds the project to the list of related projects in `README.md`.


travisbrown:
Thanks!

PR #571, opened 2017-02-20
docs: explicit encoding/decoding in parsing doc

aaronlevin:
Give a simple example of encoding and decoding explicitly. There are
many examples of the web of how to do decoding explicitly, but few on
how to do Encoding. This simple example does not expose the whole API,
but will give people a starting point to dig deeper.

PS - it's not clear to me if this is the most idiomatic way (I know there are the `tupleN` methods, but this seems a little more low-level / basic in a better/education way imo)

### DIFF
--- a/docs/src/main/tut/codec.md
+++ b/docs/src/main/tut/codec.md
@@ -120,14 +120,23 @@ do so in a couple of ways.
 Firstly, you can write a new `Encoder[A]` and `Decoder[A]` from scratch:
 
 ```tut:book
-class Thing()
+class Thing(val foo: String, val bar: Int)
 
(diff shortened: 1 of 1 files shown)


aaronlevin:
I have no idea why this is failing.

travisbrown:
@aaronlevin The failure is because Scoverage doesn't give stable results and by default Codecov fails the build on any decrease. This is only intermittently an error but it's still a bad experience for contributors and we need to fix it (I think by adding a Codecov config file—I'll either do that now or open an issue).

In any case this looks great to me even if Codecov doesn't like it, so thanks. :smile:

PR #573, opened 2017-02-22
Add .codecov.yml and set threshold

### LINKED ISSUE #572: Make Codecov less enthusiastic about failing the build
As discussed in #571, Codecov sometimes fails builds unnecessarily. I think we just need to add a `.codecov.yaml` file to raise the threshold for what counts as a failure. I'd be happy to have this be quite high (like 5%), given that we'll still get the report and major changes can be addressed in code review.

n4to4:
Fixes #572 

See [.codecov.yml spec](https://github.com/codecov/support/wiki/Codecov-Yaml#full-yaml)

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,5 @@
+coverage:
+  status:
+    project:
+      default:
+        threshold: 5%


travisbrown:
Thanks for researching this, @n4to4!

PR #577, opened 2017-03-01
Update sbt plugins

n4to4:


travisbrown:
Thanks for all these PRs, @n4to4!

PR #580, opened 2017-03-04
Change CrossVersion.full to use CrossVersion.patch for Typelevel Scala compatibility.

ClydeMachine:
Including text from the issue at [typelevel/scala#135](https://github.com/typelevel/scala/issues/135): 

> To use Typelevel Scala versions which are not exactly aligned with the corresponding Lightbend Scala version (ie. because they include an additional -bin-patch-nnn suffix in their version) we need to modify project builds which use CrossVersion.full (which includes the suffix) to use CrossVersion.patch (which doesn't) where appropriate (eg. for macro-paradise and other compiler plugins).

Replaced the aforementioned, as well as updating the use of "org.scala-lang" to scalaOrganization.value in library dependencies.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -209,8 +209,8 @@ lazy val aggregatedProjects: Seq[ProjectReference] =
 
 def macroSettings(scaladocFor210: Boolean): Seq[Setting[_]] = Seq(
   libraryDependencies ++= Seq(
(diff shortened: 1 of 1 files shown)


travisbrown:
Thanks @ClydeMachine!

PR #587, opened 2017-03-10
Define Iso[JsonObject, List[(String, Json)]]

benhutchison:
Based on gitter discussion with @travisbrown: https://gitter.im/circe/circe?at=58bc792221d548df2c94ea01

Currently there is a failing test which generates over [1000 lines of output when it fails](https://gist.github.com/benhutchison/5e096c5725dcd5b581b7c4c2261b77a3#file-propertybasedtestingisawesome-log); I cant make head or tail of what's wrong, except to suggest that pruning the test data generation somehow will probably fix it.

### DIFF
--- a/modules/optics/src/main/scala/io/circe/optics/JsonObjectOptics.scala
+++ b/modules/optics/src/main/scala/io/circe/optics/JsonObjectOptics.scala
@@ -2,7 +2,7 @@ package io.circe.optics
 
 import cats.instances.list.catsStdInstancesForList
 import io.circe.{ Json, JsonObject }
-import monocle.{ Lens, Traversal }
+import monocle.{ Iso, Lens, Traversal }
 import monocle.function.{ At, Each, FilterIndex, Index }
 import scalaz.{ Applicative, Traverse }
 import scalaz.std.ListInstances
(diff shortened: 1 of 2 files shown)


travisbrown:
Ouch. We clearly need better `Shrink` instances.

The problem is that I was wrong, and the tests caught that—these two types aren't actually isomorphic, since a list with duplicate keys won't roundtrip. I guess we could provide the `Iso` anyway if we clearly document that it's law-breaking, or we could instead provide an `Iso` from `JsonObject` to `Map[String, Json]`.

benhutchison:
Iso from JsonObject to Map[String, Json] sounds good. Does this seem right? 
```
  final lazy val jsonObjectIso: Iso[JsonObject, collection.immutable.Map[String, Json]] =
    Iso((_: JsonObject).toMap)(JsonObject.fromMap)
```

If you recall, my use case was to traverse the Key-Value pairs of a JsonObject. IIUC the other piece needed would be a `Traversal[Map[String, Json], (String, Json)]]`, to compose with the `Iso` above.
```
  final lazy val jsonObjectFields: Traversal[JsonObject, (String, Json)] =
    jsonObjectIso.composeTraversal(???)
```

I wasn't able to write one that compiled.

travisbrown:
@benhutchison I think the following should work:

```scala
import io.circe.Json
import scalaz.std.map._
import monocle.Traversal

val stringMapTraversal: Traversal[Map[String, Json], Json] =
  Traversal.fromTraverse[({ type L[x] = Map[String, x] })#L, Json]
```

And yeah, that `Iso` looks good to me.

benhutchison:
@travisbrown The `Traversal[Map[String, Json], Json]` you sketched doesn't include the names of the json fields...?

But its getting the names of the fields, `Traversal[Map[String, Json], (String, Json)]`, thats my requirement. Just getting the field _values_ is already well supported by circe.

I know that key-values isn't as easy: `Traversal.fromTraverse` expects a `T[A]` and an `A`, but Map[K, V] doesnt have the right shape to be a T. And the scalaz traverse instance just walks the values.

Maybe this trail implies the `Each` instance I [sketched on gitter back in March](https://gitter.im/circe/circe?at=58bb83a7f1a33b62757f3c25) is worthwhile?

travisbrown:
@benhutchison Hmm, yeah, sorry to come full circle but if that `Each` instance is lawful I'm happy to move forward with it.

benhutchison:
Activity moving to #612

PR #595, opened 2017-03-22
Properly encode floats; fixes #594

### LINKED ISSUE #594: Encode[Float] causes floating point errors
It seems the implementation for `Encode[Float]` uses `.toDouble`. This causes floating point errors when encoding to JSON.

```
scala> Encoder[Float].apply(1.2f)
res2: io.circe.Json = 1.2000000476837158
```

carymrobbins:


travisbrown:
Thanks! I'll aim to get this reviewed and merged today.

carymrobbins:
Heads up, I made a small change to `JsonFloat.toLong` that should make it a little simpler. In fact, its implementation could probably also be used for `JsonDouble.toLong`

carymrobbins:
I'm also adding some tests to appease the code coverage report.

carymrobbins:
Alright, looks like I've finally passed all of the build checks.

travisbrown:
Thanks, @carymrobbins. I'm looking at this more closely now and I'm wondering how bad it would be just to have the `Float` encoder go via e.g. `Double.parseDouble(Float.toString(1.2F))` instead of introducing a whole new `JsonFloat` representation. I definitely agree that this is a bug, but it seems like a shame to add so much code to support the use of `Float`.

carymrobbins:
@travisbrown - It doesn't sound too bad. That's just and extra 4 bytes before it gets converted to BigDecimal, Long, etc, but ideally I'd imagine that overhead to be fine. If that's what you'd like, I can rework the PR to go that route instead.

carymrobbins:
@travisbrown - Ok, I went ahead and updated the PR to use your suggested logic. All checks have passed, so let me know what you think.

travisbrown:
@carymrobbins Looks good to me—thanks much!

travisbrown:
Hey @carymrobbins, do you happen to have a branch around with your original `JsonFloat` version? I think we're going to have to go that route, since there are corner cases where converting to a string and then a double still doesn't work.

If it's not convenient I can reconstruct it, but I thought I'd check first. Sorry for second-guessing your original approach. :smile:

carymrobbins:
@travisbrown - I had a feeling, so yeah, I have a backup of it 😂

Want me to submit it as a new PR?

travisbrown:
@carymrobbins Sure, or you can just point me to the branch and I'll merge it with the rearrangements I made yesterday.

carymrobbins:
I just dealt with the merge conflicts, so to make it easier to review I've submitted it in PR #603.

PR #597, opened 2017-03-24
No default configuration for @ConfiguredJsonCodec

reactormonk:
There should be no default configuration for `@ConfiguredJsonCodec`. When you choose the use that instead of `@JsonCodec`, you always want to specify your own configuration.

This is mainly a problem with intellij, that tells you importing the config from somewhere is an unused import statement. And cleans it up, which means your code breaks without any warning.

### DIFF
--- a/modules/generic-extras/src/main/scala/io/circe/generic/extras/Configuration.scala
+++ b/modules/generic-extras/src/main/scala/io/circe/generic/extras/Configuration.scala
@@ -17,3 +17,7 @@ final object Configuration {
     "$1_$2"
   ).replaceAll("([a-z\\d])([A-Z])", "$1_$2").toLowerCase
 }
+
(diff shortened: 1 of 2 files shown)


travisbrown:
I'd be interested to get some community feedback on this change. If we go forward with it I think we should provide a separate import that puts an implicit default into scope without the need for an `implicit val`.

reactormonk:
Could we go ahead with this? I really don't like the idea of typelevel things misbehaving without compiler errors.

travisbrown:
@reactormonk Okay. Nobody's complained, so I think we could go forward with this in 0.8.0 (which I'm actually really truly planning to publish this week).

One thing—what do you think about renaming `DefaultConfiguration` to `defaults`? I think that would match our other naming conventions a little more closely.

PR #605, opened 2017-04-03
Fix JsonNumber equality

travisbrown:
The tl;dr for this PR is that there's some weird behavior for `JsonNumber` in some unusual cases if you're doing unusual things. Most users should be unaffected by both these changes and the weirdnesses they're intended to smooth out.

There are currently a couple of ways that the `Eq` instance for `JsonNumber` isn't transitive. The first involves negative zero:

```scala
scala> import cats.syntax.eq._, io.circe.Json
import cats.syntax.eq._
import io.circe.{Json, JsonNumber}

scala> val Some(lz) = Json.fromLong(0L).asNumber
lz: io.circe.JsonNumber = 0

scala> val Some(dz) = Json.fromDouble(0.0).flatMap(_.asNumber)
dz: io.circe.JsonNumber = 0.0

scala> val Some(nz) = Json.fromDouble(-0.0).flatMap(_.asNumber)
nz: io.circe.JsonNumber = -0.0

scala> lz === dz
res0: Boolean = true

scala> lz === nz
res1: Boolean = true

scala> dz === nz
res2: Boolean = false
```

The other involves comparing instances created from `Double`s and `Long`s once you get past the point that `Double` can represent integral values exactly:

```scala
scala> val bigLong = Long.MaxValue
bigLong: Long = 9223372036854775807

scala> val prettyBigLong = Long.MaxValue - 1L
prettyBigLong: Long = 9223372036854775806

scala> val Some(bigLongJson) = Json.fromLong(bigLong).asNumber
bigLongJson: io.circe.JsonNumber = 9223372036854775807

scala> val Some(prettyBigLongJson) = Json.fromLong(prettyBigLong).asNumber
prettyBigLongJson: io.circe.JsonNumber = 9223372036854775806

scala> val Some(bigDoubleJson) = Json.fromDouble(bigLong.toDouble).flatMap(_.asNumber)
bigDoubleJson: io.circe.JsonNumber = 9.223372036854776E18

scala> bigLongJson === bigDoubleJson
res3: Boolean = true

scala> prettyBigLongJson === bigDoubleJson
res4: Boolean = true

scala> bigLongJson === prettyBigLongJson
res5: Boolean = false
```

In both cases the current equality checks simply delegate to `==` on the various combinations of primitive values, which have the same problems with transitivity.

The fix in this PR makes all equality checks go via the `BiggerDecimal` representation except for the cases where we're comparing the same kind of value. This simplifies things a lot, and gives us transitivity, and the only cost is that equality checks are more expensive in some cases. Checking equality on `JsonNumber` isn't a terribly common operation, though, so I don't care much about the expense.

Setting aside equality, these changes do mean that the behavior of `toLong` will be different than in previous circe versions in certain situations. For example, this is 0.7.1:

```scala
scala>  val value = (Long.MaxValue - 525).toDouble
value: Double = 9.2233720368547748E18

scala> Json.fromDouble(value).flatMap(_.asNumber).flatMap(_.toLong)
res6: Option[Long] = Some(9223372036854774784)
```

And here's the new behavior:

```scala
scala> val value = (Long.MaxValue - 525).toDouble
value: Double = 9.2233720368547748E18

scala> Json.fromDouble(value).flatMap(_.asNumber).flatMap(_.toLong)
res0: Option[Long] = Some(9223372036854774800)
```

This follows our new policy of having all conversions from floating point numbers (i.e. both `Double` and `Float`) go via the `toString` for those types.

/cc @carymrobbins

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/JsonNumber.scala
+++ b/modules/core/shared/src/main/scala/io/circe/JsonNumber.scala
@@ -2,6 +2,7 @@ package io.circe
 
 import cats.Eq
 import io.circe.numbers.BiggerDecimal
+import java.math.{ BigDecimal => JavaBigDecimal }
 
 /**
  * A JSON number with optimization by cases.
@@ -189,15 +190,21 @@ private[circe] final case class JsonLong(value: Long) extends JsonNumber {
  */
 private[circe] final case class JsonDouble(value: Double) extends JsonNumber {
   private[circe] def toBiggerDecimal: BiggerDecimal = BiggerDecimal.fromDouble(value)
-  final def toBigDecimal: Option[BigDecimal] = Some(BigDecimal(value))
-  final def toBigInt: Option[BigInt] = toBigDecimal.flatMap { d =>
-    if (d.isWhole) Some(d.toBigInt) else None
+  private[this] def toJavaBigDecimal = JavaBigDecimal.valueOf(value)
+
+  final def toBigDecimal: Option[BigDecimal] = Some(toJavaBigDecimal)
+  final def toBigInt: Option[BigInt] = {
+    val asBigDecimal = toJavaBigDecimal
+
+    if (JsonNumber.bigDecimalIsWhole(asBigDecimal)) Some(new BigInt(asBigDecimal.toBigInteger)) else None
   }
+
   final def toDouble: Double = value
 
   final def toLong: Option[Long] = {
-    val asLong: Long = value.toLong
-    if (asLong.toDouble == value) Some(asLong) else None
+    val asBigDecimal = toJavaBigDecimal
+
+    if (JsonNumber.bigDecimalIsValidLong(asBigDecimal)) Some(asBigDecimal.longValue) else None
   }
 
   final def truncateToLong: Long = value.toLong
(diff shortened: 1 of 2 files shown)


modules/core/shared/src/main/scala/io/circe/JsonNumber.scala
```diff
@@ -264,20 +273,20 @@ final object JsonNumber {
     if (result.eq(null)) None else Some(JsonBiggerDecimal(result))
   }
 
+  private[this] val bigDecimalMinLong: JavaBigDecimal = new JavaBigDecimal(Long.MinValue)
+  private[this] val bigDecimalMaxLong: JavaBigDecimal = new JavaBigDecimal(Long.MaxValue)
+
+  private[circe] def bigDecimalIsWhole(value: JavaBigDecimal): Boolean =
+    value.signum == 0 || value.scale <= 0 || value.stripTrailingZeros.scale <= 0
```
SamSchlegel: Might want to also check `value.compareTo(JavaBigDecimal.ZERO) == 0` due to http://bugs.java.com/bugdatabase/view_bug.do?bug_id=6480539

SamSchlegel: value.signum should handle this though

travisbrown: Yeah, I had problems with that and other Java <8 `stripTrailingZeros` bugs back when I was first putting together `BiggerDecimal`, but I think this should always be fine (because of the `signum` check and since the scale won't be close to `Int.MaxValue`).

modules/core/shared/src/main/scala/io/circe/JsonNumber.scala
```diff
@@ -208,21 +215,23 @@ private[circe] final case class JsonDouble(value: Double) extends JsonNumber {
...
-    if (bigInt.floatValue() == value) Some(bigInt) else None
+  final def toBigInt: Option[BigInt] = {
+    val asBigDecimal = toJavaBigDecimal
+
+    if (JsonNumber.bigDecimalIsWhole(asBigDecimal)) Some(new BigInt(asBigDecimal.toBigInteger)) else None
   }
 
-  // Don't use value.toFloat due to floating point errors.
-  final def toDouble: Double = toJavaBigDecimal.doubleValue()
+  // Don't use `value.toFloat` due to floating point errors.
```
carymrobbins: This was my fault, but maybe next time you or I touch this code, we can change `toFloat` to `toDouble` in the comment.

travisbrown: :+1:, thanks for noticing. I want to add a class abstracting over `JsonFloat` and `JsonDouble` but MiMa didn't like it, so I was planning to do it separately, and can take care of this then.

carymrobbins: Sounds great, feel fee to send it my way and I can take a look at it as well.

PR #612, opened 2017-04-25
Each instance for key-value pairs of a Json object

benhutchison:
Replaces #587, re discussion in earlier PR. 

This instance doesnt seem lawful. My guess is, like in #587, its failing on the case of duplicate keys in a json object, which seems to clash with the principle that for an `Each[S, A]`, [`A` is supposed to be unique for a given `S`](https://github.com/julien-truffaut/Monocle/blob/master/core/shared/src/main/scala/monocle/function/Each.scala#L11).

Whats confusing me is that there's an exisiting [`Each` instance](https://github.com/circe/circe/blob/master/modules/optics/src/main/scala/io/circe/optics/JsonObjectOptics.scala#L17) for the fields (without keys) of a json object. But if the fields are type `A` above, they are not guaranteed to be unique either. So why is the existing instance lawful I wonder?

### DIFF
--- a/modules/optics/src/main/scala/io/circe/optics/JsonObjectOptics.scala
+++ b/modules/optics/src/main/scala/io/circe/optics/JsonObjectOptics.scala
@@ -22,6 +22,17 @@ trait JsonObjectOptics extends CatsConversions with ListInstances {
     }
   }
 
+  implicit final lazy val objectEachKV: Each[JsonObject, (String, Json)] = new Each[JsonObject, (String, Json)] {
+    final def each: Traversal[JsonObject, (String, Json)] = new Traversal[JsonObject, (String, Json)] {
+      final def modifyF[F[_]](f: ((String, Json)) => F[(String, Json)])(from: JsonObject)(implicit
+        F: Applicative[F]
+      ): F[JsonObject] =
+        F.map(from.toList.reverse.foldLeft(F.pure(List.empty[(String, Json)])) {
+          case (acc, kv) => F.apply2(acc, f(kv)){case (list, elem) => elem :: list}
+        })(JsonObject.from(_))
+    }
+  }
+
   implicit final lazy val objectAt: At[JsonObject, String, Option[Json]] =
     new At[JsonObject, String, Option[Json]] {
       final def at(field: String): Lens[JsonObject, Option[Json]] =
--- a/modules/optics/src/test/scala/io/circe/optics/OpticsSuite.scala
+++ b/modules/optics/src/test/scala/io/circe/optics/OpticsSuite.scala
@@ -14,6 +14,7 @@ import scalaz.std.math.bigInt._
 import scalaz.std.option._
 import scalaz.std.string._
 import scalaz.std.vector._
+import scalaz.std.tuple._
 
 class OpticsSuite extends CirceSuite {
   implicit val equalJson: Equal[Json] = Equal.equal(Eq[Json].eqv)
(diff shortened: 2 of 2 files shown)


travisbrown:
Okay, I've finally found some time to take a closer look at this, and there are a couple of issues. The first is that we shouldn't really have multiple `Each` instances for `JsonObject` (this is what the docs mean by `A` being unique for a given `S`), and the second is the unlawfulness of this `Each`, which is due to the fact that modification can remove elements (by mapping multiple inputs to outputs with the same key).

I don't really know what we should do. I'd consider offering an unlawful `Iso` (like in #587) if it's very, very clearly marked, but I don't really want to have unlawful type class instances (especially if they raise ambiguities with lawful ones). What do you think, @benhutchison?

benhutchison:
I agree the unlawfulness of `Each` is a problem; it appears that a `Fold[JsonObject, (String, Json)]` that _got_ the json field KVs, but wouldn't allow _modification_, would be lawful...? 

That would meet my use case because Im not updating the json doc but rather emitting a new structure based on it.

Im not sure I see the reason for "we shouldn't really have multiple Each instances for JsonObject", providing their scope can be managed .. is that due to the _Coherence_  argument? I'll have to declare myself an unbeliever in the "Coherence-is-necessary" claim in general. But Circe is your codebase, so if that's your preferred design principle, I support your right to enforce it.

travisbrown:
@benhutchison The `Fold` sounds like it should work, or at least my intuition is that it can be done lawfully. Do you want to take a shot? I'd be happy to include it in 0.8.0 if it works.

About coherence: having differently-typed `Each`s with the same first type parameter is a little more dramatic than simply breaking coherence (which I agree isn't always a terrible thing to do).

If you translate having both an `Each[JsonObject, Json]` and an `Each[JsonObject, (String, Json)]` into subtyping terms it'd be like saying `JsonObject` is both a `Seq[Json]` and a `Seq[(String, Json)]` at the same time. In an `Each[S, A]` the `S` is supposed to uniquely determine the `A` because the type class is capturing the containerness of `S`, and `S` should presumably only be a container of one type of thing.

Coherence would just say "you can't have two different ways of traversing the `Json` values in a `JsonObject` with two different `Each[JsonObject, Json]` instances in two different parts of your program. It's concerned with prohibiting multiple ways to traverse a single type of value, not the possibility that you could traverse multiple types altogether.

Having two `Each`s also puts a burden on usability in that instead of `.composeTraverse(each)` you have to write out the type arguments on `each` to disambiguate between the two.

benhutchison:
Yes I'll have a shot at the `Fold`.

WRT multiple instances for different types: to me `Each` differs from subtyping in that its a projection ("can be viewed as"), and there are multiple ways to view the fields of Json object, as a sequence of:
- `(String, Json)`: keys and values. As we've seen this seems a valid `Fold` but may not roundtrip if two keys are mapped onto the same value. In a dependently typed language we might attempt to express a uniqueness requirement among the mapped keys, but impractical in Scala I think.
- `Json`: just the values (internally the lens disambiguates identical values by ordinal position)
- `(Int, Json)`: the values, plus their ordinal position made explicit as an Int
- We could reverse the tuple order to get 2 more, `(Json, String)` and `(Json, Int)` 

So I can't see how one particular type can claim, mathematically, to be _the_ canonical `Each`. Practically, I agree there's a need to choose one as the default `Each` traversal provided for Json.

Issue #551, opened 2017-01-19
BitVector encoder doesn't preserve length


durban:
The encoder for `scodec.bits.BitVector` doesn't preserve the length of the bit vector (if it isn't a multiple of 8), but right pads it with zero bits (to be a multiple of 8). Thus, encoding and decoding is not (always) an identity.

```scala
scala> import scodec.bits._
import scodec.bits._

scala> import io.circe.syntax._
import io.circe.syntax._

scala> import io.circe.scodec._
import io.circe.scodec._

scala> val b = bin"010110"
b: scodec.bits.BitVector = BitVector(6 bits, 0x58)

scala> b.asJson.as[BitVector]
res0: io.circe.Decoder.Result[scodec.bits.BitVector] = Right(BitVector(8 bits, 0x58))
```

travisbrown:
Eek! Thanks for catching this. I probably won't get around to fixing it until Sunday, if someone else wants to take a look before then. Once it's done I'll publish an 0.7.1 release.

note:
Made a simple fix on my branch, still need to do some research as it's not perfect. Will get back to this tomorrow

travisbrown:
@note Thansk! I guess it's obvious that this totally slipped off my radar in January…

durban:
Fixed in 0.8.0.

PR #614, opened 2017-04-25
551 BitVector encoder/decoder fix

note:
refs #551 , the root of the problem was that base64 was used. It adds padding to 8 bits which leaded to not preserving length.

The solution is to add one byte header which contains number of bits of data in the last byte of base64 representation.

My first solution used `toBin` which has very simple implementation but such encoding is very inefficient in terms of resulting string size. That solution can be found here: https://github.com/note/circe/commit/c7bd54ad124f12afd5a3712fff5470ba38744beb.

@travisbrown This is my first contribution to circe and I may not be aware of some internal conventions, please be forgiving :)

### DIFF
--- a/modules/scodec/shared/src/main/scala/io/circe/scodec/package.scala
+++ b/modules/scodec/shared/src/main/scala/io/circe/scodec/package.scala
@@ -1,23 +1,65 @@
 package io.circe
 
-import _root_.scodec.bits.{BitVector, ByteVector}
+import _root_.scodec.bits.{ BitVector, ByteVector }
 
-package object scodec {
+private[circe] trait BitVectorCodec {
   implicit final val decodeBitVector: Decoder[BitVector] =
     Decoder.instance { c =>
-      Decoder.decodeString(c) match {
-        case Right(str) =>
-          BitVector.fromBase64Descriptive(str) match {
-            case r @ Right(_) => r.asInstanceOf[Decoder.Result[BitVector]]
-            case Left(err) => Left(DecodingFailure(err, c.history))
+      Decoder.decodeJsonObject(c) match {
+        case Right(jsonObject) =>
+          val decoded = for {
+            bitsJ     <- jsonObject("bits")
+            bits      <- bitsJ.asString
+            lengthJ   <- jsonObject("length")
+            length    <- lengthJ.asNumber.flatMap(_.toLong)
+            bitVector <- BitVector.fromBase64Descriptive(bits).right.toOption
+          } yield bitVector.take(length)
+
+          decoded match {
+            case Some(bitVector)  => Right(bitVector)
+            case None             => Left(DecodingFailure("Incorrect format of BitVector field", c.history))
           }
         case l @ Left(_) => l.asInstanceOf[Decoder.Result[BitVector]]
       }
     }
 
+  /**
+    * For serialization of BitVector we use base64. scodec's implementation of `toBase64` adds padding to 8 bits.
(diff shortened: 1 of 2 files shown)


modules/scodec/shared/src/main/scala/io/circe/scodec/package.scala
```diff
@@ -33,3 +85,5 @@ package object scodec {
   implicit final val encodeByteVector: Encoder[ByteVector] =
     Encoder.encodeString.contramap(_.toBase64)
 }
+
+package object scodec extends ByteVectorCodec with BitVectorCodec
```
note: I decided to split `Bit...` and `ByteVector` codecs since `BitVector` one got quite big. I find new structure more readable.

modules/scodec/shared/src/main/scala/io/circe/scodec/package.scala
```diff
@@ -2,22 +2,74 @@ package io.circe
...
+  private def significantBitsInLastByte(bitVector: BitVector): Byte = {
+    if (bitVector.isEmpty) {
+      0
+    } else {
+      val mod = (bitVector.size % 8).toByte
+      if (mod == 0) 8 else mod
+    }
+  }
+
+  private def numberOfCharsToIgnore(input: String): Option[Int] = {
```
note: This function may return `None` in two little bit different cases. Since they're both signalled with the same value it will result in the same error message. I thought it's ok for sake of simplicity but I am open for suggestions.

travisbrown:
This is great, thanks @note! The new tests fail without the change?

note:
@travisbrown checking `roundTrip` law fails without change. It was successful before just because `Arbitrary.arbitrary[Array[Byte]` was used in test suite and therefore only BitVectors of size divisible by 8 were used for tests.

New, additional non-laws tests fails without change but I would say they are not relevant for old base64 representation.

BTW I forgot to mention it explicitly - this change is a breaking one in sense that if someone serialized `BitVector` with old version he will not be able to deserialize it with new version. It's needed because the old wire format was the reason of problem. In theory we can try to distinguish protocol version but it in my judgement it would be a vast overkill in case of circe-scodec. I feel a note in changelog would be sufficient.

travisbrown:
@note I've been thinking more about this, and just to throw out an alternative proposal: what about a representation like `{ "bits": "WA==", "length": 6 }`? It's less arbitrary than having a header representing the length of the last byte, and if someone wants the more concise representation they can use `ByteVector`.

travisbrown:
Just to clarify, my concern is that with the representation in this PR, if someone encodes a `BitVector` and then accidentally tries to decode a `ByteVector` they'll get an incorrect result, as opposed to a decoding failure.

note:
@travisbrown haven't thought about it but I really like your idea. If I don't find any counterexample (and I don't think they're any) I will change my code to your proposition. Also, proposed representation adds only constant number of character (thus string size scales the same as it does now). The only thing - maybe we can use shortened keys, like `{ "b": "WA==", "l": 6 }`, what do you think? (may matter if we have many small BitVectors)?

travisbrown:
@note Thanks for trying it out! I think I prefer more explicit keys, especially since there's `ByteVector` for people who need a more concise representation, but I don't feel very strongly about it.

(Also, if JSON size is actually a big concern for people here, we could also use the tuple-like `[6, "WA=="]` representation.)

note:
@travisbrown updated according to your suggestions

travisbrown:
Thanks again @note!

PR #617, opened 2017-04-27
First attempt to fix #604

### LINKED ISSUE #604: semi-auto can't derive enc/dec for case classes with fields with tagged types
Hey folks,

I have a weird interaction between shapeless tagged types and semi automatic derivation for case classes. I created a minimal example which can be found here: http://scastie.org/29940

The values e1 to e3 are only there to show that the buildling blocks for the automatic derivation seem to be in place and should not affect the derivation.

Maybe I'm just doing something wrong with tagged types or circe?

Greetings :)

travisbrown:
This is a quick sketch of an attempt to fix #604. The following code (adapted from an example by @javierarrieta) previously compiled on 2.10 and 2.11 but failed on 2.12:

```scala
import cats.syntax.contravariant._
import io.circe.Encoder
import shapeless.tag.@@

sealed trait EmailAddressTag
sealed trait PasswordTag
type EmailAddress = String @@ EmailAddressTag
type Password = String @@ PasswordTag

case class CustomerRegistrationEntity(email: EmailAddress, password: Password)

implicit val emailEncoder: Encoder[EmailAddress] = Encoder[String].narrow
implicit val passwordEncoder: Encoder[Password] = Encoder[String].narrow

deriveEncoder[CustomerRegistrationEntity]
```

This was because of a change in the behavior of `RefinedType` in the Scala reflection API in 2.12. This PR makes the code above compile on all three versions.

/cc @Tvaroh

### DIFF
--- a/modules/generic/shared/src/main/scala/io/circe/generic/util/macros/DerivationMacros.scala
+++ b/modules/generic/shared/src/main/scala/io/circe/generic/util/macros/DerivationMacros.scala
@@ -109,7 +109,14 @@ abstract class DerivationMacros[RD[_], RE[_], DD[_], DE[_]] {
     case class Entry(label: String, keyType: Type, valueType: Type)
 
     object Entry {
+      // We need this import for the `RefinedType` constructor in Scala 2.12.
+      import compat._
+
       def unapply(tpe: Type): Option[(String, Type, Type)] = tpe.dealias match {
+        /**
+         * Before Scala 2.12 the `RefinedType` extractor returns the field type
+         * (including any refinements) as the first result in the list.
+         */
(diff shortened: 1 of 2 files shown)


Tvaroh:
@travisbrown thanks, I will check it tomorrow!

Tvaroh:
@travisbrown actually I just check it: it works! This is exciting! When can we expect to get this released?

travisbrown:
@Tvaroh Cool! It still needs tests, but I'm planning to get an 0.8.0-RC1 out tomorrow.

Tvaroh:
@travisbrown awesome! Thanks again for quick reaction and fixing this. It's really helpful to move forward to tagged types for our models.

javierarrieta:
Thanks @travisbrown I tested the code with my small project that reproduced the issue (https://github.com/javierarrieta/circe-refined) and it is working now

PR #618, opened 2017-04-28
[RFC] generic derivation dsl allowing to override individual case class field decoders

tek:
This is a tool intended to simplify decoder creation for case classes where some fields need special treatment, especially extracting values from deep json paths.
It features a DSL that looks like this (see test at https://github.com/tek/circe/blob/flex/modules/tests/shared/src/test/scala/io/circe/generic/extras/flex.scala#L107):
```scala
case class Outer(inner: Option[Inner], auto: List[V], value: V, absent: Option[String], man: V, mapval: V,

implicit val outerDecoder = derive[Outer](
  'inner \\ 'root \ 'in,
  'value.value[Int],
  (('man \\ 'custom).fetch(_.downField("path").downArray.right.right) \ 'target).value[Int],
  ('mapval \\ 'valmap >> parseInt).value[Int],
  'default \\ 'foo >> parseInt,
  'in.in('root, 'nest) \ 'val,
  'sub \ 'a \ 'b,
  'fmapval >>> parseInt,
  'mapFmap >> values[List] >>> parseInt,
  'fmap2 >>> parseInt >>> ((_: Int) + 1)
)
```

Typeclasses `Retrieve` and `Extract` allow chaining of actions per case class field, defaulting to `AutoDerivation` for omitted fields.
The `derive` macro also features an error analysis, as erroneous rules produce implicit errors containing little information about the error causes.

Would you consider this a suitable addition to `generic-extras`? If so, please advise on how to integrate it better with the existing auto derivation. The performance is really bad as it is, so hints on how to speed up compilation would be very helpful.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -279,7 +279,10 @@ lazy val generic = genericBase.jvm
 lazy val genericJS = genericBase.js
 
 lazy val genericExtrasBase = circeCrossModule("generic-extras", mima = previousCirceVersion, CrossType.Pure)
-  .settings(macroSettings(scaladocFor210 = false))
+  .settings(
+    macroSettings(scaladocFor210 = false),
+    libraryDependencies += "org.typelevel" %% "kittens" % "1.0.0-M9"
+  )
   .dependsOn(genericBase)
 
 lazy val genericExtras = genericExtrasBase.jvm
--- a/modules/generic-extras/src/main/scala/io/circe/generic/extras/flex/data.scala
+++ b/modules/generic-extras/src/main/scala/io/circe/generic/extras/flex/data.scala
@@ -1,0 +1,61 @@
+package io.circe
+package generic.extras.flex
+
+case class Chain[A, B](pre: A, post: B)
+
+object Chain
+{
+  implicit def Retrieve_Chain[A, B]
+  (implicit ra: Retrieve[A], rb: Retrieve[B])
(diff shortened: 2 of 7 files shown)


travisbrown:
@tek Wow, thanks! This is very cool, and I've wanted to try something like this out for a long time.

I can definitely imagine this as a new module, and possibly as a new package in generic-extras. My two main concerns are (first) that it's a little more asymmetrical than most of the pieces we have so far (in it supports decoding but not encoding), and (probably more importantly) that the overhead is getting close to simply writing decoders by hand.

I'd be curious to see use cases that aren't currently supported well by the facilities in circe-generic and where this DSL is substantially cleaner than just working with cursors.

In any case I don't think this can make it into the 0.8 release, but I'm definitely interested in seeing about making it happen in 0.9.

tek:
glad you're interested! the motivation for this was that we have a lot of large json structures at work with deep nesting from which we only need to extract few values, so a complete case class representation of the structure would be unfeasible; at the same time there often also are many fields that are flat on the top level, making extraction with cursors very verbose. It's surely possible to split those two cases, but I built this anyway as a proof of concept and quite like it.

As for the asymmetry, I'm not sure what use case would justify such a system for encoding, or how it would be designed.

If you have any suggestions on how to deal with these issues, I'm very open to developing a more suitable implementation.

PR #621, opened 2017-04-29
Optimize binary printing

### LINKED ISSUE #542: Printing int arrays is slower when targeting byte buffers

vkostyukov:
Fixes #542 

Turns out all that machinery around `ByteArrayOutputStream` and `OutputStreamWriter` to power binary printing is pretty inefficient given the number of indirections and locks it involves (yes, both `Writer` and `OutputStream` are thread-safe). I took a stab at replacing them with something as bare-bones as `AppeableByteBuffer` that's both dead simple and fast.

I added a new benchmark to simulate what TechEmpower does and the difference there is pretty dramatic.

Here is the printer benchmark running master:

```
[info] Benchmark                                       Mode  Cnt        Score        Error  Units
[info] PrintingBenchmark.printBooleansToByteBuffer    thrpt   20    32044.032 ±    656.794  ops/s
[info] PrintingBenchmark.printBooleansToString        thrpt   20    36763.796 ±   1017.913  ops/s
[info] PrintingBenchmark.printFoosToByteBuffer        thrpt   20     3380.742 ±    126.873  ops/s
[info] PrintingBenchmark.printFoosToString            thrpt   20     3800.775 ±     72.982  ops/s
[info] PrintingBenchmark.printHelloWorldToByteBuffer  thrpt   20   389569.913 ±  17078.594  ops/s
[info] PrintingBenchmark.printHelloWorldToString      thrpt   20  5914166.900 ± 186088.510  ops/s
[info] PrintingBenchmark.printIntsToByteBuffer        thrpt   20    19291.015 ±    920.108  ops/s
[info] PrintingBenchmark.printIntsToString            thrpt   20    22042.390 ±   1072.936  ops/s
```

Here is on this branch:

```

[info] Benchmark                                       Mode  Cnt        Score        Error  Units
[info] PrintingBenchmark.printBooleansToByteBuffer    thrpt   20    36313.138 ±   2522.476  ops/s
[info] PrintingBenchmark.printBooleansToString        thrpt   20    32649.588 ±   4852.647  ops/s
[info] PrintingBenchmark.printFoosToByteBuffer        thrpt   20     3450.146 ±    164.026  ops/s
[info] PrintingBenchmark.printFoosToString            thrpt   20     3423.746 ±    422.960  ops/s
[info] PrintingBenchmark.printHelloWorldToByteBuffer  thrpt   20  4458381.449 ± 420812.233  ops/s
[info] PrintingBenchmark.printHelloWorldToString      thrpt   20  5745264.182 ± 262477.368  ops/s
[info] PrintingBenchmark.printIntsToByteBuffer        thrpt   20    21768.940 ±   1086.858  ops/s
[info] PrintingBenchmark.printIntsToString            thrpt   20    21359.064 ±    523.034  ops/s
```

As you can clearly see, binary printing is now the same or faster in most of the cases. The important part to understand here, is that, even if binary printing scores the same as string-based printing, it'll still outperform standard printing machinery in the end-to-end HTTP test (with Finch, for example), given that `ByteBuffer` is already ready to be served over the wire while `String` will require one extra copy to get the bytes out.


travisbrown:
Thanks for looking into this, and for the fix (just in time for 0.8.0-RC1)!

modules/core/shared/src/main/scala/io/circe/Printer.scala
```diff
@@ -334,4 +334,47 @@ final object Printer {
       }
     }
   }
+
+  private[circe] final class AppendableByteBuffer(cs: Charset) extends Appendable {
+    private[this] var index = 0
+    private[this] var chars = new Array[Char](32)
+
+    private[this] def ensureToFit(n: Int): Array[Char] = {
```
travisbrown: Is there a reason not to have this return `Unit` and then work directly with `chars`? I don't have a strong opinion here, but I feel like that would be a little more straightforward about where the mutation is happening.

vkostyukov: No particular reason! I've thnk I've seen this pattern somewhere recently and decided to try it out. Let me change that to a `Unit` function so we always mutate `chars`.

vkostyukov: Updated the branch!

travisbrown:
Thanks again! I'm running a little late but am still planning on getting 0.8.0-RC1 out today.

PR #628, opened 2017-05-04
Add encoder and decoder for java.time.YearMonth

britter:
The default implementation will encode/decode `YearMonth` instances to/from strings of the form "2017-04" for example.


travisbrown:
Looks good, thanks!

PR #631, opened 2017-05-11
@JsonCodec encodeOnly/decodeOnly support

carymrobbins:
We can now supply arguments to `@JsonCodec` to signal to only derive an `Encoder` or `Decoder` instance. This is useful in cases where we can derive one but can't derive the other but still would prefer to use the macro annotation to avoid boilerplate.

The macro annotation requires that if you supply an argument, you must explicitly specify `encodeOnly` or `decodeOnly` and they must be set to `true`.  Supporting anything else is a bit pointless.

I've added test cases which demonstrate this behavior.

### DIFF
--- a/modules/generic/shared/src/main/scala/io/circe/generic/JsonCodec.scala
+++ b/modules/generic/shared/src/main/scala/io/circe/generic/JsonCodec.scala
@@ -5,7 +5,10 @@ import macrocompat.bundle
 import scala.language.experimental.macros
 import scala.reflect.macros.blackbox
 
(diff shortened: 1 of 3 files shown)


carymrobbins:
@travisbrown - Ok, my latest push fixes the build.

travisbrown:
Thanks @carymrobbins—sorry for the long delay on review, but I'm planning an early 0.9.0-M1 by the end of the weekend and will get this in.

travisbrown:
Merged in #681.

PR #636, opened 2017-05-17
Add codec for java.time.Duration

sbrunk:
Yet another type to make java.time support more complete.


sbrunk:
Guess I should have run Scalastyle locally...

travisbrown:
Looks good to me, @sbrunk, and thanks for cleaning up the test name.

Issue #589, opened 2017-03-17
Add implicits to enable a more fluent `Json.obj`


tekacs:
Under Play Json, it's possible to [construct a JSON object using Json.obj][play] much as though constructing a heterogenous (Any-typed) map:

```scala
Json.obj(
  "a" -> 1,
  "b" -> "asdf",
  "c" -> Seq(1, 2, 3),
  "d" -> CaseClassWithWriteInstance("1", 2)
)
```

This works because `Json.obj` [accepts `(String, JsValueWrapper)*`][play-jsobj], where [JsValueWrapper is convertible from any type with a Json Writes][play-jsvw].

It was proposed on Gitter and somewhat well received that circe could benefit from this fluent syntax (perhaps with the implicit stashed in another import).

I'm sure that folks here can be less crude about this, but just to show feasibility, this works now (allowing the above syntax):

```scala
import io.circe.syntax._
import io.circe.{Encoder, Json}

case class JsonValue(js: Json) extends AnyVal {
  def underlying: Json = js
}
implicit def jsonValueFromValue[T: Encoder](value: T): JsonValue = JsonValue(value.asJson)
def obj(pairs: (String, JsonValue)*): Json = Json.obj(pairs.map(v => (v._1, v._2.underlying)): _*)
```

[play]: https://www.playframework.com/documentation/2.5.x/ScalaJson#using-class-construction
[play-jsobj]: https://github.com/playframework/play-json/blob/64114eb71c995c9a9c1a087c31830b284e6b1079/play-json/shared/src/main/scala/Json.scala#L152
[play-jsvw]: https://github.com/playframework/play-json/blob/64114eb71c995c9a9c1a087c31830b284e6b1079/play-json/shared/src/main/scala/Json.scala#L209

travisbrown:
Thanks for opening this, @tekacs!

travisbrown:
Resolved in #640.

PR #640, opened 2017-05-22
Add implicits to enable a more fluent `Json.obj` #589

cponomaryov:


travisbrown:
Thanks, @cponomaryov!

PR #642, opened 2017-05-23
Fold instance over fields of a JsonObject

benhutchison:
See #612 for previous attempt at an `Each` instance.

`Fold` is a weaker operation that can get, but not modify, the labelled fields of a `JsonObject`. 

@travisbrown Would like some guidance on what test this PR should include, as`Fold` has no laws that can be verified. What about generating a json object with known fields and assert the fold visits them all?

### DIFF
--- a/modules/optics/src/main/scala/io/circe/optics/JsonObjectOptics.scala
+++ b/modules/optics/src/main/scala/io/circe/optics/JsonObjectOptics.scala
@@ -2,9 +2,9 @@ package io.circe.optics
 
 import cats.instances.list.catsStdInstancesForList
 import io.circe.{ Json, JsonObject }
-import monocle.{ Lens, Traversal }
+import monocle.{ Fold, Lens, Traversal }
 import monocle.function.{ At, Each, FilterIndex, Index }
-import scalaz.{ Applicative, Traverse }
+import scalaz.{ Applicative, Monoid, Traverse }
 import scalaz.std.ListInstances
 
 /**
@@ -20,6 +20,10 @@ trait JsonObjectOptics extends CatsConversions with ListInstances {
         F: Applicative[F]
       ): F[JsonObject] = from.traverse(f)(csApplicative(F))
     }
+  }
+
+  implicit final lazy val objectFoldKV: Fold[JsonObject, (String, Json)] = new Fold[JsonObject, (String, Json)] {
+    def foldMap[M: Monoid](f: ((String, Json)) => M)(obj: JsonObject): M = scalaz.Foldable[List].foldMap(obj.toList)(f)
   }
 
   implicit final lazy val objectAt: At[JsonObject, String, Option[Json]] =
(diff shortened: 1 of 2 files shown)


modules/optics/src/main/scala/io/circe/optics/JsonObjectOptics.scala
```diff
@@ -2,10 +2,11 @@ package io.circe.optics
 
 import cats.instances.list.catsStdInstancesForList
 import io.circe.{ Json, JsonObject }
-import monocle.{ Lens, Traversal }
+import monocle.{ Lens, Traversal, Fold }
```
travisbrown: I think Scalastyle will complain about this.

modules/optics/src/main/scala/io/circe/optics/JsonObjectOptics.scala
```diff
@@ -22,6 +23,10 @@ trait JsonObjectOptics extends CatsConversions with ListInstances {
     }
   }
 
+  implicit final lazy val objectFoldKV: Fold[JsonObject, (String, Json)] = new Fold[JsonObject, (String, Json)] {
+    def foldMap[M: Monoid](f: ((String, Json)) => M)(obj: JsonObject): M = obj.toList.foldMap(f)
```
travisbrown: This doesn't currently compile, and just adding the import won't work, since the cats syntax won't know what to do with the Scalaz monoid. I'd probably just write `scalaz.Foldable[List].foldMap(obj.toList)(f)`.

benhutchison:
Apologies for previous scrappy PR, must have been tired/distracted. Test added now.

travisbrown:
Thanks @benhutchison! Two things: since `Fold` isn't a type class it doesn't need to be implicit, and I think a name like `jsonObjectFields` would be ideal, for consistency with the optic names (vs. type class instance names) in `JsonOptics` and `JsonNumberOptics`. I'm happy to make these changes in a follow-up PR, though.

benhutchison:
Hi @travisbrown, out of curiosity and to understand Circe's design principles: I was wondering about "since Fold isn't a type class" - did you choose not to make `Fold`  a typeclass because it doesn't have laws, or for another reason?

travisbrown:
@benhutchison It's not really circe-specific at all—`Fold` is at the top of the [optics hierarchy](http://julien-truffaut.github.io/Monocle/optics.html) in Monocle (unlike [`Each`, `At`, etc.](http://julien-truffaut.github.io/Monocle/typeclasses.html)), and the types in the optics hierarchy aren't type classes. You generally don't need to (and shouldn't) have any `Lens`, `Fold`, etc. in implicit scope.

You can think of a value of an optics type as a path, where typical use has you building up path values compositionally, and for any two types `A` and `B` there are potentially many paths from `A` to `B`. The type classes like `At` are ways of selecting the next step in a path, where the index and result type are uniquely determined by the type you're stepping from—they're a piece that you use to build up paths, not paths themselves.

Issue #639, opened 2017-05-22
knownDirectSubclasses of <class> observed before subclass <class> registered


joan38:
In:
https://github.com/joan38/orchestra/tree/6d2c170f37040f2179479c6126292e030808ed08

Just rename:
`orchestra/src/main/scala/com/goyeau/orchestra/ARunStatus.scala`
to:
`orchestra/src/main/scala/com/goyeau/orchestra/RunStatus.scala`

And enjoy:
```
[error] knownDirectSubclasses of RunStatus observed before subclass Running registered
[error] knownDirectSubclasses of RunStatus observed before subclass Success registered
[error] two errors found
[error] (orchestraJVM/compile:compileIncremental) Compilation failed
```

Related:
https://github.com/lloydmeta/enumeratum/issues/90
https://issues.scala-lang.org/browse/SI-7046
https://github.com/scala/scala/pull/5284

Using:
Scala 2.12.2
Circe 0.8.0

lloydmeta:
I've been wondering about this one for a while as well and finally got around to successfully reproducing it. I had a bit of trouble reproducing it because it is highly dependent on the order in which things get compiled.

Here's my attempt at a minimal repro: https://github.com/lloydmeta/circe-auto-derive-repro

ngbinh:
@lloydmeta @joan38 putting the extending `classes` and `objects` inside the companion object of the `sealed trait` will cause `knownDirectSubclasses` (sometimes can be intermittently due to incremental compiling).

lloydmeta:
@ngbinh from what I can see, it definitely depends on whether or not that trait/object get compiled before `import io.circe.generic.auto._` gets hit. 

I mentioned this in the readme of my repo, but I'll paste it here as well

> There are 2 ways to make compilation pass:
> 
>   1. Rename `Role` to `ARole`.
>   2. Add `import Role._` at the top of `Main`, before `import io.circe.generic.auto._`

ngbinh:
@lloydmeta yeah, you need to at least try to `derive` `Encoder`/`Decoder` for a `sealed trait` to get `knowDirectSubclass`. I believe `knownDirectSubclass` is used by `shapeless` to compute all the subclasses of a sealed trait. This https://github.com/circe/circe-derivation may help once it is released.

I also think this problem only exists in scala 2.12 as we were fine in 2.11.

lloydmeta:
[...]

PR #644, opened 2017-05-23
Add workarounds for knownDirectSubclasses compile error

lloydmeta:
@travisbrown [mentioned](https://github.com/circe/circe/issues/639#issuecomment-303118282) that he would like to see some detailed docs that specifically dealt with the dreaded `knownDirectSubclasses` issue added so this is my attempt at it.

### DIFF
--- a/docs/src/main/tut/codec.md
+++ b/docs/src/main/tut/codec.md
@@ -327,4 +327,60 @@ slightly better runtime performance in some cases.
    point representation that may lose precision (even when decoding into a type like `BigDecimal`;
    see [this issue][circe-262] for an example).
 
+### `knownDirectSubclasses` error
+
+While using fully automatic derivation, you may have run into an error that looks like this:
+
+```scala
+knownDirectSubclasses of <class> observed before subclass <class> registered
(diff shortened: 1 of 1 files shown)


travisbrown:
This is awesome, thanks @lloydmeta.

ngbinh:
just a comment: we are using `semiauto` and the problem still persists (probably less frequently)

szoio:
As a relatively simple alternative when none of the above works:
Move the sealed trait case class definitions into a different module from the one that contains your generic auto derivation. 
That was the only way I was able to get around this one.

nikitapecasa:
@szoio Hi, thanks for your comment, it solved my problem.

I think this workaround should be added to docs.

@travisbrown What do you think?

travisbrown:
@nikitapecasa Sounds reasonable to me! (if it's not there already?).

szoio:
@nikitapecasa happy it worked for you.

PhilAndrew:
I'm getting this error and I don't really understand what I'm supposed to do.

PR #647, opened 2017-05-30
Add forSumN methods [WIP]

travisbrown:
This is a quick implementation of a set of `forSumN` methods on the model of `forProductN` that are designed to make the macro-less creation of instances for sealed trait hierarchies a little less painful.

For example, suppose you've got this code:

```scala
import io.circe.{ Decoder, Encoder, ObjectEncoder }, io.circe.syntax._

sealed trait Visitor { def id: String }
case class User(id: String, displayName: String) extends Visitor
case class Anon(id: String) extends Visitor

object Visitor {
  implicit val encodeUser: ObjectEncoder[User] =
    Encoder.forProduct2("id", "displayName")((user: User) => (user.id, user.displayName))
  implicit val decodeUser: Decoder[User] =
    Decoder.forProduct2("id", "displayName")(User(_, _))

  implicit val encodeAnon: ObjectEncoder[Anon] =
    ObjectEncoder[Map[String, String]].contramapObject((anon: Anon) => Map("id" -> anon.id))
  implicit val decodeAnon: Decoder[Anon] =
    Decoder[String].prepare(_.downField("id")).map(Anon(_))

  implicit val encodeVisitor: ObjectEncoder[Visitor] = ObjectEncoder.instance {
    case u @ User(_, _) => u.asJsonObject.add("type", "user".asJson)
    case a @ Anon(_) => a.asJsonObject.add("type", "anon".asJson)
  }

  implicit val decodeVisitor: Decoder[Visitor] = for {
    visitorType <- Decoder[String].prepare(_.downField("type"))
    value <- visitorType match {
      case "user" => Decoder[User]
      case "anon" => Decoder[Anon]
      case other => Decoder.failedWithMessage(s"invalid type: $other")
    }
  } yield value
}
```

The additions in this PR allow you to write the following instead:

```scala
import io.circe.{ Decoder, Encoder, ObjectEncoder }

sealed trait Visitor { def id: String }
case class User(id: String, displayName: String) extends Visitor
case class Anon(id: String) extends Visitor

object Visitor {
  implicit val encodeUser: ObjectEncoder[User] =
    Encoder.forProduct2("id", "displayName")((user: User) => (user.id, user.displayName))
  implicit val decodeUser: Decoder[User] =
    Decoder.forProduct2("id", "displayName")(User(_, _))

  implicit val encodeAnon: ObjectEncoder[Anon] =
    ObjectEncoder[Map[String, String]].contramapObject((anon: Anon) => Map("id" -> anon.id))
  implicit val decodeAnon: Decoder[Anon] =
    Decoder[String].prepare(_.downField("id")).map(Anon(_))

  implicit val encodeVisitor: ObjectEncoder[Visitor] =
    Encoder.forSum2(Some("type"))("user", "anon")(
      { case user @ User(_, _) => user },
      { case anon @ Anon(_) => anon }
    )

  implicit val decodeVisitor: Decoder[Visitor] =
    Decoder.forSum2[User, Anon, Visitor](Some("type"))("user", "anon")
}
```

It's only six or seven lines lighter but it's arguably a little cleaner and more consistent. It also makes it easy to switch between the `{ "type": "anon", "id": "foo" }` and `{ "anon": { "id": "foo" }}` formats—you'd simply replace the `Some("type")` with `None` to get the latter in this case.

I'm not sure how I feel about this. If people like it, we should add tests and maybe clean up the `forSumN` implementation a little for consistency.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -274,15 +274,18 @@ trait Decoder[A] extends Serializable { self =>
  * @groupname Product Case class and other product instances
  * @groupprio Product 7
  *
+ * @groupname Sum Sealed trait hierarchy and other sum instances
+ * @groupprio Sum 8
+ *
  * @groupname Time Java date and time instances
- * @groupprio Time 8
+ * @groupprio Time 9
  *
  * @groupname Prioritization Instance prioritization
  * @groupprio Prioritization 10
  *
  * @author Travis Brown
  */
-final object Decoder extends CollectionDecoders with TupleDecoders with ProductDecoders
+final object Decoder extends CollectionDecoders with TupleDecoders with ProductDecoders with SumDecoders
     with JavaTimeDecoders with LowPriorityDecoders {
   /**
    * @group Aliases
--- a/modules/core/shared/src/main/scala/io/circe/Encoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Encoder.scala
@@ -67,15 +67,19 @@ trait Encoder[A] extends Serializable { self =>
  * @groupname Product Case class and other product instances
  * @groupprio Product 7
  *
+ * @groupname Sum Sealed trait hierarchy and other sum instances
+ * @groupprio Sum 8
+ *
  * @groupname Time Java date and time instances
- * @groupprio Time 8
+ * @groupprio Time 9
  *
  * @groupname Prioritization Instance prioritization
  * @groupprio Prioritization 10
  *
  * @author Travis Brown
  */
-final object Encoder extends TupleEncoders with ProductEncoders with JavaTimeEncoders with MidPriorityEncoders {
+final object Encoder extends TupleEncoders with ProductEncoders with SumEncoders
+    with JavaTimeEncoders with MidPriorityEncoders {
   /**
    * Return an instance for a given type `A`.
    *
--- a/project/Boilerplate.scala
+++ b/project/Boilerplate.scala
@@ -32,7 +32,8 @@ object Boilerplate {
 
   val testTemplates: Seq[Template] = Seq(
     GenTupleTests,
-    GenProductTests
+    GenProductTests,
+    GenSumTests
   )
 
   val header = "// auto-generated boilerplate"
@@ -271,7 +272,7 @@ object Boilerplate {
         |  private[this] abstract class SumDecoder[Target] extends Decoder[Target] {
         |    protected[this] def nameMap: Map[String, Decoder[Target]]
         |  }
-        |        
+        |
         |  private[this] abstract class TypeFieldSumDecoder[Target](typeField: String) extends SumDecoder[Target] {
         |    private[this] def failure(name: String, c: HCursor): DecodingFailure =
         |      DecodingFailure(s"Unknown name in $$typeField: $$name", c.history)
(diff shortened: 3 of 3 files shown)


travisbrown:
Downsides: no exhaustivity checking, and if you're using e.g. `Some("myType")` and one of your case classes has a `myType` member, you silently lose round-tripability.

travisbrown:
I rebased this and polished it up and would merge it except that it really blows up the circe-core jar size (e.g. more than double for 2.12).

One other annoying thing is that under certain conditions (specifically no non-implicit scope implicits in scope), `Decoder.forSumN` can compile with inferred type parameters but then NPE (since the inferred arguments are `Base, Base, Base, ...` and the instances for the constructors are the current instance). I don't think this would be enough to sink the whole thing on its own, but with the lack of exhaustivity, etc., I just don't think it's worth the extra weight.

asakaev:
I am working on the BSON library right now. It's heavily inspired by circe and reflect circe API. My colleague asks for a simple way of constructing ADTs, so I come with this draft API.

```
sealed trait AorB
object AorB {
  final case class A(i: Int)    extends AorB
  final case class B(s: String) extends AorB
}

implicit val eqAorB: Eq[AorB] = Eq.fromUniversalEquals

implicit val encodeA: Encoder[A] = Encoder.forProduct1("i")(_.i)
implicit val encodeB: Encoder[B] = Encoder.forProduct1("s")(_.s)
implicit val encodeAorB: Encoder[AorB] = Encoder.forCoproduct {
  case a @ A(_) => "a" -> a.asBson
  case b @ B(_) => "b" -> b.asBson
}

implicit val decodeA: Decoder[A]       = Decoder.forProduct1("i")(A.apply)
implicit val decodeB: Decoder[B]       = Decoder.forProduct1("s")(B.apply)
implicit val decodeAorB: Decoder[AorB] = Decoder.forCoproduct2[AorB, A, B]("a", "b")

val aBson: BsonValue = (A(42): AorB).asBson      // {"tag": "a", "value": {"i": 42}}
val bBson: BsonValue = (B("hello"): AorB).asBson // {"tag": "b", "value": {"s": "hello"}}

val a: Either[String, AorB] = aBson.as[AorB] // Right(A(42))
val b: Either[String, AorB] = bBson.as[AorB] // Right(B(hello))
```

Method signatures are like that:
```
def forCoproduct[A](f: A => (String, BsonValue)): Encoder[A]
def forCoproduct2[A, B <: A: Decoder, C <: A: Decoder](tb: String, tc: String): Decoder[A]
```

Scala compiler with enabled -Xfatal-warnings flag force to handle all coproduct choices. What do you think?

travisbrown:
@asakaev My main concern about that approach is that it encodes a representation of ADTs that isn't round-trippable in the general case and doesn't match the one used by default in circe-generic, circe-derivation, etc. It's a reasonable representation, and if we had some kind of circe-core-extras module I'd be tempted to include, but I don't think it belongs in circe-core, where we do our best to provide encodings that always round-trip and that don't make arbitrary decisions for the user (like `tag` vs. `type`, etc.).

asakaev:
Thanks for the detailed explanation. Can you give me an example of non-round-trappable data with this representation?

travisbrown:
@asakaev Oh, sorry—I missed that there's both a `tag` and a `value` (not just a discriminator, like e.g. `{"tag": "a", "i": 42}`), so you're right, it will round-trip. The "arbitrary decision" part still stands (see [this change](https://github.com/circe/circe/pull/1013) for a similar case).

I guess we could just create a new circe-extras module here now, for this, and the duration instances, and maybe some stuff that's currently in core that doesn't really belong there (like `disjunctionCodecs` and `Decoder.state`).

hamnis:
Closed due to age.

PR #648, opened 2017-05-30
Allow transforming the value of the discriminator field.

nathankleyn:
Adds a `transformDiscriminator` option to `Configuration` which allows
transformation of the value used for the discriminator (whether inside
the object or outside). This works the same way as `transformKeys` does.

As an example, given the following classes to encode/decode:

```scala
sealed trait State
case object RedLight extends State
case object YellowLight extends State
case object GreenLight extends State

case class TrafficLight(state: State)
```

One can use the following configuration to get a customised JSON output
even for the `type` field:

```scala
implicit val config = Configuration.default.withDiscriminator("type").copy(transformDiscriminators = _.toLowerCase)
```

```json
{
  "state" : {
    "type" : "redlight"
  }
}
```

Normally, one would get the following JSON:

```json
{
  "state" : {
    "type" : "RedLight"
  }
}
```

It's worth noting as well that, as with `transformKeys`, the
`transformDiscriminator` has a convenience transformation to snake case
(which is undoubtedly one of the most common use cases for this):

```
implicit val config = Configuration.default.withDiscriminator("type").withSnakeCaseDiscriminators
```

### DIFF
--- a/modules/generic-extras/src/main/scala/io/circe/generic/extras/Configuration.scala
+++ b/modules/generic-extras/src/main/scala/io/circe/generic/extras/Configuration.scala
@@ -1,8 +1,27 @@
 package io.circe.generic.extras
 
-final case class Configuration(transformKeys: String => String, useDefaults: Boolean, discriminator: Option[String]) {
+/**
+  * Configuration allowing customisation of the JSON produced when encoding, or expected when decoding. Can be used
+  * with the [[ConfiguredJsonCodec]] annotation to allow customisation of the semi-automatic derivation.
+  *
+  * @param transformKeys Transforms the names of the keys in the JSON allowing, for example, formatting or case changes.
+  * @param useDefaults Whether to allow default values as specified in any case-classes/ADT fields.
+  * @param discriminator Optional key name that, when given, will be used to store the name of the type of an ADT in a
+  *                      nested field with this name. If not given, the type is instead stored as a key under which the
+  *                      contents of the ADT are stored as an object.
+  * @param transformDiscriminators Transforms the value of the discriminator in the JSON allowing, for example,
+  *                                formatting or case changes.
+  */
+final case class Configuration(transformKeys: String => String,
+                               useDefaults: Boolean,
+                               discriminator: Option[String],
+                               transformDiscriminators: String => String) {
   def withSnakeCaseKeys: Configuration = copy(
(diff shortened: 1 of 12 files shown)


nathankleyn:
I have some open questions about the stuff I've written in here!:

- [ ] I'm not sold on the `transformDiscriminators` naming, which only really makes sense if `discriminator` is possibly renamed (as that naming implies you only get a thing called a "discriminator" if that optional value is set).
- [ ] I have added documentation to `Configuration` as it was notably absent from the Scaladoc pages (and the naming is potentially a little inaccessible without it). Is the documentation correct?
- [ ] What is the policy towards adding new fields to `Configuration, and is this considered a breaking change? I wonder whether the constructor for `Configuration` should be public at all and instead the builder pattern encouraged through the use of `Configuration.default` and friends is a better way to avoid this being a breaking change for anybody using the constructor.

modules/generic-extras/src/main/scala/io/circe/generic/extras/Configuration.scala
```diff
@@ -1,16 +1,35 @@
 package io.circe.generic.extras
 
-final case class Configuration(transformKeys: String => String, useDefaults: Boolean, discriminator: Option[String]) {
+/**
+  * Configuration allowing customisation of the JSON produced when encoding, or expected when decoding. Can be used
+  * with the [[ConfiguredJsonCodec]] annotation to allow customisation of the semi-automatic derivation.
+  *
+  * @param transformKeys Transforms the names of the keys in the JSON allowing, for example, formatting or case changes.
+  * @param useDefaults Whether to allow default values as specified in any case-classes/ADT fields.
```
travisbrown: This is really only applicable for case classes.

nathankleyn: Good spot! Fixed!

nathankleyn:
@travisbrown Thanks for the quick response! I've pushed a new commit with the changes you suggested, I really like the new names (feels much more natural)!

One quick question: I've left `withSnakeCaseKeys` and marked it as deprecated as of `0.9.0`, but not sure if you meant explicitly deprecate or just to remove it and replace it with the new name and show it in the release notes. Happy to change if you meant the later!

travisbrown:
Commits merged in #684, thanks @nathankleyn!

Issue #643, opened 2017-05-23
Error messages lost for recusive decoders


EdgeCaseBerg:
When using `emap` to create a custom validation for a recusive decoder the error messages are lost for the parts of the object which are decoded recursively:

```
import io.circe._, io.circe.generic.semiauto._, io.circe.generic.decoding.DerivedDecoder, io.circe.generic.encoding.DerivedObjectEncoder

case class A(i: Int, l: List[A])

def validA(a: A): Boolean = {
    a.i > 0 && a.l.map(validA).foldLeft(true)(_ && _)
}

implicit val ae: Decoder[A] = deriveDecoder[A].emap { a =>
    if(!validA(a)) {
        Left("Bad!")
    } else {
        Right(a)
    }
}

val s = """{"i":1, "l": [{"i": -1, "l": []}, {"i": 2, "l": []}]}"""

parser.parse(s).right.get.as[A]
```

When running the last line of code you get 
```
io.circe.Decoder.Result[A] = Left(DecodingFailure([A]List[A], List(DownArray, DownField(l))))
```
Which is a valid decoding failure, but the error message of `[A]List[A]` isn't what is expected. I would expect `Bad!` or something similar to  `DecodingFailure at .l[0]: Bad!`

I'm using circe 0.7.0 for reference. And this issue was originally [raised in the gitter channel](https://gitter.im/circe/circe?at=5924556500efc2bb3e98405d)

EdgeCaseBerg:
I came up with another example, not sure if it will help. But this one doesn't use `emap` and instead tries to be explicit about which decoder to use during the decoding of the children, but it still results in the same issue:

```
import io.circe._, io.circe.generic.semiauto._, io.circe.generic.decoding.DerivedDecoder, io.circe.generic.encoding.DerivedObjectEncoder

case class A(id: Int, children: List[A])

import cats.syntax.either._

implicit val decoder: Decoder[A] = Decoder.instance { hCursor =>
	val idDecodingResult = hCursor.downField("id").as[Int] match {
		case Right(id) if id > 0 => Right[DecodingFailure, Int](id)
		case Right(id) => Left(DecodingFailure("Id must be greater than 0", hCursor.history))
		case l => l
	}
	val childrenResult = hCursor.downField("children").as[List[A]](Decoder.decodeList(decoder))
	for {
		id <- idDecodingResult
		children <- childrenResult
	} yield A(id, children)
}

[...]

PR #649, opened 2017-05-31
decodeList discards error messages from list items #596; Error messages lost for recusive decoders #643

cponomaryov:

PR #651, opened 2017-06-01
Do not override error msg's of CanBuildFrom #643

EdgeCaseBerg:
Addresses issue #643 which contained an example of a custom error
message being lost for a List type when an item in that list failed to
decode properly.

The problem is present for three basic types of collections

	- Set
	- List
	- Vector

since each of these decoders defined in Decoder.scala call
withErrorMessage and therefore lose any message bubbled up from within
that collection.


cponomaryov:
I've already created a pull request for this with tests:
https://github.com/circe/circe/pull/649

travisbrown:
Sorry about the duplication, @EdgeCaseBerg—I guess we should probably have a waffle.io board or something like that to handle coordination.

EdgeCaseBerg:
No worries, glad to see it got done!

Issue #654, opened 2017-06-05
Bring circe-numbers back into circe-core


travisbrown:
The original motivation for having circe-numbers as a separate module was that it could be a general purpose zero-dependency thing on a slower release cycle (I even thought that the JVM version might even be implemented in Java, so that it wouldn't have to be cross-published for different Scala versions—I ported it at one point but never came up with a satisfying way to handle the Scala.js side).

I'm not sure publishing a separate artifact for one type is really worth the trouble and confusion, though. As far as I know it's not the case that any projects are using it but not circe, and it's still changing more often than I'd like.

So I'm thinking of bringing the contents of `io.circe.numbers` back into circe-core. Specifically I'm thinking of combining some stuff:

1. `io.circe.JsonNumber` will go away entirely.
2. `io.circe.numbers.BiggerDecimal` will get a lazy implementation and will become `io.circe.JsonNumber`.
3. `Json` will get new `JDouble` and `JLong` constructors (still private like all the others, of course).

This is a little simpler and a little flatter, so that a JSON AST value representing a `Long`-sized integer or `Double`-sized floating-point number will only be one case class deep, not two.

I've got [a branch](https://github.com/circe/circe/tree/topic/flatten-numbers) sketching out this idea and it's getting ~45% more throughput in the integer array decoding benchmark, which is one of the places circe has still been lagging behind spray-json. The primary motivation is to cut down the number of types, though.

This shouldn't affect most users at all. If you're one of the few people using `BiggerDecimal` directly, you'll have to break out `sed`, and if you're doing unusual stuff with `JsonNumber` you might have to rework it a bit, but that's it. (If you're depending on circe-numbers but not circe-core then you're just out of luck. Sorry about that, but I didn't know you existed.)

PR #656, opened 2017-06-07
Add Json.Folder

travisbrown:
This change introduces a new way to fold `Json` values that's a little more verbose than `Json#fold` but is faster and involves fewer allocations. It also seems to perform better than pattern matching on the (package-private) `Json` constructors in the benchmark included here:

```
FoldingBenchmark.withFold          thrpt   40   6073.949 ±  47.519  ops/s
FoldingBenchmark.withFoldWith      thrpt   40  11401.499 ±  89.627  ops/s
FoldingBenchmark.withPatternMatch  thrpt   40   7537.532 ± 219.453  ops/s
```

The allocation rates also look pretty good:

```
FoldingBenchmark.withFold:gc.alloc.rate.norm          thrpt  5  150744.081 ±  0.003  B/op
FoldingBenchmark.withFoldWith:gc.alloc.rate.norm      thrpt  5  117752.045 ±  0.008  B/op
FoldingBenchmark.withPatternMatch:gc.alloc.rate.norm  thrpt  5  132024.068 ±  0.005  B/op
```

The benchmark is intended to test idiomatic use of the three styles.

The primary motivation for this change is #654—I'm adding a couple new `Json` constructors and would prefer to avoid using pattern matching even internally, but I don't want to pay the extra cost of `fold`. I believe it's a reasonable addition even apart from #654, though.

I've also added `asNull` and `withNull` methods for the sake of consistency, since I've found myself wanting these a couple of times recently.

After this addition the recommendation would continue to be something like "start with `fold` or the `asX` or `withX` methods" but with a new "unless you know the performance cost is unacceptable, in which case there's `foldWith`".

### DIFF
--- a/modules/benchmark/src/main/scala/io/circe/benchmark/FoldingBenchmark.scala
+++ b/modules/benchmark/src/main/scala/io/circe/benchmark/FoldingBenchmark.scala
@@ -1,0 +1,69 @@
+package io.circe.benchmark
+
+import io.circe.{ Json, JsonNumber, JsonObject }
+import java.util.concurrent.TimeUnit
+import org.openjdk.jmh.annotations._
+
+/**
+ * Compare the performance of various ways of folding JSON values.
+ *
+ * The following command will run the benchmarks with reasonable settings:
+ *
+ * > sbt "benchmark/jmh:run -i 10 -wi 10 -f 2 -t 1 io.circe.benchmark.FoldingBenchmark"
+ */
+@State(Scope.Thread)
+@BenchmarkMode(Array(Mode.Throughput))
+@OutputTimeUnit(TimeUnit.SECONDS)
+class FoldingBenchmark extends ExampleData {
+  val doc: Json = Json.arr(intsJson, booleansJson, foosJson, helloWorldJson)
+
+  @Benchmark
+  def withFoldWith: Int = doc.foldWith(
+    new Json.Folder[Int] {
+      private[this] val accumulate: (Int, Json) => Int = _ + _.foldWith(this)
+
+      def onNull: Int = 0
+      def onBoolean(value: Boolean): Int = if (value) 1 else 0
+      def onNumber(value: JsonNumber): Int = value.truncateToInt
+      def onString(value: String): Int = value.length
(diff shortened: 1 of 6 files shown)


modules/benchmark/src/main/scala/io/circe/benchmark/FoldingBenchmark.scala
```diff
@@ -0,0 +1,69 @@
...
+ */
+@State(Scope.Thread)
+@BenchmarkMode(Array(Mode.Throughput))
+@OutputTimeUnit(TimeUnit.SECONDS)
+class FoldingBenchmark extends ExampleData {
+  val doc: Json = Json.arr(intsJson, booleansJson, foosJson, helloWorldJson)
+
+  @Benchmark
+  def withFoldWith: Int = doc.foldWith(
+    new Json.Folder[Int] {
```
vkostyukov: One cool trick that I learned from Flavio Brasi is embedding this `val`ed functions into their parent scope/class. Somewhat - merging interfaces.

```scala
new Json.Folder[Int] with ((Int, Json) => Int) {
   def apply(i: Int, j: Json): Int = i + j.foldWith(this)
   ...
}
```

This will allocate less.

travisbrown: :+1:, yeah, this is neat. It only has a tiny effect on allocations in this benchmark, but I think it looks a little nicer anyway, so let's consider it idiomatic.

travisbrown:
@vkostyukov Yep—the change is pretty small with this PR, but I've got a follow-up coming with some bigger improvements (In particular I'm thinking of adding an optional setting for printers that lets them reuse string builders via thread locals, which has a pretty big impact).

vkostyukov:
Nice @travisbrown! That sounds pretty cool - I'd probably be down to have that setting on by default in finch-circe.

travisbrown:
@vkostyukov Yeah, I think for most users it would make sense, but I'm wary of cases where it could cause problems in application servers, so it'll probably be behind a `printer.reuseStringBuilders` option, at least for now.

PR #657, opened 2017-06-08
Printing optimizations

travisbrown:
This change does a couple of things to make printing faster. The first is that it adds package-private methods for writing values to string builders, which lets us (safely) avoid allocating a lot of intermediate strings.

It also adds a new `reuseWriters` option to `Printer`. This option is off by default but allows users to reuse string builders via Java's `ThreadLocal` if they know that they're not on an application server or whatever where thread locals can be a problem (is that still a thing? I have no idea whether that's still a thing).

The results are pretty good—well over twice the throughput vs. 0.8.0 for our integer array printing benchmark, and the allocation rate for that benchmark is down 89%. The case class benchmark doesn't see quite as big of an improvement but it's still not bad.

Spray JSON still has a slight edge in the case class printing benchmark, but this puts us ahead in the integer array printing one (and well ahead of all of the other Scala JSON libraries in both).

Here's throughput for 0.8.0:

```
Benchmark                          Mode  Cnt      Score     Error  Units
PrintingBenchmark.printFoosCirce  thrpt   20   5038.843 ±  92.059  ops/s
PrintingBenchmark.printIntsCirce  thrpt   20  23628.537 ± 260.963  ops/s
```

And here's this PR with `reuseWriters` set to false (the default):

```
Benchmark                          Mode  Cnt      Score     Error  Units
PrintingBenchmark.printFoosCirce  thrpt   20   5317.051 ±  15.540  ops/s
PrintingBenchmark.printIntsCirce  thrpt   20  43372.400 ± 823.195  ops/s
```

And here's this PR with `reuseWriters` enabled:

```
Benchmark                          Mode  Cnt      Score      Error  Units
PrintingBenchmark.printFoosCirce  thrpt   20   5968.110 ±   16.130  ops/s
PrintingBenchmark.printIntsCirce  thrpt   20  50645.811 ± 3065.231  ops/s
```

And allocation rates for 0.8.0:

```
Benchmark                                             Mode  Cnt       Score    Error Units
PrintingBenchmark.printFoosCirce:gc.alloc.rate.norm  thrpt    5  385192.098 ±  0.013  B/op
PrintingBenchmark.printIntsCirce:gc.alloc.rate.norm  thrpt    5   74504.021 ±  0.003  B/op
```

Without `reuseWriters`:

```
Benchmark                                             Mode  Cnt       Score    Error Units
PrintingBenchmark.printFoosCirce:gc.alloc.rate.norm  thrpt    5  367336.091 ±  0.025  B/op
PrintingBenchmark.printIntsCirce:gc.alloc.rate.norm  thrpt    5   26488.010 ±  0.001  B/op
```

And with it:

```
Benchmark                                             Mode  Cnt       Score    Error Units
PrintingBenchmark.printFoosCirce:gc.alloc.rate.norm  thrpt    5  109168.082 ±  0.022  B/op
PrintingBenchmark.printIntsCirce:gc.alloc.rate.norm  thrpt    5    7928.009 ±  0.001  B/op
```

Unfortunately we can't straightforwardly give the `reuseWriters` treatment to `prettyByteBuffer`, since we're handing the user back a `ByteBuffer` that wraps its internal array. Also I really wish Java's `Appendable` had all of the `append` overloads for doubles and stuff that `StringBuilder` has, so that we could use them here for both. I guess we could add our own to `AppendableByteBuffer`, but I'm not sure I care enough.

/cc @vkostyukov

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Printer.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Printer.scala
@@ -61,12 +61,22 @@ final case class Printer(
   private[this] final val commaText = ","
   private[this] final val colonText = ":"
 
-  private[this] final class PrintingFolder(writer: Appendable) extends Json.Folder[Unit] {
+  private[this] final class StringBuilderFolder(writer: StringBuilder) extends PrintingFolder(writer) {
+    final def onBoolean(value: Boolean): Unit = writer.append(value)
+    final def onNumber(value: JsonNumber): Unit = value.appendToStringBuilder(writer)
+  }
+
+  private[this] final class AppendableByteBufferFolder(
+    writer: Printer.AppendableByteBuffer
+  ) extends PrintingFolder(writer) {
+    final def onBoolean(value: Boolean): Unit = writer.append(java.lang.Boolean.toString(value))
+    final def onNumber(value: JsonNumber): Unit = writer.append(value.toString)
+  }
+
+  private[this] abstract class PrintingFolder(writer: Appendable) extends Json.Folder[Unit] {
     private[this] var depth: Int = 0
 
     final def onNull: Unit = writer.append("null")
-    final def onBoolean(value: Boolean): Unit = writer.append(java.lang.Boolean.toString(value))
-    final def onNumber(value: JsonNumber): Unit = writer.append(value.toString)
 
     final def onString(value: String): Unit = {
       writer.append('"')
@@ -260,7 +270,7 @@ final case class Printer(
       w
     } else new StringBuilder()
 
-    val folder = new PrintingFolder(writer)
+    val folder = new StringBuilderFolder(writer)
 
     json.foldWith(folder)
 
@@ -269,7 +279,7 @@ final case class Printer(
 
   final def prettyByteBuffer(json: Json, cs: Charset): ByteBuffer = {
     val writer = new Printer.AppendableByteBuffer(cs)
-    val folder = new PrintingFolder(writer)
+    val folder = new AppendableByteBufferFolder(writer)
 
     json.foldWith(folder)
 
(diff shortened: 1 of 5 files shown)


travisbrown-stripe:
Hmm, Travis CI doesn't like my new tests (they run fine locally). Will take a look later.

modules/core/shared/src/main/scala/io/circe/Printer.scala
```diff
@@ -242,12 +255,22 @@ final case class Printer(
...
+    override final def initialValue: StringBuilder = new StringBuilder()
+  }
+
   /**
    * Returns a string representation of a pretty-printed JSON value.
    */
   final def pretty(json: Json): String = {
-    val writer = new StringBuilder()
-    val folder = new PrintingFolder(writer)
+    val writer = if (reuseWriters) {
```
vkostyukov: This is nice! I can open a follow-up PR to add the same machinery for `AppendableByteBuffer`.

It's going to me a bit more complicated given it also involves a `Charset`, but I think we can start with caching only UTF-8 writers (like 99% of the use cases).

travisbrown: I'm not sure how that would work, though—we can't reuse the array because we've returned a byte buffer wrapping it, and the user still could care about it. We could copy into the buffer but that seems to defeat the purpose to some degree…

vkostyukov: Ah right. I have a strong feeling that as of this change string-base printing will probably be a best choice for the TE benchmark given we send the same message over and over again. The fact we can cache a `StringBuilder` means we don't need to resize/grow it for the next message (given it's the same message). This should translate into a lot of RPS.

I'm thinking of something like message-size-guestimation machinery would do the same trick for binary-printing.

Issue #469, opened 2016-11-24
Generic instances for value classes


pjan:
It would be great if these would make it to the library. Happy to open a PR for it, if somebody points me to the right place for these:

```scala
  implicit def decoderValueClass[T <: AnyVal, V](
    implicit
    gen:     Lazy[Generic.Aux[T, V :: HNil]],
    decoder: Decoder[V]
  ): Decoder[T] = Decoder.instance { cursor ⇒
    decoder(cursor).map { value ⇒
      gen.value.from(value :: HNil)
    }
  }

  implicit def encoderValueClass[T <: AnyVal, V](
    implicit
    gen:     Lazy[Generic.Aux[T, V :: HNil]],
    encoder: Encoder[V]
  ): Encoder[T] = Encoder.instance { value ⇒
    encoder(gen.value.to(value).head)
  }
```

travisbrown:
I'd say let's go either with this (or something using Shapeless's `Unwrapped`) in circe-generic-extras for now.

I'd support making it the default there for fully automatic derivation, and if it works well it could graduate to circe-generic in an upcoming version.

cakper:
👍 for adding it to circe-generic :)

kciesielski:
@pjan I tried to use this but I'm getting compilation error:
```
package.scala:42: wrong number of type arguments for ::, should be 1
[error]       implicit gen: Lazy[Generic.Aux[T, V :: HNil]],
```
in the `implicit gen: Lazy[Generic.Aux[T, V :: HNil]],` line. Circe 0.7.0, shapeless 2.3.2, Scala 2.12.1. I'm not very good with Shapeless, can you give me a hint what can I fix to get this thing working? I managed to find a compilable encoder:
```scala
  implicit def encodeAnyVal[A <: AnyVal, L <: HList, T](
      implicit gen: Generic.Aux[A, L],
      hcons: IsHCons.Aux[L, T, HNil],
      anyval: Encoder[T]
  ): Encoder[A] =
    anyval.contramap(a => hcons.head(gen.to(a)))
```
and I need a decoder.

radusw:
@kciesielski try this:

```scala
  implicit def decodeAnyVal[T, U](
    implicit
    ev: T <:< AnyVal,
    unwrapped: Unwrapped.Aux[T, U],
    decoder: Decoder[U]): Decoder[T] = Decoder.instance[T] { cursor =>

    decoder(cursor).map(value => unwrapped.wrap(value))
  }

[...]

PR #661, opened 2017-06-15
adds generic encoder/decoder for valueclasses.

pjan:
Attempt to address #469.

I'm not really sure how to add it to the `auto` package.
At the same time, in `semiauto`, the `deriveValueClassEncoder` works, but there's implicit resolution issues with the `deriveValueClassDecoder` that I seem unable to wrap my head around at this point.

Happy to make changes/amendments base upon input, @travisbrown.


modules/generic-extras/src/main/scala/io/circe/generic/extras/decoding/ValueClassDecoder.scala
```diff
@@ -0,0 +1,19 @@
...
+import shapeless.{::, Generic, HNil, Lazy}
+
+abstract class ValueClassDecoder[A] extends Decoder[A]
+
+final object ValueClassDecoder {
+  implicit def decodeValueClass[A <: AnyVal, R](
+    implicit
+    gen:    Lazy[Generic.Aux[A, R :: HNil]],
+    decode: Decoder[R]
+  ): Decoder[A] = new ValueClassDecoder[A] {
```
Wogan: Shouldn't this return type be ValueClassDecoder[A] not just Decoder[A] ?

pjan:
@travisbrown I fixed the PR, and added tests (currently for a ValueClass wrapping a String, but that should be sufficient). This is currently only for semiauto derivation.

If you want it for fully automatic derivation, please provide some pointers on how you envision it, and I can try to find some time.

fwiw, we keep repeating code for this in all our projects, so it would be great if this could make it in a release sooner rather than later. That's why I'm happy to put in some extra effort to get it done.

travisbrown:
Thanks for the tests! Looks good to me, although I'm kind of tempted to change the names to `...Unwrapped...`, both for consistency with Shapeless and since it's perfectly reasonable to want the wrapped representation for value classes. What do you think?

I'll merge now and we can make a final decision about the names in the next few days before the next release.

PR #666, opened 2017-06-17
Clean up laws and use better Shrink instances

travisbrown:
`ParserTests` was using the default `Shrink` for `Json`, which made debugging test failures a huge headache. This PR lets the `ParserTests` user provide both the `Arbitrary` and `Shrink` instances (and does the same for `CodecTests` and `PrinterTests` for consistency).


fommil:
:smiling_imp:

Issue #664, opened 2017-06-16
Encoder/Decoder fail java serialization due to scala IsTraversableOnce


Astrac:
When sending an `Encoder` instance over the wire using the Java serialisation it is possible to encounter the following error:

```
java.io.NotSerializableException: scala.collection.generic.IsTraversableOnce$$anon$2
```

Adding `"-Dsun.io.serialization.extendedDebugInfo=true"` to the build reveals a bit more information about that:

```
field (class "io.circe.MidPriorityEncoders$$anon$9", name: "is$2", type: "interface scala.collection.generic.IsTraversableOnce")
object (class "io.circe.MidPriorityEncoders$$anon$9", io.circe.MidPriorityEncoders$$anon$9@553d98f4)
field (class "io.chumps.model.json.PostFormats$$anon$10", name: "fresh$macro$341", type: "interface io.circe.ArrayEncoder")
```

It looks like some closuere inside `IsTraversableOnce` is not java serialisable; in my case this is happening here:

https://github.com/circe/circe/blob/master/modules/core/shared/src/main/scala/io/circe/Encoder.scala#L344

But there is also another instance of this in the same file for the `cats.data.OneAnd` encoder:

https://github.com/circe/circe/blob/master/modules/core/shared/src/main/scala/io/circe/Encoder.scala#L272

In my case, providing this implicit to override the one built in circe seems to work fine:

```
  implicit def forIterableEncoder[T, I[_]](
    implicit e: Encoder[T],
    ev: I[T] <:< Iterable[T]
  ): ArrayEncoder[I[T]] = new ArrayEncoder[I[T]] {
    final def encodeArray(in: I[T]): Vector[Json] = {
      val items = Vector.newBuilder[Json]
      val it = ev(in).toIterator

      while (it.hasNext) {
        items += e(it.next())
      }

      items.result()
    }
  }
```

Another instance seem to be with the `Decoder` for `Seq`s:

```
[info]   Cause: java.io.NotSerializableException: scala.collection.generic.GenTraversableFactory$$anon$1
[info]  - field (class "io.circe.SeqDecoder", name: "cbf", type: "interface scala.collection.generic.CanBuildFrom")
[info]  - object (class "io.circe.SeqDecoder", io.circe.SeqDecoder@18ca78b6)
[info]  - field (class "io.circe.Decoder$$anon$13", name: "$outer", type: "interface io.circe.Decoder")
[...]

PR #671, opened 2017-06-19
Fix Serializable issues (partially)

travisbrown:
This is a pretty big change that addresses #664 and does a couple of other things.

The problem is that Scala's `CanBuildFrom` and `IsTraversableOnce` type classes by default don't extend `Serializable`, which means that `Decoder` and `Encoder` instances that capture them aren't serializable (which causes problems on frameworks like Spark).

This change makes it so that instances for the following types are `Serializable`:

* `scala.collection.Seq`
* `scala.collection.immutable.List`
* `scala.collection.immutable.Vector`
* `scala.collection.immutable.Set`
* `scala.collection.immutable.Map` 
* `cats.data.NonEmptyList`
* `cats.data.NonEmpyVector`

Instances for other collection types will not be serializable unless you provide a serializable `CanBuildFrom` or `C[A] => Iterable[A]` (see [the tests](https://github.com/circe/circe/blob/bb728f12e627079cfef97db3025d029ca2057a81/modules/tests/shared/src/test/scala/io/circe/CodecSuites.scala#L48) for an example of how to do this).

I've made a few other changes for the sake of cleanliness and consistency while I'm here. Some type parameters are rearranged, and we now provide instances for mutable maps (to match the instances we've been providing for mutable sequences). I've also changed `Encoder` to use evidence of a subtype of / conversion to `Iterable` rather than the `IsTraversableOnce` type class.

I've also added better tests for `Serializable` that turned up a few other problems. The `java.time` instances in circe-java8 are not serializable for the types that require a `java.time.format.DateTimeFormatter` (which isn't `Serializable`), and `forProductN`-created instances aren't serializable unless the function you provide is.

The changes here shouldn't be user-facing for the most part.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/ACursor.scala
+++ b/modules/core/shared/src/main/scala/io/circe/ACursor.scala
@@ -2,6 +2,7 @@ package io.circe
 
 import cats.Applicative
 import cats.kernel.Eq
+import java.io.Serializable
 import scala.collection.immutable.Set
 
 /**
--- a/modules/core/shared/src/main/scala/io/circe/AccumulatingDecoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/AccumulatingDecoder.scala
@@ -2,6 +2,7 @@ package io.circe
 
 import cats.{ ApplicativeError, Semigroup, SemigroupK }
 import cats.data.{ NonEmptyList, Validated, ValidatedNel }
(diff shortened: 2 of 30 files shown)


travisbrown:
For what it's worth these changes also seem to improve performance a bit (with the case class-writing benchmark seeing the largest change—around 8% more throughput).

Issue #630, opened 2017-05-10
Make Encoder.forProductN methods actually use… ProductN


travisbrown:
Currently all the `Encoder.forProductN` methods take a function from the type that we're defining an `Encoder` for to a tuple with the appropriate arity. I don't see any reason not to allow the more general `Source => ProductN[...]`, which would make definitions simpler for case classes (no more `(User.unapply _).andThen(_.get)` stuff) and would also help with e.g. Scrooge-generated types.

lvicentesanchez:
I could help with this one, I would just need a bit more of context.

travisbrown:
Thanks @lvicentesanchez! Essentially I think (hope?) it'll just be changing tuples to `Product` in the code gen and making sure nothing breaks.

lvicentesanchez:
@travisbrown that should be easy, I re-read the ticket after my commit and I realise that all the information needed was already in the ticket. I will give it a go when I get home.

lvicentesanchez:
@travisbrown I tried yesterday and found a problem. If I replace the tuples of `ProductN` of the right arity, examples like the `WrappedStringOptional` don't compile, you need to do something like `Encoder.forProduct1(o => Tuple1(o))`. I can avoid that by doing something like this,

```
val outputType = if (arity != 1) s"Product$arity[${`A..N`}]" else `A..N`
```

and the forProductN definition would be rewritten this way:

```
final def forProduct$arity[${`A..N`}, Source]($memberNames)(f: Source => $outputType)
```

So... no `Product1`. That makes possible to define encoders for case classes like this:

```
Encoder.forProductN(CaseClass3.unapply)
```
If you are happy with this solution I can submit a PR.

lvicentesanchez:
And just after sending the message... everything compiles, but it seems that there is a problem with some JS tests.

lvicentesanchez:
So... SBT bit me, the managed classes were not discarded and that's why everything was still compiling before I removed all target folders:
```
[error] /home/luis/development/repositories/github/bigcentech/circe/modules/tests/jvm/target/scala-2.12/src_managed/test/io/circe/ProductCodecSuite.scala:89: type mismatch;
[error]  found   : Option[(String, String, String, String, String, String, String)]
[error]  required: Product7[?,?,?,?,?,?,?]
[error]       Encoder.forProduct7("s0", "s1", "s2", "s3", "s4", "s5", "s6")(Cc7.unapply)
```
[...]

PR #673, opened 2017-06-20
make forProductN expect ProductN[...]

lvicentesanchez:
This PR tries to address #630 by changing the signature of  `Encoder.forProductN` to accept `Option[ProductN[...]]` instead of tuples. That allows to create encoders for case classes using the unapply method. 

I'm not sure if this was what you had in mind @travisbrown.

### DIFF
--- a/modules/tests/shared/src/main/scala/io/circe/tests/examples/package.scala
+++ b/modules/tests/shared/src/main/scala/io/circe/tests/examples/package.scala
@@ -82,7 +82,9 @@ package examples {
     )
 
     val decodeBar: Decoder[Bar] = Decoder.forProduct2("i", "s")(Bar.apply)
-    val encodeBar: Encoder[Bar] = Encoder.forProduct2("i", "s")(Bar.unapply)
+    val encodeBar: Encoder[Bar] = Encoder.forProduct2("i", "s") {
+      case Bar(i, s) => (i, s)
+    }
   }
 
   object Baz {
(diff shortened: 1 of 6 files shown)


project/Boilerplate.scala
```diff
@@ -268,13 +269,13 @@ object Boilerplate {
...
         -   */
-        -  final def forProduct$arity[${`A..N`}, Source]($memberNames)(f: Source => (${`A..N`}))(implicit
+        -  final def forProduct$arity[${`A..N`}, Source]($memberNames)(f: Source => Option[$outputType])(implicit
         -    $instances
         -  ): ObjectEncoder[Source] =
         -    new ObjectEncoder[Source] {
         -      final def encodeObject(a: Source): JsonObject = {
         -        val members = f(a)
-        -        JsonObject.fromIterable(Vector($kvs))
+        -        JsonObject.fromIterable(Vector($kvs).flatten)
```
lvicentesanchez: Not happy about this.

modules/tests/shared/src/main/scala/io/circe/tests/examples/package.scala
```diff
@@ -109,9 +107,7 @@ package examples {
     )
 
     val decodeBam: Decoder[Bam] = Decoder.forProduct2("w", "d")(Bam.apply)(Wub.decodeWub, implicitly)
-    val encodeBam: Encoder[Bam] = Encoder.forProduct2[Wub, Double, Bam]("w", "d") {
-      case Bam(w, d) => (w, d)
-    }(Wub.encodeWub, implicitly)
+    val encodeBam: Encoder[Bam] = Encoder.forProduct2[Wub, Double, Bam]("w", "d")(Bam.unapply)(Wub.encodeWub, implicitly)
```
travisbrown: Looks like Scalastyle is unhappy about the long line here.

lvicentesanchez:
No, I don't think is needed and I don't think it makes sense to make a breaking change just to make it play better with unapply. I have just made a commit that removes it.

travisbrown:
Thanks, @lvicentesanchez!

PR #697, opened 2017-07-03
Rework JsonNumber [WIP]

travisbrown:
This is a work-in-progress PR that aims to use the fact that jawn 0.11 can now work with `CharSequence` values instead of strings, which means that for number parsing we don't need to instantiate additional strings in many situations.

Decoding into most numeric types is currently significantly faster (like 60% more throughput for `decodeDouble` in the new number parsing benchmark, and 11% more for `decodeLong`), but in some situations parsing is slower, since we no longer create lazily parsed `JsonNumber` instances from known valid JSON number strings. This is because we want to avoid instantiating strings for every number whenever we can.

I'm not sure I'm entirely happy about this trade-off, and I'm still experimenting with different options here.


hamnis:
Closed due to age.

Issue #711, opened 2017-07-17
Pass AsyncParser modes in circe-streaming


ImLiar:
Currently I'm in process of integrating _circe-streaming_ and _io.iteratee_ for request processing in finch: https://github.com/finagle/finch/pull/812

We've found out that there is a hardcoded `AsyncParser.Mode` inside of `circe-streaming`: https://github.com/finagle/finch/pull/812#discussion_r127653437

Question: is it possible to make some kind of configuration for `io.circe.streaming` enumeratees to pass required mode?

I could contribute a solution if it's sensible.

ImLiar:
Merged

PR #712, opened 2017-07-17
Enumeratees with different parsing modes

ImLiar:
Related to: https://github.com/circe/circe/issues/711

Additional enumeratees for `circe-streaming` with different `AsyncParser.Mode`s


modules/streaming/src/main/scala/io/circe/streaming/ParsingConfiguration.scala
```diff
@@ -0,0 +1,11 @@
+package io.circe.streaming
+
+import jawn.AsyncParser
+
+case class ParsingConfiguration(parseMode: AsyncParser.Mode)
+
+object ParsingConfiguration {
+
+  implicit val defaultParsingConfiguration: ParsingConfiguration =
```
vkostyukov: I'm not a huge fan of implicit parameters used for configuration. I'd rather prefer us to be explicit about this. One way of doing this is to encode the parsing mode in the function name (e.g., `stringStreamParser` and `byteStreamParser` - maybe there are better names).

An alternative solution would be to add a new parsing mode to Jawn (something like `Smart` mode) when it tries to use `StreamValue` by default and then fall back to `UnwrapArray` if the top-level object is an array.

ImLiar: @vkostyukov I'm afraid it's not possible to distinguish `SingleValue` and `StreamValue` modes easily

ImLiar: @vkostyukov I'd go with implicitis for the sake of backward compatibility. In other case it'd be required to rename old functions in favour of new logic.

vkostyukov: I don't think we have to rename. We can keep `stringParser`/`byteParser` in place and just document what parsing mode they use.

I guess I'm missing something, but I'm not sure I understand the reason we need `SingleValue` given `StreamValue` exists.

travisbrown:
I'm traveling right now but I'm also -1 on implicit configuration, and would much prefer breaking compatibility than adding it here.

modules/streaming/src/main/scala/io/circe/streaming/package.scala
```diff
@@ -6,23 +6,58 @@ import io.circe.jawn.CirceSupportParser
...
+  /**
+    * String parser of JSON values wrapped in a single array
+    */
+  final def stringArrayParser[F[_]](implicit F: ApplicativeError[F, Throwable]): Enumeratee[F, String, Json] =
+    stringParser(AsyncParser.UnwrapArray)
+
+  /**
+    * String parser of JSON values delimited by whitespace/newline
+    */
+  final def stringStreamParser[F[_]](implicit F: ApplicativeError[F, Throwable]): Enumeratee[F, String, Json] =
```
vkostyukov: I like it. Even though I can totally see how stream parser vs. array parser may be misinterpreted as the "parser that takes stream or array as an input", I think these names our best bet (given we document the difference).

ImLiar:
👍

travisbrown:
At a first pass this looks good to me—I'll aim to get it merged later today.

vkostyukov:
@travisbrown @ImLiar Anything's left here?

travisbrown:
No, just me getting my act together. :smile:

By the way I was hoping to publish an iteratee.io 0.13.0 release this week, before the circe 0.9 milestone. The main thing I'm doing in the 0.13 release is using the new cats-effect type classes in iteratee-files, and providing a module for `cats.effect.IO`. (I hoped to have this done last weekend but I'm moving across the country in a couple of weeks and all my weekends have been totally lost to moving- and packing-related tasks recently.)

vkostyukov:
Thanks, Travis! The moving burden is totally understandable - even moving within a city makes me crazy nervous!

I think I will cut a 0.16-M2 Finch release once iteratee is available.

ImLiar:
Oh, this is awesome! Thanks
Waiting for new release 👍

PR #714, opened 2017-07-23
Update Scalastyle and other plugins

travisbrown:


alexarchambault:
Beware if you switch to sbt-coursier `1.0.0-RC8` right now, it has some issues with `publishSigned` (https://github.com/coursier/coursier/issues/619), which should be fixed in `1.0.0-RC9`.

travisbrown:
Thanks @alexarchambault, I've switched back to RC7 for now.

PR #723, opened 2017-08-10
Add sttp to related project

adamw:
The latest sttp release, 0.0.6, contains a circe extensions module.


travisbrown:
Thanks, @adamw, this is great!

PR #724, opened 2017-08-10
Upgrade to cats 1.0.0-MF

dwijnand:


rkrzewski:
I was just about to post almost identical PR :)
I'm also seeing the same test failures. I did some investigation and it appears that partial unification is messing with derivation of partial decoders. 

In this test case: https://github.com/circe/circe/blob/e9d99ba0ab80694e45ca969b02dcb68bd6ee880b/modules/generic/shared/src/test/scala/io/circe/generic/AutoDerivedSuite.scala#L127-L134
Decoding ends up here:
https://github.com/circe/circe/blob/e9d99ba0ab80694e45ca969b02dcb68bd6ee880b/modules/core/shared/src/main/scala/io/circe/SeqDecoder.scala#L28

My guess is that `Function1[Int, Qux[String]]` is partially applied as `[A]Function1[Int, A]` and this type has the shape of `C` type parameter of `SeqDecoder[A, C[_]]`. I have no idea though, how does the compiler arrive here through the `implicit def`s in https://github.com/circe/circe/blob/master/modules/core/shared/src/main/scala/io/circe/Decoder.scala

Interestingly, only auto derived decoder is affected with this problem. semi-auto derivation of incomplete decoder is OK: https://github.com/circe/circe/blob/e9d99ba0ab80694e45ca969b02dcb68bd6ee880b/modules/generic/shared/src/test/scala/io/circe/generic/SemiautoDerivedSuite.scala#L27-L28

Anyway, adding `.disablePlugins(PartialUnification)` to `genericBase` and `genericExtrasBase` makes the test failures go away. Hopefully it'll be possible to guide the compiler to avoid inferring `SeqDecoder` in this situation with additional constraints - I tried fiddling with that with no luck so far.

dwijnand:
I completely forgot what Travis had said to me on twitter:

> adding an sbt plugin to avoid providing type parameters in a single place seems kind of like overkill, though.

I'll try and get rid of the plugin entirely.

On a complete tangent, @rkrzewski what's the markup for these fancy code block references?

rkrzewski:
I agree, but it may show up as a problem on the Circe user's side: when they'll use Cats 1.+, they'll probably use `-Ypartial-unification` to have `traverse` work and then it'll interfere with codec derivation. In the tests only incomplete auto-derived decoders are affected but I can imagine that if any multi-parameter type constructor can be partially applied to conform to `C[_]` shape, `SeqDecoder` may pop-up in some unexpected places.

Regarding the code blocks, just paste URL or a source file on github and use a line range eg. `#L127-L134` as the fragment.

rkrzewski:
I was trying to dig into the failure in `AutoDerivedSuite Decoder[Int => Qux[String]] - should decode partial JSON representations`
I tried bringing decoders into the test's scope in various ways, and here's what I found:
1) using semiauto derivation makes the test case succeed
```scala
implicit val decoder = io.circe.generic.semiauto.deriveFor[Int => Qux[String]].incomplete
```
2) doing what `AutoDerivation` and `LowPriorityDecoders` do (AFAICT) by hand also makes the test case  succeed
```scala
implicit val decoder = implicitly[io.circe.export.Exported[Decoder[Int => Qux[String]]]].instance
```
3) summoning the decoder directly has no effect (`SeqDecoder` is derived and testcase fails)
```scala
implicit val decoder = implicitly[Decoder[Int => Qux[String]]]
```
2 and 3 should basically be the same thing, right? Maybe it's a scalac bug?

The upside is that even with `-Ypartial-unification` on, `semiauto` provides a reliable workaround when automatic derivation goes wrong.

travisbrown:
Thanks for the investigation, @dwijnand, @rkrzewski.

I don't have a sense of why the compiler might think that `SeqDecoder` is a candidate here—there definitely shouldn't be `CanBuildFrom` instances for `I => ?` types.

My feeling is that we could release 0.9.0-M1 with this as a known issue (i.e. "if you use `-Ypartial-unification` and `auto`, incomplete decoders will not work"), and then make sure it's resolved before 0.9.0.

rkrzewski:
> I don't have a sense of why the compiler might think that SeqDecoder is a candidate here—there definitely shouldn't be CanBuildFrom instances for I => ? types.

Yeah, there must be a rouge `CanBuidFrom` instance around, because the decoder gets derived and then fails and runtime. Tracking down it's source is a promising line of investigation.


> My feeling is that we could release 0.9.0-M1 with this as a known issue (i.e. "if you use -Ypartial-unification and auto, incomplete decoders will not work"), and then make sure it's resolved before 0.9.0.

:+1:

dwijnand:
@travisbrown I went with your suggestion. However I didn't know where to document the known issue, so for now I've put it next to the reference to incomplete decoders in the docs.

PR #727, opened 2017-08-15
update to cats 1.0.0-MF

chris-vale-ck:
since the MF version should be API compliant this seems like a reasonable time to start porting over. It all went nearly seamlessly for a release noted for breaking changes. The only thing the least work was the `traverseU` -> `traverse` To do that I created a shim type that converts `Either[ParserException,Json]` to `Shim[Json]` This seems to trick the compiler into agreeing that the type fits the definition of `traverse` (where the `S => G[B]` has a `G` of only one slot) I honestly don't know if it's a good or bad way of fixing the issue but it works.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -20,7 +20,7 @@ lazy val compilerOptions = Seq(
   "-Yno-predef"
 )
 
-lazy val catsVersion = "0.9.0"
+lazy val catsVersion = "1.0.0-MF"
 lazy val jawnVersion = "0.11.0"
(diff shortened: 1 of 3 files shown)


Wogan:
Appreciate the effort though there is already an open pull request for this, see #724.

travisbrown:
Apologies for the duplication of work. I'm merging #724 now but still need to update iteratee.io before we can publish the 0.9.0-M1 milestone.

chris-vale-ck:
it's all good, I didn't notice #724 when I started this

Issue #598, opened 2017-03-27
NPE when serializing case class with fields set to NULL


haggy:
Repro:

```scala
import io.circe.generic.auto._
import io.circe.syntax._

case class StringFieldsSetToNull(field1: String, field2: String)
val n = StringFieldsSetToNull(null, "foo")

n.asJson.noSpaces // <-- This causes NPE since field1 is null
```

The error:
```
java.lang.NullPointerException
	at io.circe.Printer.printJsonString(scratch_117.scala:168)
	at io.circe.Printer.printJsonAtDepth(scratch_117.scala:227)
	at io.circe.Printer.printJsonObjectAtDepth(scratch_117.scala:200)
	at io.circe.Printer.printJsonAtDepth(scratch_117.scala:230)
	at io.circe.Printer.pretty(scratch_117.scala:240)
	at io.circe.Json.noSpaces(scratch_117.scala:103)
	at #worksheet#.get$$instance$$res0(scratch_117.scala:7)
	at #worksheet#.#worksheet#(scratch_117.scala:22)
```

Now I don't think this needs to be "fixed" in the sense that Circe knows how to gracefully handle it. I just think circe should have better logging of what field (if possible) is null when it shouldn't be and of course avoid NPE's wherever possible. 

Apologies if there's already an issue for this. I couldn't find one. Let me know if you need any other info.

knutwalker:
Using `null` directly is quite unusual for Scala and libraries often treating it like it doesn't exist (or rather, is used). More explanation also in https://github.com/circe/circe/issues/538#issuecomment-271174626.

haggy:
@knutwalker I agree that you would not want to pass in null explicitly, that was just done for the most simply repro case. My issue was a little different in that we had data coming in via JSON but we have a client that uses a different json lib (`json4s` with jackson backend in this case) which allowed the JSON null to map directly to a `NULL` in the case class for that field. Because of that, a case class with a null field was passed into `asJson` and subsequently `noSpaces`. 

The odd thing is that the serialization methods from `Json -> String` treat null differently than the transformer from cass class `T -> Json`. In the repro case above, `n.asJson` works fine. It is only a problem when you try to then call `noSpaces` on that `Json` instance.

travisbrown:
@haggy `n.asJson` "works" because the NPE won't happen until the null string is dereferenced. This matches the behavior of the standard library:

```scala
scala> val strings = List("foo", null)
strings: List[String] = List(foo, null)

scala> strings.size
[...]

PR #734, opened 2017-08-23
Handle null explicitly to prevent cryptic exception

alenkacz:
When you for any reason get an object that contains nulls, circe just prints out cryptic exception. From my point of view:
- it should print out better exception
- handle the situation automatically

In this pull request I picked the second solution because I think that null can be always serialized to json :-)

Example of such exception:
```
null
scala.MatchError: null
	at io.circe.Encoder$$anon$28.apply(Encoder.scala:216)
	at io.circe.Encoder$$anon$28.apply(Encoder.scala:213)
```

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Encoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Encoder.scala
@@ -224,6 +224,7 @@ final object Encoder extends TupleEncoders with ProductEncoders with MidPriority
     final def apply(a: Option[A]): Json = a match {
       case Some(v) => e(v)
       case None => Json.Null
+      case null => Json.Null
     }
   }
 


travisbrown:
I've resisted this kind of change in the past—see for example [this comment](https://github.com/circe/circe/issues/598#issuecomment-289589725).

The main issue is that I want to aim for consistency in the core module, even at the cost of users having to write a few extra lines here and there. I'm not sure I see a principled reason to encode `null: Option[A]` as a JSON null but not an arbitrary `null: A`, or to provide null-checking that's not given in similar situations by the standard library (or Cats, or a majority of other Scala libraries).

This has come up enough times that I think we may want some kind of solution, but I'm still not convinced it should be in circe-core.

travisbrown:
Closing this for now for the reasons given above, but am happy to reconsider other approaches to the general problem in the future.

PR #744, opened 2017-09-05
Renames semi-auto ValueClassCodecs to UnwrappedCodecs

pjan:
Addresses the naming remarks made in #661 (I fully agree, `...Unwrapped...` is way better/more consistent).

PR at the same time increases the `decoding` test coverage by including a failure case.


travisbrown:
Thanks @pjan!

PR #750, opened 2017-09-10
update refined to 0.8.3

sh0hei:


travisbrown:
Thanks! Will merge when green.

Issue #732, opened 2017-08-22
Include Java 8 time codecs in core for Scala 2.12


travisbrown:
Since Scala 2.12 requires Java 8 it might make sense to have a version-specific trait mixed into the `Decoder` companion object that would provide the `java.time` codecs there (without a separate dependency / import). I think with the right SBT magic we could do this without (much) code duplication.

gpoirier:
I created a PR for this ticket. Hopefully that's what you had in mind. I should note that the time codecs are a cut-n-paste from the time project but split in two different traits to be mixed in `Encoder` and `Decoder`. The only change that was made was to change some of the `val` to `def` because it was breaking the serialization tests for the other codecs, since `DateTimeFormatter` is not serializable.

gpoirier:
I'm just realizing a potential issue. I didn't remove the java8 project, and it might still be useful for people that wants to use java.time on Scala 2.11 with jdk8. However, I'm not sure how to share the code between core and java8. May be the best fix is to leave the time codecs in the shared sources, and use an additional trait to have it mixed in `Encoder` and `Decoder` only in Scala 2.12. Having the time codec instances shouldn't cause in the jars shouldn't cause issue in jdk7 if they aren't mixed in `Encoder` and `Decoder` I assume?

gpoirier:
I just tried the approach described above, but it seems to fail with scalajs, so it might not be possible to include time codecs at all in the Scala 2.10 and Scala 2.11 code base, even if they are not mixed in `Encoder` and `Decoder` objects?

travisbrown:
Done in #754 and #945.

PR #754, opened 2017-09-21
Adding time encoders and decoders for scala 2.12

gpoirier:
This is a Pull Request for the following issue: https://github.com/circe/circe/issues/732

It seems to be the best way to avoid duplication with the cross version support. But it does leave an awkward empty trait in Scala 2.10 and Scala 2.11.


modules/core/shared/src/main/scala-2.10/io.circe/TimeInstances.scala
```diff
@@ -0,0 +1,8 @@
+package io.circe
+
+trait TimeEncoders {
```
travisbrown: Can we make these traits package-private as well?

travisbrown: Also I'd be tempted to name them `JavaTimeEncoder`, etc., but not a big deal either way.

gpoirier: The only reason the name was `TimeEncoders` was to match what was in the java8 project, I have no preference personally, I'll make the change. But did you omit the "s" intentionally? The other Encoder traits mixed in `Encoder` all have an "s" at the end (`TupleEncoders`, `ProductEncoders`,  `MidPriorityEncoders`).

travisbrown: Thanks, and yes, I meant to include the `s`.

gpoirier:
I can't find the document again with the steps to create a pull request. Was this one good to go, or was it missing something like a rebase or something like that?

travisbrown:
I do still think we want to get the Scala.js issue sorted, and haven't had time to take a closer look at it yet, myself. Otherwise it looks good to me.

gpoirier:
I created a [branch](https://github.com/gpoirier/circe/tree/gpoirier/time-2) with the java time encoders/decoders in the main source file, and moved the tests from the scala-2.12 folder to the scala folder. When running `sbt clean test` locally, it failed. But then I pushed to github, and the travis [build](https://travis-ci.org/gpoirier/circe/builds/279118697) actually passed. I'm not sure why it was failing locally.

travisbrown:
I cherry-picked these commits into #945, so this functionality with be available in 0.10.0-M2.

gpoirier:
That's good new, looking forward to the 0.10.0 release.

PR #757, opened 2017-09-22
Use ObjectEncoder in disjunctionCodecs

nigredo-tori:
Make `encodeEither` and `encodeValidated` in `disjunctionCodecs` return `ObjectEncoder`s to match corresponding functions in `Encoder`.


travisbrown:
Good catch, thanks!

PR #761, opened 2017-09-26
Bump sbt to version 1.0.2

jonas:


travisbrown:
Nice, thanks @jonas—I'd been putting this off until I updated iteratee.io, where I ran into [this issue](https://stackoverflow.com/q/46271505/334519), but I'd forgotten that this update doesn't use `copy(id = ...)`. (I should probably be using your approach here in iteratee.io as well.)

Will merge when green.

Issue #735, opened 2017-08-26
Configured JSON codec fails serialization law in 0.9.0-M1


liff:
I added a failing test case [in my fork](https://github.com/liff/circe/blob/issue/configured-serialization-failure/modules/generic-extras/src/test/scala/io/circe/generic/extras/ConfiguredSemiautoDerivedSuite.scala).

[Relevant commit](https://github.com/liff/circe/commit/d2856c246ef976d244cf8df97f82ec6213d7f963#diff-3c91e4b3474c46bb6d3c7fb6a4c8ce75)

Relevant scalatest output:

```
[info] Codec[ConfigExampleBase]
[info] - should obey codec.consistency with accumulating
[info] - should obey codec.decoder serializability *** FAILED ***
[info]   NotSerializableException was thrown during property evaluation.
[info]     Message: org.scalatest.Assertions$AssertionsHelper
[info]     Occurred when passed generated values (
[info]
[info]     )
[info] - should obey codec.encoder serializability *** FAILED ***
[info]   NotSerializableException was thrown during property evaluation.
[info]     Message: org.scalatest.Assertions$AssertionsHelper
[info]     Occurred when passed generated values (
[info]
[info]     )
[info] - should obey codec.roundTrip
```

liff:
Looks like it's just because the `Decoder` and `Encoder` capture an outer pointer to the contaning class.

Moving the `ConfigExampleBase` & friends to a companion object or such fixes the issue.

Close or PR (with `CodecTests` law checks) with fix?

travisbrown:
Thanks for catching this. We should definitely fix it—I'm happy to take a look later today.

PR #763, opened 2017-09-27
Add law tests for configured JSON codecs

liff:
Kind of fixes/inspired by #735.

Not sure if any of this is really necessary but you get to have it anyway :)


travisbrown:
Thanks much for doing this, @liff, and sorry for the delay in review. I'll try to fix the `Serializable` issues that are causing the build to fail this evening, and then merge and open a follow up.

travisbrown:
Commits included in #796.

PR #764, opened 2017-10-05
Update refined to 0.8.4

fthomas:


travisbrown:
Thanks much @fthomas! Will merge when green.

PR #765, opened 2017-10-14
Update Monocle to 1.5.0-cats-M1

jhnsmth:
This PR updates Monocle to the first version based on cats 1.0.0-MF. Maybe I rushed too early and it's better to wait a stable release, if so let's keep the PR open and I will update it when a stable version is released.


jhnsmth:
Oops, I haven't noticed that monocle on cats doesn't support scala 2.10. Sorry for the noise.

PR #772, opened 2017-10-30
added Chartboost as adopter

ckozak:
Added [Chartboost](https://www.chartboost.com) to the list of circe adopters


travisbrown:
Excellent, thanks @ckozak!

PR #773, opened 2017-11-01
Update to Cats 1.0.0-RC1

travisbrown:
I want to wait to publish until we have an iteratee.io release for Cats 1.0.0-RC1 (see [this issue](https://github.com/travisbrown/iteratee/pull/218)), but this seems to compile.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -20,7 +20,7 @@ lazy val compilerOptions = Seq(
   "-Yno-predef"
 )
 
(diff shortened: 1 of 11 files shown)


travisbrown:
Okay, after #774 this is no longer WIP.

travisbrown:
It'll take me a bit to run all the Scala.js tests, etc., but I'm planning to publish 0.9.0-M2 this afternoon.

PR #774, opened 2017-11-01
Remove circe-streaming module

aeons:
`circe-streaming` was extracted.

I have put it here for now: https://github.com/aeons/circe-streaming

I just did `git filter-branch` with a subdirectory, so commits touching `circe-streaming` in other files, like `build.sbt` are not captured.


aeons:
The "missing" parts were basically copied from `circe-fs2`

travisbrown:
Excellent, thanks, @aeons! Will merge when green, then #773 should pass and we can publish 0.9.0-M2. I'll follow up on getting the new repo moved to the circe org after all the other dust settles.

travisbrown:
(And right, I've never bothered with a better way to do this than filter-branching with a subdirectory, so your approach looks reasonable to me.)

travisbrown:
The failure for that one job looked spurious so I restarted it.

PR #785, opened 2017-11-27
Add Zendesk to the list of adopters

erwan:


travisbrown:
Excellent, thanks @erwan!

PR #790, opened 2017-11-30
Added kebab-case member name and case constructor transformations

tlangs:
Added kebab-case member and case constructor transformations + tests


travisbrown:
Looks good to me, thanks!

PR #794, opened 2017-12-07
Printer support for escaping non-ASCII chars

haraldme:
Adds Printer support for having all non-ASCII chars in strings \u-escaped.

This can be useful when generating JSON that is to be used on a transport layer that is not unicode safe (e.g. in an HTTP header).

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Printer.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Printer.scala
@@ -35,6 +35,7 @@ import scala.annotation.switch
  *        storage.
  * @param predictSize Uses an adaptive size predictor to avoid grow-and-copy steps while printing
  *        into a binary output.
+ * @param escapeNonAscii Unicode-escape any non-ASCII characters in strings.
  */
 final case class Printer(
   preserveOrder: Boolean,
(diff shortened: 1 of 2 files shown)


travisbrown:
Thanks, @haraldme! One request: can we call this parameter `escapeNonAscii`? `alwaysEscapeUnicode` seems to me a little inexact.

haraldme:
Sure; I stole the name from json4s.  Do you agree that having a convenience method for turning the parameter on is a good idea?  Without it, you'd have to do something like

    foo.asJson.pretty(Printer.noSpaces.copy(escapeNonAscii = true))

instead of

    foo.asJson.pretty(Printer.noSpaces.withNonAsciiEscaped)

If you agree, is `withNonAsciiEscaped` an OK name for the convenience method?  I think it's a bit of a mouthful, so I'm open for better suggestions.

travisbrown:
Actually, do you mind removing the convenience method? (I missed it in the review round this morning.) If we add something like this I'd like to have a consistent offering for at least the most commonly used configuration options (and probably all of them). Thanks!

travisbrown:
To follow up: I'm not personally 100% convinced that the convenience methods are necessary, but I'd be happy to see them added if other people find them useful—I just think we should do that consistently in a separate PR.

travisbrown:
Excellent—thanks much! Will merge when green.

PR #802, opened 2017-12-14
Encoders/Decoders for Java boxed types

howyp:
Adds support for decoding/encoding in Java's boxed types. I wasn't sure:

- Where/if to add some tests - I couldn't find existing tests for the Decoder/Encoder of Scala's value types apart from specific ones around out-of-range values
- Whether the new instances should pick up the ones for Scala's value types implicitly or just directly

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -466,6 +466,14 @@ final object Decoder extends TupleDecoders with ProductDecoders with LowPriority
   }
 
   /**
+    * Decode a JSON value into a [[java.lang.Boolean]].
+    *
+    * @group Decoding
+    */
+  implicit final def decodeJavaBoolean(implicit d: Decoder[Boolean]): Decoder[java.lang.Boolean] =
+    d.map(java.lang.Boolean.valueOf)
+
+  /**
    * @group Decoding
    */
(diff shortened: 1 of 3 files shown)


travisbrown:
Thanks @howyp! The tests for `Int`, etc. are [here](https://github.com/circe/circe/blob/master/modules/tests/shared/src/test/scala/io/circe/CodecSuites.scala#L40), and look like this:

```scala
checkLaws("Codec[Int]", CodecTests[Int].codec)
```

…which just requires that the available encoder and decoder for `Int` pass the codec laws, which enforce round-trip-ability and some other reasonable things.

Also I think I'd be inclined to prefer these instances to use the Scala versions explicitly, but I'd be curious to hear arguments for the version here.

howyp:
Ah thanks - I'll take a look at those tests. 

On using the Scala instances directly - I can't say I'm fussed either way. It seems unlikely to me that  people will want to provide their own instances of say `Decoder[Int]`, but didn't want to close that off unless there was a good reason to.

howyp:
I've added tests for all of the instances now, let me know if anything else is missing. 

I found that the codecs didn't resolve properly (it was diverging on `encodeMapLike` which I didn't have time to investigate) with the implicit parameters present, so I've removed them and referenced the scala value codecs directly.

travisbrown:
This looks good to me—thanks, @howyp!

PR #807, opened 2017-12-19
Add Earnest as adopter

BlackVegetable:
I am a recently hired engineer at Earnest. I noticed Earnest uses circe but isn't on your list. I was told to submit a PR and change that!


travisbrown:
Great, thanks!

PR #810, opened 2017-12-20
Adds java8 time instance for `OffsetTime`

### LINKED ISSUE #809: Missing Java8 time instance for OffsetTime
There are encoder/decoder instances for most of the time types, but `java.time.OffsetTime` doesn't have an instance.

s-nel:
Fixes #809


travisbrown:
👍, looks reasonable to me, thanks!

PR #811, opened 2017-12-22
[WIP] Codec typeclass #133

lorandszakacs:
Hello folks,

Trying to address issue #133 

You can see it in action in the `io.circe.generic.SemiautoDerivedCodecSuite` test.

There's still some work to be done here, but I wanted to get feedback first, because said work heavily depends on it.

Design choices:
 1. `Codec` is a trait, rather than a simple type alias
    a. In the first commit of this PR it was a type-alias, but then I had to change it. The full rationale, and the code changes can be seen in commit b0b755e ** of this PR. Basically, without this you cannot satisfy point 1 from @travisbrown's wishlist, namely that a `Codec` instance should be available (without imports) whenever there is an `Encoder` + `Decoder` available.
 2. lol, I really was expecting a longer list, but I guess this is it

Now, if we go down the route that @tpolecat mentioned in [this comment](https://github.com/circe/circe/issues/133#issuecomment-161133882), then the need for auto-magic `Codec` summoning is not necessary., ergo no need for a special trait, and a type alias will suffice. And I tend to agree with him. I'm fine either way, at the end of the day tough. The only case I think `Codec` is valuable anyway is reducing `semiauto` derivation boilerplate.

TODO:
- [x] write tests for `generic.extras.semiauto` derivation of `Codec`
- [ ] provide the `Codec` instances @travisbrown mentioned in issue #133 @ number 5 wishlist item, but this is dependent on keeping `Codec` as a trait.
- [ ] address other issues pointed out by the community

** Full [commit](https://github.com/lorandszakacs/circe/commit/b0b755e4bbcd5c07b33f54981711b437556ae289) message:
```
Transform Codec from type-alias to trait #133

This was done to allow the definition of an implicit def that creates
a Codec for some type 'A' if we have an Encoder + Decoder for it in scope.

This implicit def can't be put anywhere else because then nothing compiles
anymore due to implicit search divergence problems, and the only way to
put it into a "companion object" is if it's a top level type.

This can be reverted if it is deemed necessary that such an implicit def
is undesirable.
```


leonardehrenfried:
I like this PR a lot. Is there a reason it has fallen silent?

lorandszakacs:
@leonardehrenfried well, I can pick it up again, but I need input from the maintainers to finalize it. Stuck on the design decision: `trait` vs `type-alias`. Each with advantages and limitations.

travisbrown:
Sorry for the long delay on review here—I've been meaning to spend some more time working with this approach but haven't had a chance yet. I'm also interested in feedback from other users.

leonardehrenfried:
I have a slight preference for it being a `trait` but would be fine with a type alias. I personally would not expect an Encoder/Decoder pair to magically convert to a Codec.

I would simply get rid of this duplication: 

```scala
implicit val coordinates: Encoder[Coordinates]        = deriveEncoder
implicit val coordinatesD: Decoder[Coordinates]       = deriveDecoder
```

lorandszakacs:
@leonardehrenfried I use `Codec` for the same purpose.

modules/core/shared/src/main/scala/io/circe/Codec.scala
```diff
@@ -0,0 +1,38 @@
...
+  * @author Lorand Szakacs, lsz@lorandszakacs.com
+  * @since 21 Dec 2017
+  *
+  */
+trait Codec[A] extends Encoder[A] with Decoder[A]
+
+object Codec {
+  def apply[A](implicit instance: Codec[A]): Codec[A] = instance
+
+  def instance[A](encode: Encoder[A], decode: Decoder[A]): Codec[A] =
```
kubukoz: WDYT about making this curried? It'll probably be easier to pass anonymous functions:

```scala
case class Foo(value: String_

//more or less
Codec.instance[Foo] {
  _.value.asJson
} {
  case "foo" => Foo("Foo")
  //...
}
```

modules/core/shared/src/main/scala/io/circe/Codec.scala
```diff
@@ -0,0 +1,38 @@
...
+    new Codec[A] {
+      private val enc = encode
```
kubukoz: What's the benefit of having these private vals? `apply` implementations could use the parameters directly

lorandszakacs:
Closing since it was brought in in `0.12.0-M3` 😄

PR #812, opened 2017-12-24
Add Gutefrage as adopter

felixbr:
Hey Travis, 
I just wanted to say thank you for all your hard work. We're big circe fans at gutefrage.net and use it in many projects. It is by far one of the most enjoyable libraries to work with and has never let us down. I also often refer to circe as a shining example of Scala code and library design.

With that I wish you a Merry Christmas and all the best for 2018!

~ Felix (on behalf of the @gutefrage engineering team)


travisbrown:
Thanks so much @felixbr!

Issue #306, opened 2016-06-19
Clarify the role of DecodingFailure's "message"


travisbrown:
circe's `DecodingFailure` is a descendant of Argonaut's `(String, CursorHistory)` in `DecodeResult`, and in Argonaut there's some ambiguity about what the string represents—in some places it's just a string representation of the type, and in other places it's a more detailed message about what went wrong. Since it doesn't have a name in Argonaut (and the corresponding argument to e.g. `DecodeResult.fail` is just named `s`), the library doesn't seem to make any particular commitment to how the string should be interpreted.

I originally planned to provide more guidance here in circe, but we're at version 0.5 and `message` in `DecodingFailure` is still playing both roles, and now the new `DecodingFailure.fromThrowable` in #303 puts a printed stack trace in the message field, which introduces yet another kind of thing that it can do.

I'd be interested in any suggestions about how we can refine the fields in `DecodingFailure` to make it clear what each one should do. One straightforward approach would be just to add new fields: something like `name: String`, `message: String`, an optional `underlying: Option[Throwable]`, and the lazy `CursorHistory`, but I'm definitely open to other ideas.

tomjadams:
Without knowing the particular piece of code overly well, your suggestions seem like a good first approach. Have you tried it and seen how the code pans out?

liff:
Just my €0.000002:

Personally, I'd be very much in favour of more descriptive error messages.

I think, compared to `message` being just the name of the type, The `name`, `message`, `underlying` triplet sounds very good. In particular, the stack trace in `DecodingFailure.fromThrowable` would make more sense as a separate field. Right now the failure messages from `Decoder#emapTry` are rather inconveniently messy :)

jatcwang:
My preference is actual error classes representing common error cases. This gives the developers freedom to transform a decoding error into w/e output they need.

I think these should cover the common errors:
* Missing Field
* Incorrect Type (wrong JSON type)
* Invalid data (e.g. constraint violation across multiple fields like start cannot be after end for a java.time.Period)

fleipold:
While this is being pondered it would be nice to just use the Show instance added in #222 to produce the message. That way we would get the nice javascript-style path for the history which is much more usable for bigger documents.

hamnis:
Closing in cleanup run, If anyone cares about this, please comment and I'll reopen.

PR #813, opened 2017-12-24
Encoder and Decoder for java.time.ZoneId

isomarcte:
Encoder and Decoder for java.time.ZoneId

### DIFF
--- a/modules/java8/src/main/scala/io/circe/java8/time/TimeInstances.scala
+++ b/modules/java8/src/main/scala/io/circe/java8/time/TimeInstances.scala
@@ -2,6 +2,7 @@ package io.circe.java8.time
 
 import io.circe.{ Decoder, DecodingFailure, Encoder, Json }
 import java.time.{
+  DateTimeException,
   Duration,
   Instant,
   LocalDate,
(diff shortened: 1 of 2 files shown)


modules/java8/src/main/scala/io/circe/java8/time/TimeInstances.scala
```diff
@@ -36,6 +37,17 @@ trait TimeInstances {
 
   implicit final val encodeInstant: Encoder[Instant] = Encoder.instance(time => Json.fromString(time.toString))
 
+  implicit final val decodeZoneId: Decoder[ZoneId] =
+    Decoder[String].emap{
+      case s if ZoneId.getAvailableZoneIds.contains(s) =>
```
travisbrown: I know this might be silly, but the timezone database is a big horrible global mutable thingy, and I'd rather have this instance follow the `try Right(...)` pattern of the others, so that it doesn't throw an exception if somehow the database changes between these two lines.

modules/java8/src/main/scala/io/circe/java8/time/TimeInstances.scala
```diff
@@ -36,6 +37,17 @@ trait TimeInstances {
...
+        Right(ZoneId.of(s))
+      case s =>
+        Left(s"$s is not an recognized ZoneId.")
```
travisbrown: Can you change this message to just be `"ZoneId"` on the model of the others? We're revisiting the role of these messages in 1.0 (see #306), but for now I'd prefer consistency here.

isomarcte:
@travisbrown Thanks for the feedback. I've updated the PR accordingly. Let me know if you have any further thoughts.

travisbrown:
Thanks @isomarcte!

PR #814, opened 2017-12-26
use cats 1.0.0

lorandszakacs:
it's here!


travisbrown:
Thanks @lorandszakacs! I'll be publishing 0.9.0 today!

Issue #819, opened 2018-01-01
Automatic derivation fails for List in 0.9.0


ochrons:
The changes in codec macro derivation in 0.9.0 breaks when using `List` and its `::` class.

Error introduced in https://github.com/circe/circe/commit/5193b5b8bcc3910535d036f09adee259493c7cb8

For example this fails to compile:
```scala
import io.circe._, io.circe.generic.auto._, io.circe.parser._, io.circe.syntax._

sealed trait Foo
case class Bar(xs: List[String]) extends Foo
case class Qux(i: Int, d: Option[Double]) extends Foo

val foo: Foo = Qux(13, Some(14.0))

val json = foo.asJson.noSpaces
```
with
`ERROR: requirement failed: circeGenericInrBindingFor:: is not a valid identifier`

ochrons:
This only happens with full auto derivation, semiauto and JsonCodec work fine.

travisbrown:
@ochrons Thanks again for the report. This seems to be Scala.js-specific, and I'm guessing it's related to this [Scala.js issue](https://github.com/scala-js/scala-js/issues/1790), which is kind of frustrating, but the effect is bad enough that I think we should work around it here. I don't think I'll have time this morning, but I'll publish an 0.9.1 with a fix as soon as possible.

BenFradet:
Is is just a matter of turning `::` in the label into something alphanumeric (e.g. `Cons`)? If so happy to open a PR.

BenFradet:
I tried the above without success, at least at runtime, compiling went fine.

@travisbrown any indication on how to proceed?

sjrd:
When you do `TermName(s"circeGenericInrBindingFor$label")`, you should call `.encodedName` on it to encode all invalid characters the same way scalac does. Same thing everywhere you construct `Name`s out of arbitrary strings.

BenFradet:
@sjrd cool thanks, I'll try that out.

PR #820, opened 2018-01-01
Add integrated ScalaFiddle example to documentation.

ochrons:
This adds support for integrated ScalaFiddle code examples in sbt-microsites documentation.

The example code in quickstart.md has been converted into a ScalaFiddle, to provide an example of how the integration is used. In the code example the `List` is replaced with `Vector` due to #819 

Because sbt-microsites doesn't support Jekyll-plugins via regular Ruby dependencies, the plugin code is copied into `plugins` directory.

For more information on ScalaFiddle documentation integration, check out https://github.com/scalafiddle/scalafiddle-core/blob/master/integrations/README.md

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,5 +1,6 @@
 import ReleaseTransformations._
 import microsites.ExtraMdFileConfig
+import microsites.ConfigYml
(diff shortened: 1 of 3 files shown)


travisbrown:
Merged in #830, thanks @ochrons!

PR #822, opened 2018-01-04
Update Monocle to 1.5.1-cats

guersam:
Yay!


modules/optics/src/test/scala/io/circe/optics/OpticsSuite.scala
```diff
@@ -1,31 +1,23 @@
...
    * itself.
    */
-  implicit val doubleInstance: Equal[Double] = Equal.equal { (a, b) =>
-    (a.isNaN && b.isNaN) || scalaz.std.anyVal.doubleInstance.equal(a, b)
-  }
+  implicit override val catsKernelStdOrderForDouble: Order[Double] with Hash[Double] = 
+    new cats.kernel.instances.DoubleOrder {
+      override def eqv(x: Double, y: Double): Boolean =
+        (x.isNaN && y.isNaN) || x == y
+    }
```
guersam: Tried inheritance indirection like below but couldn't overcome the ambiguous implicit error, so that I ended up with overriding.
```scala
trait CirceSuiteWithCustomDoubleEq extends CirceSuite  {
  implicit val doubleInstance: Eq[Double] = // ...
}

class OpticsSuite extends CirceSuiteWithCustomDoubleEq {
  // ...
}
```

guersam:
Oops, Monocle 1.5.0-cats lacks Scala 2.10 version :(

guersam:
And the reason is that [newts](https://github.com/julien-truffaut/newts) doesn't support Scala 2.10.

travisbrown:
Thanks, @guersam! I think we should make this update and drop 2.10 support for circe-optics soon, but I'm not sure yet whether it should be in 0.10 or 1.0 (which is likely to drop 2.10 support across the board). In any case I really appreciate this work, and it definitely will be merged—just not sure on the timing yet.

travisbrown:
Excellent, thanks @guersam!

PR #828, opened 2018-01-18
Use encoded names in derivation macros

BenFradet:
Here's how I tested it:

```scala
sealed trait Foo
case class Bar(xs: List[String]) extends Foo
case class Qux(i: Int, d: Option[Double]) extends Foo

val foo: Foo = Bar(List("abc", "def", "ghi"))
val json = foo.asJson.noSpaces
println(json)
```


travisbrown:
It'd be nice to have a failing test, but we can do that in a follow-up PR. Thanks, @BenFradet!

BenFradet:
I can add it in this PR, no worries 👍

travisbrown:
Great, thanks @BenFradet! I'll publish 0.9.1 as soon as it's ready and merged.

BenFradet:
@travisbrown added an hygiene test for scala js, is this what you had in mind?

travisbrown:
Yep, looks good to me (I just confirmed that cherry-picking the test commit onto master fails).

BenFradet:
Cool stuff, looking forward to 0.9.1 👍

PR #838, opened 2018-01-30
Keep wrapped instance's decodeAccumulating method in Decoder.validate

### LINKED ISSUE #837: Accumulation of error messages is broken when a DerivedDecoder is wrapped using 'validate'
Accumulation of error messages is broken when a `DerivedDecoder` is wrapped using `validate`

```scala
scala> import io.circe._, io.circe.generic.auto, io.circe.generic.semiauto._, io.circe.parser._
import io.circe._
import io.circe.generic.auto
import io.circe.parser._

scala> case class Foo(bar: String, baz: Int)
defined class Foo

scala> val json = "{}"
json: String = {}

scala> decodeAccumulating[Foo](json)
res0: cats.data.ValidatedNel[io.circe.Error,Foo] = Invalid(NonEmptyList(DecodingFailure(Attempt to decode value on failed cursor, List(DownField(bar))), DecodingFailure(Attempt to decode value on failed cursor, List(DownField(baz)))))

scala> implicit val fooDecoder: Decoder[Foo] = deriveDecoder[Foo].validate(c => true, "foobar")
fooDecoder: io.circe.Decoder[Foo] = io.circe.Decoder$$anon$25@701cf91b

scala> decodeAccumulating[Foo](json)
res1: cats.data.ValidatedNel[io.circe.Error,Foo] = Invalid(NonEmptyList(DecodingFailure(Attempt to decode value on failed cursor, List(DownField(bar)))))
```
As you can see from this minimal example, after adding the validation we only get 1 DecodingFailure, the one about the field "baz" disappears.

I think the problem is in the wrapper encoder implementation as it's reusing the wrapped instance `apply` method, but not `decodeAccumulating`. This means, while the original `DerivedDecoder` has this implemention

```scala
override def decodeAccumulating(c: HCursor): AccumulatingDecoder.Result[A] =
      decode.value.decodeAccumulating(c).map(gen.from)
```

the new instance returned by `validate` will fall back the one in the `Decoder` trait:

```scala
 private[circe] def decodeAccumulating(c: HCursor): AccumulatingDecoder.Result[A] = apply(c) match {
    case Right(a) => Validated.valid(a)
    case Left(e) => Validated.invalidNel(e)
  }
```

This can be fixed by reusing the wrapped instance method:

```diff
 trait Decoder[A] extends Serializable { self =>
   /**
@@ -126,6 +129,8 @@ trait Decoder[A] extends Serializable { self =>
   final def validate(pred: HCursor => Boolean, message: => String): Decoder[A] = new Decoder[A] {
     final def apply(c: HCursor): Decoder.Result[A] =
       if (pred(c)) self(c) else Left(DecodingFailure(message, c.history))
+
+    override def decodeAccumulating(c: HCursor): Result[A] = self.decodeAccumulating(c)
   }

```

voidcontext:
Fixes #837

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -126,6 +126,8 @@ trait Decoder[A] extends Serializable { self =>
   final def validate(pred: HCursor => Boolean, message: => String): Decoder[A] = new Decoder[A] {
     final def apply(c: HCursor): Decoder.Result[A] =
       if (pred(c)) self(c) else Left(DecodingFailure(message, c.history))
+
+    override def decodeAccumulating(c: HCursor): AccumulatingDecoder.Result[A] = self.decodeAccumulating(c)
   }
 
   /**
--- a/modules/tests/shared/src/test/scala/io/circe/DecoderSuite.scala
+++ b/modules/tests/shared/src/test/scala/io/circe/DecoderSuite.scala
@@ -1,6 +1,7 @@
 package io.circe
 
-import cats.data.Validated
+import cats.data.Validated.Invalid
+import cats.data.{NonEmptyList, Validated}
 import cats.kernel.Eq
 import cats.laws.discipline.{ MonadErrorTests, SemigroupKTests }
 import io.circe.CursorOp.DownArray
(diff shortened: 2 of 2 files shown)


travisbrown:
Good catch, thanks @voidcontext!

PR #842, opened 2018-02-09
Add download data for January 2018

travisbrown:
In January we had 413,553 downloads (from 17,279 unique IPs) for all artifacts in the circe organization, up from 313,264 in December (and from our previous all-time high of 355,131 in November 2018).

Also especially cool: Scala 2.10 now makes up only 3.8% of circe-core downloads, vs. 7.0% in December. Even better, for the 0.9.x releases specifically, 2.10 is only 1.1% of circe-core downloads, which makes me pretty confident that we can drop 2.10 support after 0.10 without making too many people angry.


propensive:
s/November 2018/November 2017/.
Well done!

travisbrown:
Ah, thanks Jon!

PR #843, opened 2018-02-15
Add Apollo Agriculture as user of circe

estsauver:
See [My Keybase](https://keybase.io/estsauver) for proof of domain ownership.


travisbrown:
Thanks, @estsauver!

PR #844, opened 2018-02-22
Update Macro Paradise to 2.1.1

travisbrown:


alexarchambault:
FYI for sbt-coursier, you can bump the version in `project/project/plugins.sbt`, and replace its line in `project/plugins.sbt` by `addSbtCoursier` (so that the coursier version is only set in one place).

travisbrown:
@alexarchambault Nice, thanks! I always forget about the other version since only some of my projects have it configured.

PR #847, opened 2018-02-27
@ConfiguredJsonCodec encodeOnly/decodeOnly support

espenhw:
I guess this was forgotten in #631 / #681 ?


travisbrown:
Good catch, thanks, @espenhw!

Issue #851, opened 2018-03-07
Different behaviour for optional decoders depending on whether they're at top level


hgiddens:
Given a `Decoder[A]`, trying to decode an `Option[A]` will fail differently (assuming the underlying `Decoder[A]` fails) depending on whether we're looking at the top level or not:

```
import io.circe.{Decoder, Json}

Decoder.
  instance(_.as[Option[Int]]).
  decodeJson(Json.fromString(""))
// Right(None)

Decoder.
  instance(_.downField("foo").as[Option[Int]]).
  decodeJson(Json.obj(("foo", Json.fromString(""))))
// Left(DecodingFailure(Int, List(DownField(foo))))
```

Is this behaviour deliberate? And if so is there an easy way out of it (so that the first example would also fail with the `Left`)?

Tested with 0.9.1 and current master; the behaviour from a quick glance looks like it's always been this way though.

travisbrown:
I'd categorize this as more "known" than "deliberate". It's behavior that's inherited from Argonaut, although Argonaut 6.2 no longer does this.

I think I agree that the only thing that should decode into a `None: Option[A]` at the top-level is a JSON `null`, and would be happy to see this change in 0.10. PR welcome, otherwise I'm happy to handle it this weekend.

hgiddens:
I'll have a go at fixing it up.

travisbrown:
Fixed in #853.

PR #853, opened 2018-03-08
Make optional decoders fail with Left at the top-level for non-nulls

hgiddens:
This addresses #851.

I think I've implemented the change correctly, so that only `null` at the top level decodes as `Right(None)` and decoding failures from the inner decoder are otherwise passed through. If this wasn't quite what you had in mind please let me know, I'm more than happy to fix it up.


travisbrown:
Thanks, looks good to me!

PR #860, opened 2018-03-18
Fix #849. Optimize performance for encoding/decoding when using custom transformations of member names or constructor

### LINKED ISSUE #849: Low performance for encoding/decoding when using custom member names transformation with semiauto
The issue is occured when member names should be rewritten, i.e. with using `io.circe.generic.extras.Configuration.default.withSnakeCaseMemberNames` regexp transformation are occured on every json encoding/decoding (see `io.circe.generic.extras.ConfiguredDecoder`, `io.circe.generic.extras.ConfiguredObjectEncoder`). 
I think they should be performed only once when encoder/decoder are derived.

vndzzz:
Remove calling transformMemberName and transformConstructorNames of Configuration on every encoding, decoding to/from json

### DIFF
--- a/modules/generic-extras/src/main/scala/io/circe/generic/extras/decoding/ConfiguredDecoder.scala
+++ b/modules/generic-extras/src/main/scala/io/circe/generic/extras/decoding/ConfiguredDecoder.scala
@@ -9,6 +9,8 @@ import shapeless.{ Annotations, Coproduct, Default, HList, LabelledGeneric, Lazy
 import shapeless.ops.hlist.ToTraversable
 import shapeless.ops.record.Keys
 
+import scala.collection.concurrent.TrieMap
+
 abstract class ConfiguredDecoder[A] extends DerivedDecoder[A]
 
 final object ConfiguredDecoder extends IncompleteConfiguredDecoders {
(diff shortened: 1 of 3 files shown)


travisbrown:
This looks great, thanks @vndzzz!

modules/generic-extras/src/main/scala/io/circe/generic/extras/decoding/ConfiguredDecoder.scala
```diff
@@ -54,9 +56,12 @@ final object ConfiguredDecoder extends IncompleteConfiguredDecoders {
     decodeR: Lazy[ReprDecoder[R]],
     config: Configuration
   ) extends ConfiguredDecoder[A] {
+
+    private[this] val constructorNameCache: TrieMap[String, String] = new TrieMap()
+
     final def apply(c: HCursor): Decoder.Result[A] = decodeR.value.configuredDecode(c)(
       Predef.identity,
-      config.transformConstructorNames,
+      v => constructorNameCache.getOrElseUpdate(v, config.transformConstructorNames(v)),
```
travisbrown: I'm wondering whether we should construct this function outside of `apply` to avoid allocations.

vndzzz: I think, you are right. I'll extract it outside of apply.

travisbrown:
Just to confirm—this is okay to merge? (Asking because of the "WIP" on the last commit.)

vndzzz:
@travisbrown yes it is okay to merge it.

PR #864, opened 2018-03-23
Updates list of adopters with REA Group

cwmyers:
We are heavy users of Circe, serving Australia's largest property website.


travisbrown:
Thanks @cwmyers!

PR #866, opened 2018-03-27
Check predicate in error-accumulation mode in Decoder#validate

### LINKED ISSUE #865: ErrorAccumulatingUnmarshaller from akka-http-json lib no longer applying validations in 0.9.2
We have some Decoders with attached validations like this:

```
implicit val decodeSomething = deriveDecoder[Something].validate(..., ...)
```

And we use this for our akka-http routes (https://github.com/hseeberger/akka-http-json):
```
trait AkkaHttpCirceSupport extends BaseCirceSupport with ErrorAccumulatingUnmarshaller
```

And after updating to 0.9.2 that validation is no longer applied.

Switching to `FailFastCirceSupport` makes the validation run again. It's probably related to https://github.com/circe/circe/pull/838/files, but I'm not sure.

Reverting to `0.9.1` makes `ErrorAccumulatingUnmarshaller` run validations again.

I put together a couple of tests showing this here: https://github.com/yurique/akka-http-json-fail/tree/validation-fail (validation-fail branch)

`sbt test` fails in one of the cases; changing circe to `v0.9.1` makes them pass.

I'm not sure if this is related to circe or to `akka-http-json` lib. 

/cc @hseeberger

travisbrown:
Fixes #865 and adds two tests, one of which fails without the change in `Decoder`. Will backport this for an 0.9.3 release as soon as it gets merged, and will merge as soon as it's green if there are no objections.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -127,7 +127,8 @@ trait Decoder[A] extends Serializable { self =>
     final def apply(c: HCursor): Decoder.Result[A] =
       if (pred(c)) self(c) else Left(DecodingFailure(message, c.history))
 
-    override def decodeAccumulating(c: HCursor): AccumulatingDecoder.Result[A] = self.decodeAccumulating(c)
+    override def decodeAccumulating(c: HCursor): AccumulatingDecoder.Result[A] =
+      if (pred(c)) self.decodeAccumulating(c) else Validated.invalidNel(DecodingFailure(message, c.history))
   }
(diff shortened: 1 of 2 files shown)


voidcontext:
Hm, I've just seen this PR. It definitely seems it's my fault. What baffles me right now is how I couldn't see this. I'm trying to recreate my thought process and see where I was wrong.

I might have been that I've seen that `apply` checks the predicate, which is called from the default `decodeAccumulating` method defined in the trait.

I'm also not sure how accumulation worked in the first place with the validated decoder in `akka-http-json`.

Can you please share your findings and thought process?

PR #869, opened 2018-04-04
Migrate cats-optics to monocle-cats

Odomontois:
Initial attempt to make obvious migration


Odomontois:
Closing since there is already #822

PR #873, opened 2018-04-05
Update quickstart's circe version

BenFradet:


travisbrown:
👍, thanks—will republish today.

PR #880, opened 2018-04-16
Fix stateful decoding of optional fields

### LINKED ISSUE #708: Decoder.state.decodeField[Option[_]] moves cursor on None
Thanks for your patience if I'm missing something.

I'm still new to `circe`… and `cats`… and `shapeless`… and have never tried pure haskell-- but I was trying to put together a generic version of #375 and tripped over a case where `Decoder.state.decodeField[Option[_]]` seems to move the cursor even when the value is `None`.

```scala
import io.circe.Decoder
import io.circe.parser._
import cats.instances.either._

case class Foo(a: Option[String], b: Int)

implicit val decodeFoo: Decoder[Foo] = Decoder.fromState(
  for {
    a <- Decoder.state.decodeField[Option[String]]("a")
    b <- Decoder.state.decodeField[Int]("b")
    _ <- Decoder.state.requireEmpty
  } yield Foo(a, b)
)

val good =
  """|{
     |  "a": "hello",
     |  "b": 1
     |}
     |""".stripMargin

val shouldBeGood =
  """|{
     |  "b": 1
     |}
     |""".stripMargin

println(decode[Foo](good))
println(decode[Foo](shouldBeGood))
```

outputs

```scala
Right(Foo(Some(hello),1))
Left(DecodingFailure(Attempt to decode value on failed cursor, List(DownField(a))))
```

I managed to copy the original [`decodeField`](https://github.com/circe/circe/blob/v0.8.0/modules/core/shared/src/main/scala/io/circe/Decoder.scala#L819-L829) to a modified [`decodeFieldOption`](https://github.com/broadinstitute/clio/blob/d1d35bf98bf379c257bb495ab426141a340c227f/clio-server/src/main/scala/org/broadinstitute/clio/util/json/StateDecoder.scala#L72-L86) that seems to work for our purposes. However it's not a PR as I'm not sure if it's "category-theory" correct. I'm still searching for a good underscore book/video to explain `StateT`, `Kleisli`, etc.

jadlr:
This fixes #708 

It is the simplest solution I could come up with that should work with all decoders that do not consume json. 
If a decoder succeeds even tough the field was non-existent we just use the initial cursor state.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -978,6 +978,7 @@ final object Decoder extends TupleDecoders with ProductDecoders with LowPriority
       val field = c.downField(k)
 
       field.as[A] match {
+        case Right(a) if field.failed => Right((c, a))
         case Right(a) => Right((field.delete, a))
         case l @ Left(_) => l.asInstanceOf[Result[(ACursor, A)]]
       }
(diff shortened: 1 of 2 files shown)


travisbrown:
Thanks for figuring this out, @jadlr, and thanks extra for the tests!

PR #891, opened 2018-04-25
Implement Raw* classes directly in circe-jawn

travisbrown:
After publishing 0.10.0-M1 yesterday, I noticed that the [benchmark](https://github.com/circe/circe-benchmarks) results looked kind of bad:

```
Benchmark                        Mode  Cnt      Score     Error  Units
ReadingBenchmark.readFoosCirce  thrpt  100   3385.838 ±  27.815  ops/s
ReadingBenchmark.readIntsCirce  thrpt  100  14774.716 ± 115.221  ops/s
```

Compared to the following for 0.9.3:

```
Benchmark                        Mode  Cnt      Score     Error  Units
ReadingBenchmark.readFoosCirce  thrpt  100   3432.876 ±  33.256  ops/s
ReadingBenchmark.readIntsCirce  thrpt  100  17841.714 ± 136.597  ops/s
```

The change in this PR recovers the ~17% of throughput we lost in the integer array reading benchmark:

```
Benchmark                        Mode  Cnt      Score     Error  Units
ReadingBenchmark.readFoosCirce  thrpt  100   3461.370 ±  19.840  ops/s
ReadingBenchmark.readIntsCirce  thrpt  100  17546.385 ± 281.803  ops/s
```

I probably won't publish an M2 yet unless someone feels very strongly that they need it.

### DIFF
--- a/modules/jawn/src/main/scala/io/circe/jawn/CirceSupportParser.scala
+++ b/modules/jawn/src/main/scala/io/circe/jawn/CirceSupportParser.scala
@@ -2,51 +2,51 @@ package io.circe.jawn
 
 import io.circe.{ Json, JsonNumber, JsonObject }
 import java.util.LinkedHashMap
-import jawn.{ Facade, FContext, SupportParser }
+import jawn.{ RawFacade, RawFContext, SupportParser }
 
 final object CirceSupportParser extends SupportParser[Json] {
-  implicit final val facade: Facade[Json] = new Facade[Json] {
-    final def jnull(): Json = Json.Null
-    final def jfalse(): Json = Json.False
-    final def jtrue(): Json = Json.True
-    final def jnum(s: CharSequence, decIndex: Int, expIndex: Int): Json =
+  implicit final val facade: RawFacade[Json] = new RawFacade[Json] {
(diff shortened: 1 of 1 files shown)


travisbrown:
Just for fun I implemented the `RawFacade` [in Java](https://github.com/circe/circe/tree/topic/jawn-perf-java) and am consistently seeing even more of an improvement:

```
Benchmark                        Mode  Cnt      Score     Error  Units
ReadingBenchmark.readFoosCirce  thrpt  100   3516.797 ±  36.290  ops/s
ReadingBenchmark.readIntsCirce  thrpt  100  17931.383 ± 139.069  ops/s
```

I'm not sure I want to go that route, but I'm curious about why the difference is this big, given that the Scala was already pretty Java-y.

knutwalker:
If i were to make a guess, I would say the difference comes from named subclasses for the {Single,Array,Object}Context vs ad-hoc anonymous implementations in Java vs Scala. Just a guess, though.

travisbrown:
@knutwalker My first draft of the Java version used anonymous classes there, and the results were roughly the same as the current approach in that branch (I just gave them names for clarity).

PR #893, opened 2018-05-09
Introducing documentation for ADT codecs

gvolpe:
- Introducing documentation for ADT encoders / decoders as explained [here](https://stackoverflow.com/questions/42165460/how-to-decode-an-adt-with-circe-without-disambiguating-objects).
- Reorganizing the Encoding and Decoding section in different sub-sections.
- All the examples are compiled with `tut`.

I took the liberty to reorganize the codecs section, let me know what you think :smiley:


travisbrown:
Thanks much, @gvolpe! At a first pass this looks great—will get it merged later today.

PR #901, opened 2018-06-03
Support for Money and Currency types

AndreasKostler:
Simple support for currencies.


travisbrown:
Thanks for the PR! It looks like it'll need a few minor changes to compile on 2.11, and to pass `sbt scalastyle`. My one bigger question is whether we could define the `Money` instances explicitly to avoid imposing a circe-generic dependency on users?

AndreasKostler:
Yes, I had similar thoughts. The library does what it says it does but the lack of activity is disconcerting. Let me have a think about it - I might provide support for https://github.com/typelevel/squants monetary types instead.

diesalbla:
@AndreasKostler There is a recent PR to add support for the Currency type from the Java library. #1578 

Given that this one is more standard, I suppose that the `monies` library should provide conversion functions to/from it? If so, the instances for encoders and decoders would just become mappings.

hamnis:
Closing in favor of #1578

PR #910, opened 2018-06-18
Update readme.md: added SpotX to list of adopters

dcastro:
We use circe in pretty much every scala project here at SpotX.
Thank you for this awesome library!


travisbrown:
Great to hear, thanks @dcastro!

PR #912, opened 2018-06-22
Resolves #911 by renaming UserCodec -> User

### LINKED ISSUE #911: Incorrect example in Semi-automatic Derivation
The example shown in Semi-automatic Derivation page when using `forProductN` helper methods is incorrect (https://circe.github.io/circe/codecs/semiauto-derivation.html#forproductn-helper-methods). The name of the object must be the same of the case class. I think that this might be obvious for experienced scala users, but as a newbie I spent some time before understanding why it wasn't working.

se7entyse7en:


travisbrown:
Thanks @se7entyse7en!

PR #914, opened 2018-06-23
Add circe-kafka to list of integrations

NeQuissimus:
I wrote a micro-library that makes using circe in combination with Kafka (Streams) very convenient by generating Kafka's `Serializer`, `Deserializer` and `Serde` implicitly from `Encoder` and `Decoder` instances.

This PR adds a reference to the library to the README


travisbrown-stripe:
Excellent, thanks!

Issue #917, opened 2018-07-01
Better error messages for circe-java8 time Decoders


felixbr:
Hi everyone,

I'm using the circe-java8 time package a lot and one thing that trips me up again and again is that parsing errors are currently thrown away and you'll only see a generic `DecodingFailure` in the logs.

In order to debug what's going on I usually have to do the following:

1. Log the raw json String if a `DecodingFailure` happens (non-trivial in certain frameworks/codebases)
2. Check what method the Decoder instance is using (e.g. `Instant#parse` for `Decoder[Instant]`)
3. Use Ammonite to pass the raw input from the json to the parsing method manually
4. Finally see what the problem is

The current implementation looks like this (`Decoder[OffsetDateTime]` as example):
```
final def decodeOffsetDateTime(formatter: DateTimeFormatter): Decoder[OffsetDateTime] =
    Decoder.instance { c =>
      c.as[String] match {
        case Right(s) => try Right(OffsetDateTime.parse(s, formatter)) catch {
          case _: DateTimeParseException => Left(DecodingFailure("OffsetDateTime", c.history))
        }
        case l @ Left(_) => l.asInstanceOf[Decoder.Result[OffsetDateTime]]
      }
    }
```
As you can see the `DateTimeParseException` is unused and `DecodingFailure` only contains `"OffsetDateTime"` as error message.
It would be very helpful to include the error message of the parsing failure in the `DecodingFailure` message (e.g. `"OffsetDateTime (DateTimeParseException: Text 'Sun Jul  1 10:41:03 CEST 2018' could not be parsed at index 0)"`.

Full example: https://gist.github.com/felixbr/fc2aec365cc390e59effc75bfb0ab378

If you agree this is a worthwhile endeavor, I will start working on a PR, which enhances the error message :)

Cheers
Felix

travisbrown:
I do agree this is a worthwhile endeavor! I've got some changes to error messages planned for circe-algebra, but I think going ahead and making this more useful for now should be a priority. Thanks, @felixbr!

travisbrown:
Fixed in #917.

PR #918, opened 2018-07-02
Add more error information to java8 time Decoder instances

felixbr:
Original issue: https://github.com/circe/circe/issues/917

Since this is my first time here my changes might not fit the stylistic choices of the rest of the project. Feel free to nitpick :)

I tried to keep the additional error text close to the original `java.time` text, but added the input String where it was missing so it is consistent (namely `Duration` and `Period`). This might make the `DecodingFailure` message a lot more verbose overall but also a lot more helpful if it appears somewhere in a log.

If you think this is too much I could also rewrite the error messages to be more concise and/or truncate the input String, so it doesn't put input text of arbitrary length in the message. I tried to keep it simple for now.

~Felix


felixbr:
This is already merged as a rebased PR #951

PR #921, opened 2018-07-09
fix a link to shapeless

saint1991:
As stated in the title.


travisbrown:
Thanks for catching this!

PR #922, opened 2018-07-11
Parser#decodeAccumulating does not accumulate errors

saint1991:
I found that Parser#decodeAccumulating doesn't accumulate errors.

```
import cats.syntax.apply._
import cats.syntax.either._

import io.circe._
import io.circe.parser._

case class Foo(name: String, age: Int)

object FooDecoder {

  implicit val decoder: Decoder[Foo] = Decoder.instance {
    cursor => for {
      name <- cursor.get[String]("name")
      age <- cursor.get[Int]("age")
    } yield Foo(name, age)
  }
}

object FooAccumulatingDecoder {

  implicit val accDecoder: AccumulatingDecoder[Foo] = AccumulatingDecoder.instance {
    cursor => (
      cursor.get[String]("name").toValidatedNel,
      cursor.get[Int]("age").toValidatedNel
    ).mapN(Foo.apply)
  }
}

// cannot use AccumulatingDecoder here...
println(decodeAccumulating("{}")(FooDecoder.decoder)) 
```

This produces the following result:
```
Invalid(NonEmptyList(DecodingFailure(Attempt to decode value on failed cursor, List(DownField(name)))))
```

My expectation from its name was:
```
Invalid(NonEmptyList(DecodingFailure(Attempt to decode value on failed cursor, List(DownField(name))), DecodingFailure(Attempt to decode value on failed cursor, List(DownField(age)))))
```

I think a problem here is that `decodeAccumulating` takes `Decoder` that is fail-fast implementation so I replace it with `AccumulatingDecoder`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -19,7 +19,8 @@ val compilerOptions = Seq(
   "-Ywarn-numeric-widen",
   "-Xfuture",
   "-Yno-predef",
-  "-Ywarn-unused-import"
+  "-Ywarn-unused-import",
+  "-Ypartial-unification"
 )
 
 val catsVersion = "1.1.0"
--- a/modules/core/shared/src/main/scala/io/circe/Parser.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Parser.scala
@@ -1,6 +1,6 @@
 package io.circe
 
-import cats.data.{ NonEmptyList, Validated, ValidatedNel }
+import cats.data.{ Validated, ValidatedNel }
 import java.io.Serializable
 
 trait Parser extends Serializable {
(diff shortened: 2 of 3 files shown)


saint1991:
Should I add a test for this?

travisbrown:
This is a tricky case, but I think we want the original behavior here. The idea is that `AccumulatingDecoder` isn't a type class, but more like a mode that a decoder can be converted into, and that any decoder can be written in such a way that it can accumulate errors.

There's some discussion of the motivation for this design decision here: https://meta.plasm.us/posts/2015/12/17/error-accumulating-decoders-in-circe/

The whole shape of this part of the API will change a bit in circe-algebra, so I'd lean toward leaving things as they are for now.

PR #924, opened 2018-07-13
Add ohNoMyCirce to related project.

djx314:
* Now I added shapeless and slick compile error messages support, include code snippet hints support.
* The ohNoMyCirce name will no longer change.

Fun for the insteresting repo.


travisbrown:
Thanks again, @djx314!

PR #928, opened 2018-07-28
Fix JVM options

jiminhsieh:
* `-Xmx` exceeds the limitation of Travis CI can offer. I use the same parameter as cats.
* `MaxPermSize` was replaced by `MaxMetaspaceSize`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: scala
 
+sudo: required
+
 scala:
   - 2.11.12
   - 2.12.6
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,5 +1,7 @@
 language: scala
 
+sudo: required
```
jiminhsieh: This will cause Travis CI to use GCP instance instead of EC2. Travis CI with GCP will have larger memory. [Here](https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system) is the detailed information from Travis CI.

travisbrown: Do we actually need the additional memory?

travisbrown:
Thanks for looking at this, @jiminhsieh!

.travis.yml
jiminhsieh: I tried `-Xmx` with 3G, it seems without any issue. Travis CI can offer 4G when we don't use `sudo: required`. I think we don't need to use 6G. If it's fine with 3G, I will push the change later. :)

Issue #927, opened 2018-07-27
JsonLiteralMacros silently supresses exceptons


pshirshov:
```
            case Left(t: Throwable) => c.abort(
              c.enclosingPosition,
              "Invalid JSON in interpolated string"
            )
```

It's completely impossible to diagnose what exactly went wrong in this situation. You can't put a breakpoint into a macro code, etc, etc.

travisbrown:
Fixed in #927.

PR #930, opened 2018-07-30
Fix for #927: verbose exception reporting in literal macro

pshirshov:
See #927 for details


travisbrown:
Thanks! I'm not sure I feel very strongly that this is necessary (in general JSON literals should be very short), but since you've done the work… 😄

pshirshov:
> I'm not sure I feel very strongly that this is necessary (in general JSON literals should be very short), but since you've done the work…

In my case it happened with `{}` literal. One of transitive dependencies has broken my classpath and verbose error message would save me ~15 minutes I guess...

Issue #907, opened 2018-06-13
Accumulation of error messages from validation of multiple fields


JakubKahovec:
Hi,
is it possible to validate multiple fields and have all error messages accumulated and returned in ValidatedNel ?
Thank you.

JakubKahovec:
Hi Felix,
thank you for your answer and the example. What I meant though was accumulating results of all explicitly validated fields as when I specify multiple validations on a decoder I only get the first result of a failed validation.
I've customised your example to show you what I mean - https://scalafiddle.io/sf/dbl631F/0

felixbr:
I didn't even know about `.validate` :D

The method definition is the following:
```
  /**
   * Build a new instance that fails if the condition does not hold.
   */
  final def validate(pred: HCursor => Boolean, message: => String): Decoder[A] = new Decoder[A] {
    final def apply(c: HCursor): Decoder.Result[A] =
      if (pred(c)) self(c) else Left(DecodingFailure(message, c.history))

    override def decodeAccumulating(c: HCursor): AccumulatingDecoder.Result[A] =
      if (pred(c)) self.decodeAccumulating(c) else Validated.invalidNel(DecodingFailure(message, c.history))
  }
```
As you can see the `decodeAccumulating` part is returning an error value if the predicate is `false` and doesn't evaluate the underlying errors. This seems like a bug to me as we typically expect to evaluate all predicates and get back all the errors.

I tried to patch it using ´.ap` (check `cats.Apply` for more info) to include the underlying validations as well and it seems to work fine: https://scalafiddle.io/sf/dbl631F/5

You can use this extension method to fix the behavior in your code right now, but we should probably open a PR to fix this in circe itself. :)

saint1991:
I had the same question but I got the answer from the creator of Circe in #922.

JakubKahovec:
Thank you @felixbr for the patch. It'd be great if this functionality is part of the Circe codebase.

JakubKahovec:
Thanks @saint1991 for the example, it's useful for simple case classes, it just might turn a bit tedious when there are lot fields in a case class and you have to specify decoders for each field manually if you want to validate just few fields.  
I've read the discussion but I'd still prefer to accumulate all errors when calling decodeAccumulating ,regardless of using monadic or applicative decoder, as what is accumulating meant for. For fail-fast behaviour I'd use just decode method.

JakubKahovec:
[...]

PR #931, opened 2018-08-02
Adding ensure method

saint1991:
This PR adds `ensure` method to the `Decoder`.
The motivation here is to provide composable error checking mechanism in both fail-fast and accumulating mode as discussed in #907.

A usage example is below.

```
  import io.circe.parser._

  case class Foo(a: String, b: List[String])

  implicit val decoder = deriveDecoder[Foo]
    .ensure(_.a.nonEmpty, "a should not be empty")
    .ensure(_.b.nonEmpty, "b should not be an empty list")


  val json =
    """{
      |  "a": "",
      |  "b": []
      |}
    """.stripMargin

  // Left(DecodingFailure(a should not be empty, List()))
  decode(json)

  // Invalid(NonEmptyList(DecodingFailure(a should not be empty, List()), DecodingFailure(b should not be an empty list, List())))
  decodeAccumulating(json) 
```

BTW, I felt that `validation` method has a bit inconsistent semantics because it is neither fail-fast nor accumulating. The final validation call overwrites all other calls. 
Is it the intended behaviour?

If so, its signature is misleading...
It's only my feeling but returning Decoder implies it is composable by method chain like the above example.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -121,6 +121,23 @@ trait Decoder[A] extends Serializable { self =>
   }
 
   /**
+    * Build a new instance that fails if the condition does not hold.
+    */
+  final def ensure(pred: A => Boolean, message: => String): Decoder[A] = new Decoder[A] {
+    final def apply(c: HCursor): Decoder.Result[A] = self(c) match {
+      case l @ Left(_) => l
+      case Right(a) => if (pred(a)) Right(a) else Left(DecodingFailure(message, c.history))
+    }
+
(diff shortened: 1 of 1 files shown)


saint1991:
I'm going to add tests against `ensure` after the discussion

travisbrown:
Thanks @saint1991! Will get feedback posted asap.

saint1991:
Sorry I misunderstood, `Decoder` is a type class, so available choices for returning type of `validate` are only `Decoder` or `Unit`.
The former is a bit misleading as discussed in #907 but another one is not functional...

travisbrown:
Hi @saint1991! Sorry for the long delay, but I think this would be a good addition for 0.10.0. If you happen to have tests around I'd definitely appreciate a follow-up PR, but otherwise I can add some later today.

saint1991:
@travisbrown No problem. Thank you for merging this!
I'll add tests for this in a following PR.

PR #942, opened 2018-08-13
Adding adopter ITV

dandxy89:
Adding [ITV](https://www.itv.com/) as an adopter of Circe.


dandxy89:
@davesmith00000

travisbrown:
Okay, works just fine after clearing the Travis CI build cache.

PR #945, opened 2018-08-19
2.13.0!

travisbrown:


lloydmeta:
Does this work with ScalaJS? The reason why I'm asking is because it looks like the references to `java.time._` classes [seems to have broken it](https://travis-ci.org/lloydmeta/enumeratum/jobs/553829396#L1286-L1289) when someone[ tried to re-enable the 2.13 integration that Enumeratum has with Circe](https://github.com/lloydmeta/enumeratum/pull/241):

```
[info] Full optimizing /home/travis/build/lloydmeta/enumeratum/enumeratum-circe/.js/target/scala-2.13/enumeratum-circe-test-opt.js
[error] Referring to non-existent class java.time.Year
[error]   called from io.circe.Encoder$$anon$58.apply(java.lang.Object)io.circe.Json
[error]   called from io.circe.syntax.package$EncoderOps$.asJson$extension(java.lang.Object,io.circe.Encoder)io.circe.Json
```

travisbrown:
@lloydmeta See https://github.com/circe/circe/releases/tag/v0.12.0-M4 and the M3 notes, where this is announced and a workaround is given.

lloydmeta:
Thanks @travisbrown 

> Because I kind of hate Scala.js and refuse to let it make more of a mess of my projects than it already has and also I'm stubborn.

The feels :D

PR #948, opened 2018-08-23
fix broken links on known-issues.html

stsatlantis:
fix broken links on https://circe.github.io/circe/codecs/known-issues.html


stsatlantis:
In line 89 (https://github.com/circe/circe/pull/948/files#diff-cf030dce207351719f3272e77becf040R89) The number displayed is not matching with the number presented in the link. How to fix that?

travisbrown:
Oh, good catch. I think what we want to do is add `{% include references.md %}` to the bottom of each of these files—it looks like somehow it got dropped when these were split out into the `codecs` directory.

travisbrown:
Thanks, @stsatlantis—I'll look into getting this fixed with actual references, but this looks fine for now.

PR #950, opened 2018-08-29
Added encoders/decoders for NonEmptySet and NonEmptyMap (from cats.data)

dcastro:


travisbrown:
Looks good, thanks @dcastro!

PR #964, opened 2018-09-09
Add encoder and decoder for Chain and NonEmptyChain

dcastro:
Added encoders/decoders for the data types introduced in cats 1.3, `Chain` and `NonEmptyChain`.


dcastro:
Code coverage seems to have gone down a bit for `JsonNumber.scala`, but I'm not sure how it's related to the changes I made. Could this be a false positive caused by some property-based test that just happened to not cover a piece of code this one time?

travisbrown:
@dcastro Thanks again for this! And yes, we see that kind of drop due to property-based testing occasionally. I'll get this merged before 0.10.0 (probably this weekend, finally).

dcastro:
@travisbrown thanks, my pleasure! :)

PR #968, opened 2018-09-13
Add MatchesFashion and ClearScore to the adopters

joan38:


travisbrown:
Nice, thanks @joan38!

PR #978, opened 2018-09-28
Add FiniteDuration codec

### LINKED ISSUE #977: Instances for scala.concurrent.Duration/FiniteDuration
It would be nice to be able to encode and decode Durations with Circe, was it a conscious design choice to not include these, or would you accept a PR to add these? :)

LukaJCB:
Should fix #977


travisbrown:
Thanks @LukaJCB!

Issue #965, opened 2018-09-10
Broken benchmark link in documentation


vhiairrassary:
The link `My initial benchmarks` in https://circe.github.io/circe/performance.html points to nothing in GitHub (404).

PR #982, opened 2018-10-08
Fix broken Benchmark link

exoego:
Fixes https://github.com/circe/circe/issues/965

I used [**Permalink to specific commit**](https://help.github.com/articles/getting-permanent-links-to-files/#press-y-to-permalink-to-a-file-in-a-specific-commit
) in case file is gone or structure is changed.


travisbrown:
:+1:, thanks!

Issue #585, opened 2017-03-10
Exclude Options with None


darkfrog26:
I'm generating JSON for working with ArangoDB and it distinguishes between `null` and not set.  This creates a problem since the default operation of Circe semi-auto encoding of `Option` set to `None` is `null`.  Is there some way to configure this?  I would also argue that exclusion should be the default operation anyway since it would result in more compact JSON.

jonas:
This looks like a duplicate of #584. Here's an example of how to omit `None` fields:

```
scala> :paste
// Entering paste mode (ctrl-D to finish)

import io.circe.generic.auto._
import io.circe.Printer
import io.circe.syntax._

case class C(a: Option[String], b: Option[Int], c: Boolean)

val printer = Printer.noSpaces.copy(dropNullKeys = true)


// Exiting paste mode, now interpreting.

import io.circe.generic.auto._
import io.circe.Printer
import io.circe.syntax._
defined class C
printer: io.circe.Printer = Printer(true,true,,,,,,,,,,,,,,,,)

scala> printer.pretty(C(Some("a"), None, false).asJson)
res0: String = {"a":"a","c":false}

scala> printer.pretty(C(None, None, false).asJson)
res1: String = {"c":false}
```

ksilin:
`dropNullKeys` has been changed to the more fitting `dropNullValues`: 

`val printer = Printer.noSpaces.copy(dropNullValues = true)`

hairylime:
Not sure if this is what the OP meant, but it sounds like they are asking for a selective solution - I have something similar with a Tri valued object (think Option, with 2 'None' values, None meaning 'unset'/'set to null', and 'Do not care' meaning ignore that field) - I can just map the 'Some'-ish and the None-ish case to Option, but can't see a way to drop the Json for the Do not care case...

travisbrown:
Now that we have `dropNullValues` on both the printer and `Json` I think we can close this.

Issue #584, opened 2017-03-07
Encoder should remove None: Option[T] values


mikaelstaldal:
It would be nice if the Encoder would remove the key/value completely when the type is `Option[T]` and the value is `None`. Currently, it encodes None to JSON null.

(Json4s works this way.)

This code:
```
  case class Test(foo: String, bar: Option[String])
  implicit val testEncoder: Encoder[Test] = deriveEncoder
  println(Test("FOO", None).asJson.noSpaces)
```
currently prints:
```
{"foo":"FOO","bar":null}
```

I would like it to be:
```
{"foo":"FOO"}
```

vpavkin:
@mikaelstaldal you should just use a custom `Printer` with `dropNullKeys = true`.
By default it's `true`.

mikaelstaldal:
Ah, that works. But it seems like `dropNullKeys` by default is `false`?

This is also something that is missing from documentation.

vpavkin:
yes, it's false :) I made a mistake and fixed the comment afterwards

ckozak:
What was the reason to make `dropNullKeys = false` by default? It seems that not putting nulls is more common among other libraries.

bergmark:
I commented on this elsewhere, so a quick repeat:

My take is that the printer should not modify the actual JSON format. This is something each `Encoder` should decide for itself. Inevitably there will be a case where you want to combine the two options for different types, AFAICT that's not possible at the moment.

Mahoney:
Could nested options be interpreted differently? For
```case class Foo(bar: Option[Option[String])```
```Foo(None)``` = ```{}```
```Foo(Some(None)``` = ```{"bar": null}```
```Foo(Some("actual data")``` = ```{"bar": "actual data"}```

naderghanbari:
I agree that the `Printer` should not touch the JSON by default but I also agree that the `Printer` should have the power to make the result nicer if you want to. This could help in some cases where the actual value might be really `null` (Java interop). So in other words I think the `Printer` is fine as is.

I think the encoder, though, should drop `None` values by default and not encode them as `null`. It should, however, encode genuine `null` values as `null` and not drop them. This is probably what you'd expect if you are a Scala developers (or a functional programmer in general). 

Similar to `dropNullKeys = false` I think the encoder configuration can have something like `dropNone = true`.

travisbrown:
Now that we have dropNullValues on both the printer and Json I think we can close this.

PR #983, opened 2018-10-09
Added `Json.dropNullValues`.

orium:


orium:
Not sure if this is "too specific" to be added to `Json`, but it is something that I often use and can be useful for others (see #584 and #585).

Maybe it would be nicer to make it `def dropNullValues(deep: Boolean = false): Json`?

poslegm:
It seems very useful for me! Any chance that it will be merged?

travisbrown:
I've been wanting to avoid this and leave the presence or absence of null values as a printing-level issue, but there seems to be enough demand that I guess I should give that up. I'll merge this as-is, and if someone wants to add the deep version we can review it when it comes up.

realvictorprm:
funny thing, this is just something I wanted to use and then I noticed that this was merged a day ago :smile: nice stuff :smiley:

PR #985, opened 2018-10-11
Get maven version for badge from Scala 2.12

aaabramov:


travisbrown:
:+1:, makes sense, thanks!

PR #986, opened 2018-10-12
Extend ':=' operator any type with a KeyEncoder instance

tejinders:
In the cases where JSON keys have a richer type
than plain strings, it is handy to have the ':='
operator work seamlessly. As circe already has
'KeyEncoder' abstraction, the assignment operator
can work off the same abstraction. This is also
backwards compatible with the 'String' based
implementation.


travisbrown:
@tejinders This is great, thanks (especially for the tests)!

PR #992, opened 2018-10-19
Remove redundant collection conversion

exoego:
Found when reading code base on Intellij IDEA.
It seems that there are no redundant conversions but this.


travisbrown:
Good catch!

PR #994, opened 2018-10-20
[generic-extra] Decode default value on Option if key is missing.

### LINKED ISSUE #878: [generic extras] Default value of Option type will be ignored
Hi, The following code returns `Right(Foo(a, None))`　
but, I expect `Right(Foo(a, Some(0)))`

Is this a bug, or am I doing something wrong?

```
import io.circe.generic.extras.auto._
import io.circe.generic.extras.Configuration

implicit val customConfig: Configuration = Configuration.default.withDefaults

case class Foo(a: String = "a", b: Option[Int] = Some(0))
io.circe.parser.decode[Foo]("{}")
```

My dependencies.
```
"io.circe" %% "circe-core" % "0.9.3"
"io.circe" %% "circe-generic" % "0.9.3"
"io.circe" %% "circe-parser" % "0.9.3"
"io.circe" %% "circe-generic-extras" % "0.9.3"
```

Thanks.

exoego:
Closes #878 

| |  without default <br>`Foo(a: Option[Int])` | with default <br>`Foo(a: Option[Int]=Some(0))` |
|---|---|---|
| **null given <br> `{"a":null}`** | `Foo(None)` | `Foo(None)` |
| **key missing <br> `{}`** | `Foo(None)` | right now: `Foo(None)`<br>This PR: `Foo(Some(0))` |

### DIFF
--- a/modules/generic-extras/src/main/scala/io/circe/generic/extras/decoding/ReprDecoder.scala
+++ b/modules/generic-extras/src/main/scala/io/circe/generic/extras/decoding/ReprDecoder.scala
@@ -34,11 +34,11 @@ abstract class ReprDecoder[A] extends Decoder[A] {
     name: String,
     defaults: Map[String, Any]
   ): Decoder.Result[B] = result match {
-    case r @ Right(_) => r
-    case l @ Left(_) =>
+    case r @ Right(_) if r ne Decoder.keyMissingNone => r
+    case _ =>
       defaults.get(name) match {
         case Some(d: B @unchecked) => Right(d)
-        case _                     => l
+        case _                     => result
       }
   }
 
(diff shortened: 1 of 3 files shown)


modules/generic-extras/src/main/scala/io/circe/generic/extras/decoding/ReprDecoder.scala
```diff
@@ -34,10 +34,10 @@ abstract class ReprDecoder[A] extends Decoder[A] {
     name: String,
     defaults: Map[String, Any]
   ): Decoder.Result[B] = result match {
-    case r @ Right(_) => r
-    case l @ Left(_) => defaults.get(name) match {
+    case r @ Right(_) if r ne Decoder.keyMissingNone => r
```
travisbrown: Why not just `case r @ Right(v) if v.isEmpty => r`?

exoego: @travisbrown 
Can I have your moment to get review?
Perhaps, you may missed my response due to last GitHub outage. 
Thanks in advance 😃

travisbrown:
Thanks much, @exoego—sorry for the long delay in review, but this looks reasonable to me, and will be released in 0.11.0.

niij:
@exoego Thank you for this!

Issue #976, opened 2018-09-26
JavaTimeDecoders/JavaTimeEncoders are package-private


adamw:
After updating to 0.10.0, there's no longer a `TimeInstances` trait (in `circe-java8`). This seems to be replaced by `JavaTimeDecoders` and `JavaTimeEncoders`. However, they are now package-private.

I used to have a single `Json` object which extended `AutoDerivation`, `TimeInstaces` and provided some custom encoders for my application-specific types. That way, all I needed to do to use json was a single `import Json._`.

Maybe it would be possible to make those non-package-private again?

travisbrown:
Hmm, yes, will get a fix together—thanks.

adamw:
Awesome, thanks! :)

adamw:
If it would make things easier I can prepare a (rather simple) PR that would change the visibility, unless you have some other design in mind? (I could attempt to work on this as well :) )

travisbrown:
@adamw I'd been thinking of splitting circe-java8 into its own repo but hadn't gotten around to it yet, and now I'm not sure that's the best idea. If you want to open a PR I'd appreciate it, but otherwise will get to it this week.

adamw:
I did a super-simple PR (https://github.com/circe/circe/pull/995), let me know if it's enough :)

travisbrown:
Fixed in #995.

PR #995, opened 2018-10-23
Make JavaTimeEncoders/Decoders public

adamw:
So that they can be extended by custom helper objects.

See #976


travisbrown:
Thanks!

PR #980, opened 2018-10-05
Close #979 by adding encoders and decoders for MonthDay, Year, and ZoneOffset

plokhotnyuk:
...and improve throughput of parsing and serialization of YearMonth values, by using build-in parser and formatter

PR #1000, opened 2018-10-30
Fix some issues in java8 module

travisbrown:
…introduced in #980. This is my fault—I shouldn't have thought it would be okay to run the circe-java8 tests manually.


plokhotnyuk:
@travisbrown Thank you a lot! It was my fault too :)

PR #1003, opened 2018-11-05
Update README to include shaclex

labra:
Update README to include [shaclex](http://labra.weso.es/shaclex/), an RDF validation library that uses circe for JSON encoding and decoding.


travisbrown:
Thanks!

PR #1004, opened 2018-11-05
Update README to add reference at the end of README

labra:
Added forgotten link at the end of README


travisbrown:
Ah, even better—good catch.

PR #1013, opened 2018-11-09
Revert "Add FiniteDuration codec"

travisbrown:
This reverts commit e73328882f38dd74483f31a6032a4011a5a56f7e from #978. I've been thinking about this some more in preparation for 0.11, @LukaJCB, and I think this is just too far from our general principle of "don't do anything that can be done more than one way". We could consider a `io.circe.durationCodecs` object (on the model of `io.circe.disjunctionCodecs`)—would you find that useful?


LukaJCB:
Sounds useful to me, thanks a lot @travisbrown! :)

guizmaii:
Hi @travisbrown,

What is the current state of the `FiniteDuration` Encoder/Decoder in circe? 


Thanks,
Jules

Issue #1019, opened 2018-11-22
Allow generating validation messages from the HCursor


pfcoperez:
`Decoder[T]` offers a nice way of generating a new decoder which fails when a predicate over the cursor is `true`.

https://github.com/circe/circe/blob/9e3a589d19f0a222d65ea739d0036af3b44f035c/modules/core/shared/src/main/scala/io/circe/Decoder.scala#L166-L172

It currently receives the predicate and a call-by-name message for the potentially generated failures.

I find it useful to be able to generate the error message from the `HCursor` evaluated by the predicated. For example, to implement strict de-serialization and feed the error with hints about which fields are un-expected.

I've already worked on this and have a PR ready.

PR #1020, opened 2018-11-22
`Decoder#validate` allowing building the message from the trigger `HCursor`

pfcoperez:
Resolves https://github.com/circe/circe/issues/1019

This PR pretends to enable generating `validate` messages from the `HCursor` value which triggered the failure by satisfying the passed predicate.

It overloads the original `validate` method which now gets implemented by invoking this new implementation.


travisbrown:
This looks reasonable to me, thanks! We'll be sure it makes it into 0.11.

pfcoperez:
@travisbrown As discussed in https://github.com/circe/circe/pull/1025#issuecomment-444067444 , I've updated this PR so the new `validate` method receives a `HCursor => List[String]` function to represent both the predicate and the errors generated by falsifying it.

travisbrown:
Thanks, looks good to me!

PR #1025, opened 2018-12-02
Fix ensure

travisbrown:
In the [0.10.0] release we introduced a new `Decoder#ensure` method that was semantically kind of weird and included some behavior that (when used together with error accumulation) was definitely buggy:

```scala
scala> import io.circe.{ Decoder, Json }
import io.circe.{Decoder, Json}

scala> val decodePos: Decoder[Int] = Decoder[Int].ensure(_ > 0, "not positive")
decodePos: io.circe.Decoder[Int] = io.circe.Decoder$$anon$25@5b9070e3

scala> decodePos.accumulating(Json.Null.hcursor)
res0: io.circe.AccumulatingDecoder.Result[Int] = Invalid(NonEmptyList(DecodingFailure(Int, List()), DecodingFailure(not positive, List())))
```

Thanks to @ylaurent for noticing and [reporting this](https://gitter.im/circe/circe?at=5bfdf97dcfa682348d961433).

The tests would have caught this bug, but the relevant test was itself buggy, containing a check for oddness that said e.g. `-1` was not odd.

I've changed the documentation for the previous version of `ensure` to state explicitly that `dec.ensure(p1, "p1").ensure(p2, "p2")` will only provide the first failure, and have added a new, more powerful `ensure` method that allows the user to provide multiple errors that depend on the decoded value.

@pfcoperez, I was thinking we might be able to add a similar `HCursor => List[String]` variant for `validate` that would accommodate the use case supported by your #1020, and that would keep these two methods consistent. What do you think?

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -138,16 +138,13 @@ trait Decoder[A] extends Serializable { self =>
    */
   final def ensure(pred: A => Boolean, message: => String): Decoder[A] = new Decoder[A] {
     final def apply(c: HCursor): Decoder.Result[A] = self(c) match {
-      case l @ Left(_) => l
-      case Right(a)    => if (pred(a)) Right(a) else Left(DecodingFailure(message, c.history))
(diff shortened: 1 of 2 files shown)


travisbrown:
Thanks, @pfcoperez!

PR #1032, opened 2018-12-13
Make KeyDecoder / KeyEncoder as trait 

andwxh:
Make KeyDecoder / KeyEncoder as trait instead of abstract class so they can be extend together.


travisbrown:
I've tended to default to abstract class in the past (in part because of the mess traits make of Java interop before 2.12), but I think it's reasonable for these type classes to be traits, if only for consistency with the non-key `Decoder` and `Encoder`. Thanks.

andwxh:
Ah... I was forgot that trait with default method implementations cannot be used in Java before Scala2.12. Thank you for generously merge this pull request.

PR #1036, opened 2018-12-18
Remove AnyVal constraint on *codeUnwrapped

### LINKED ISSUE #1023: Drop restriction to AnyVal for deriveUnwrappedEncoder and deriveUnwrappedDecoder
I would like to use `deriveUnwrappedEncoder` and `deriveUnwrappedDecoder` for a case class that contains a single field, but is not a `AnyVal`:
```scala
sealed trait FilterParameter
case class And(filters: Seq[FilterParameter]) extends FilterParameter
case class Or(filters: Seq[FilterParameter]) extends FilterParameter
case class MapIncludes(key: String, value: String) extends FilterParameter
```
In the JSON representation I would like to have something like this, skipping the wrapper for the `filters` key:
```javascript
{
  "And": [
    "MapIncludes": {
      "key": "foo",
      "value": "bar"
    },
    "MapIncludes": {
      "key": "one",
      "value": "two"
    }
  ]
}
```
`deriveUnwrappedEncoder` and `deriveUnwrappedDecoder` would work nicely for this use-case, except for the fact that they are restricted to `A <: AnyVal`.
Is this restriction necessary or was that just the primary use-case that came to mind and in principle this could be applied to any `Product` that only has a single field?

travisbrown:
Fixes #1023.

### DIFF
--- a/modules/generic-extras/src/main/scala/io/circe/generic/extras/decoding/UnwrappedDecoder.scala
+++ b/modules/generic-extras/src/main/scala/io/circe/generic/extras/decoding/UnwrappedDecoder.scala
@@ -6,7 +6,7 @@ import shapeless.{ ::, Generic, HNil, Lazy }
 abstract class UnwrappedDecoder[A] extends Decoder[A]
 
 final object UnwrappedDecoder {
-  implicit def decodeUnwrapped[A <: AnyVal, R](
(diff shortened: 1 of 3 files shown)


nightscape:
Cool, thanks! I had the tab with the issue open all the time as a reminder, but didn't get around to do it :wink:

PR #1039, opened 2018-12-19
Include VEACT as adopter

ArturSoler:
As suggested, including VEACT as a new organisation using circe.

PS: Thanks for all, circe is great! :)


travisbrown:
Thanks much, @ArturSoler!

(The build failure is because of a Ruby change on Travis CI, and should be fixed after a rebase, but I'm just going to merge as-is, since this change is obviously fine.)

PR #1048, opened 2019-01-05
Option to print JSON objects with sorted keys

koterpillar:
This fixes #1045 by adding:

* `Printer.spaces2SortKeys`, `Printer.spaces4SortKeys` and `Printer.noSpacesSortKeys` to print JSON objects with keys sorted alphabetically
* `withSortedKeys` modifier on any `Printer`

Appreciate feedback around the implementation and tests - I couldn't find a lot of examples to go on from.

PR #1050, opened 2019-01-07
Remove unused preserveOrder

koterpillar:
Extracted this out of #1048 because this will break source compatibility.


travisbrown:
Not sure what the intent is here? #1048 is merged and I'm not too worried about breaking source compatibility in non-standard usage in 0.12.0, so I think we can close this. Please reopen if I've misunderstood.

koterpillar:
Sorry I wasn't clear in the description. This PR removes a completely unused parameter called `preserveOrder`. It doesn't do anything, and now with #1048 merged, just creates confusion. So I'd like for this to go in still.

Unfortunately, I can't reopen the PR myself.

travisbrown:
@koterpillar Got it, thanks!

PR #1053, opened 2019-01-09
Added a visualisation of code evolution

gitviscode:
If you like it, feel free to merge it :)


travisbrown:
I'm thinking this is spam? In any case the video is no longer available and I don't think this belongs on the README anyway.

Issue #1040, opened 2018-12-20
Denial of service when parsing too long JSON numbers as double or float


plokhotnyuk:
Sub-quadratic decreasing of throughput when length of the JSON number to parse is increasing.

On contemporary CPUs parsing of such JSON numbers that are bound on doubles or floats and has 1000000 decimal digits (~1Mb) can took more than 14 seconds. Below are results of the benchmark where the size parameter is a number of digits to parse:
```
[info] REMEMBER: The numbers below are just data. To gain reusable insights, you need to follow up on
[info] why the numbers are the way they are. Use profilers (see -prof, -lprof), design factorial
[info] experiments, perform baseline and negative tests that provide experimental control, make sure
[info] the benchmarking environment is safe on JVM/OS/HW level, ask for reviews from the domain experts.
[info] Do not assume the numbers tell you what you want them to tell.
[info] Benchmark                              (size)   Mode  Cnt         Score         Error  Units
[info] DoubleBenchmark.readCirce                   1  thrpt    5  13406952.737 ± 2032432.395  ops/s
[info] DoubleBenchmark.readCirce                  10  thrpt    5  10997675.235 ±  619560.310  ops/s
[info] DoubleBenchmark.readCirce                 100  thrpt    5    351697.951 ±    4377.854  ops/s
[info] DoubleBenchmark.readCirce                1000  thrpt    5     19590.316 ±    1601.505  ops/s
[info] DoubleBenchmark.readCirce               10000  thrpt    5       482.626 ±       2.149  ops/s
[info] DoubleBenchmark.readCirce              100000  thrpt    5         6.579 ±       0.629  ops/s
[info] DoubleBenchmark.readCirce             1000000  thrpt    5         0.071 ±       0.005  ops/s
```
### Reproducible Test Case

To run that benchmarks on your JDK:
1. [Install latest version of `sbt`](https://www.scala-sbt.org/1.0/docs/Setup.html) and/or ensure that it already installed properly: 
```
sbt about
```

2. Clone `jsoniter-scala` repo: 
```
git clone https://github.com/plokhotnyuk/jsoniter-scala.git
```

3. Enter to the cloned directory and checkout for the specific branch:
```
cd jsoniter-scala
git checkout circe-DoS-for-doubles-and-floats
```

4. Run benchmarks using a path parameter to your JDK:
```
[...]

PR #1054, opened 2019-01-09
Avoid BiggerDecimal construction for some decoding to Double

travisbrown:
This is a simple fix for some instances of #1040 (I've got a bigger overhaul on the way). Specifically, things like `io.circe.parser.decode[Double]("9" * 10000000)` now return almost instantaneously.


plokhotnyuk:
What about `NumberFormatException` in case of invalid input? Should not be them catched and returned as `NaN`?

travisbrown:
@plokhotnyuk From the docs for `JsonDecimal`: "Represent a valid JSON number as a `String`." The `value` should always be a validated JSON number (the only way it wouldn't is through invalid use of the `JsonNumber.fromDecimalStringUnsafe` method), and `Double.parseDouble` [accepts a superset](https://docs.oracle.com/javase/7/docs/api/java/lang/Double.html#valueOf(java.lang.String)) of all valid JSON number strings.

PR #1056, opened 2019-01-10
Avoid BiggerDecimal construction for some decoding to Long

travisbrown:
Similar to #1055—lets us avoid constructing a `BiggerDecimal` for more instances of `Long` decoding, so that e.g. `io.circe.jawn.decode[Long]("9" * 1000000)` fails immediately.


travisbrown:
Oh, right, of course this doesn't work.

PR #1061, opened 2019-01-15
Fix #1060

### LINKED ISSUE #1060: Bring 2.12+ core time instances into sync with circe-java8
Ugh, these are already out of sync (`MonthDay`, etc. are only in circe-java8).

Managing duplicate code like this in order to support a Java version that's been EOL for almost four years is the kind of ridiculous bullshit that makes people (me) hate open source project maintenance. I'm inclined to drop Java 7 support in the next release.

travisbrown:
Fixes #1060. I've switched both to the standard `uuuu-MM` for `YearMonth` for consistency. Now the diff looks reasonable:

```scala
travis@sidmouth circe(fix/time-instance-discrepancies)$ diff modules/java8/src/main/scala/io/circe/java8/time/JavaTimeDecoders.scala modules/core/shared/src/main/scala-with-jdk8/io/circe/JavaTimeDecoders.scala 
1c1
< package io.circe.java8.time
---
> package io.circe
3d2
< import io.circe.{ Decoder, DecodingFailure, HCursor, Json }
31c30
< private[time] abstract class JavaTimeDecoder[A](name: String) extends Decoder[A] {
---
> private[circe] abstract class JavaTimeDecoder[A](name: String) extends Decoder[A] {
56c55
< private[time] abstract class StandardJavaTimeDecoder[A](name: String) extends JavaTimeDecoder[A](name) {
---
> private[circe] abstract class StandardJavaTimeDecoder[A](name: String) extends JavaTimeDecoder[A](name) {
61c60
< trait JavaTimeDecoders {
---
> private[circe] trait JavaTimeDecoders {
travis@sidmouth circe(fix/time-instance-discrepancies)$ diff modules/java8/src/main/scala/io/circe/java8/time/JavaTimeEncoders.scala modules/core/shared/src/main/scala-with-jdk8/io/circe/JavaTimeEncoders.scala 
1c1
< package io.circe.java8.time
---
> package io.circe
3d2
< import io.circe.{ Encoder, Json }
31c30
< private[time] abstract class JavaTimeEncoder[A <: TemporalAccessor] extends Encoder[A] {
---
> private[circe] abstract class JavaTimeEncoder[A <: TemporalAccessor] extends Encoder[A] {
37c36
< trait JavaTimeEncoders {
---
> private[circe] trait JavaTimeEncoders {
```
Unfortunately these changes break binary compatibility, so they'll have to wait for 0.12.0.

### DIFF
--- a/modules/core/shared/src/main/scala-with-jdk8/io/circe/JavaTimeDecoders.scala
+++ b/modules/core/shared/src/main/scala-with-jdk8/io/circe/JavaTimeDecoders.scala
@@ -14,8 +14,8 @@ import java.time.{
   Year,
   YearMonth,
   ZonedDateTime,
-  ZoneOffset,
-  ZoneId
+  ZoneId,
+  ZoneOffset
 }
 import java.time.format.DateTimeFormatter
 import java.time.format.DateTimeFormatter.{
--- a/modules/core/shared/src/main/scala-with-jdk8/io/circe/JavaTimeDecoders.scala
+++ b/modules/core/shared/src/main/scala-with-jdk8/io/circe/JavaTimeDecoders.scala
@@ -223,11 +223,8 @@ private[circe] trait JavaTimeDecoders {
    */
   implicit final val decodeMonthDay: Decoder[MonthDay] =
     new StandardJavaTimeDecoder[MonthDay]("MonthDay") {
-      private[this] final def monthDayFormatter: DateTimeFormatter =
-        DateTimeFormatter.ofPattern("--MM-dd")
-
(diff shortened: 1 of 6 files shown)


plokhotnyuk:
For default encoders and decoders of `java.time.YearMonth` and `java.time.MonthDay` which does not require user's formatter it better to use `toString` and `parse` methods of these classes: they works faster and match with user expectations in format.

travisbrown:
@plokhotnyuk Fair enough—I've updated.

travisbrown:
(…although in both cases the formatters I was using were the defaults used with `parse` and `toString`.)

PR #1081, opened 2019-02-27
More helpful message on missing field.

### LINKED ISSUE #588: Allow customization of the error message for missing element when decoding
Currently decoding a JSON that misses an element will result in a `DecodingFailure` with a message saying: `"Attempt to decode value on failed cursor"`.
This error cannot be customized other than by doing string matching on `DecodingFailure.message`.
I would suggest to add a default parameter on `decode` and `decodeAccumulating`:
```scala
final def decode[A: Decoder](input: String, missingElementErrorMessage: String =
                             defaultMissingElementErrorMessage)
``` 

An alternative would be to encode that case in the type:
```scala
sealed trait DecodingFailure { def history: List[CursorOp] }
case class MissingElement(val history: List[CursorOp]) extends DecodingFailure
case class IncorrectValue( /* ... */ ) extends DecodingFailure
```

exoego:
Closes #588.

Honestly, I am unfamiliar with internal of circe and how to use cats.
The changes are mix of suggested changes in https://github.com/circe/circe/issues/588#issue-214064936 and https://github.com/circe/circe/issues/588#issuecomment-286423496.

I believe addressing the long-unadressed issue improves developer experience of circe, so opened up this PR.
I am happy to update PRs based on feedbacks and guidance.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Error.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Error.scala
@@ -84,6 +84,10 @@ final object DecodingFailure {
       DecodingFailure(sw.toString, ops)
   }
 
+  case class MissingElement(field: String, history: List[CursorOp])
+      extends DecodingFailure(s"The field $field is missing")
+  case class IncorrectValue(field: String /* ... */, history: List[CursorOp]) extends DecodingFailure("Incorrect value")
+
   implicit final val eqDecodingFailure: Eq[DecodingFailure] = Eq.instance {
     case (DecodingFailure(m1, h1), DecodingFailure(m2, h2)) => m1 == m2 && CursorOp.eqCursorOpList.eqv(h1, h2)
   }
--- a/modules/tests/shared/src/test/scala/io/circe/DecoderSuite.scala
+++ b/modules/tests/shared/src/test/scala/io/circe/DecoderSuite.scala
@@ -570,6 +570,20 @@ class DecoderSuite extends CirceSuite with LargeNumberDecoderTests with TableDri
     assert(statefulOpt.decodeJson(json).swap.exists(_.message === "The field b is missing"))
   }
 
+  it should "fail when a nested field is missing" in {
+    case class Foo(x: Int, bar: Bar)
+    case class Bar(nested: String)
+    val decoder: Decoder[Foo] = new Decoder[Foo] {
+      override def apply(c: HCursor): Decoder.Result[Foo] =
+        for {
+          x <- c.downField("x").as[Int]
+          a <- c.downField("bar").downField("nested").as[String]
+        } yield Foo(x, Bar(a))
+    }
+    val Right(fooJson) = parse("""{"x":42, "bar": {}}""")
+    assert(decoder.decodeJson(fooJson).swap.exists(_.message === "The field bar.nested is missing"))
+  }
+
   checkLaws("Codec[WrappedOptionalField]", CodecTests[WrappedOptionalField].codec)
 
   "decodeSet" should "match sequence decoders" in forAll { (xs: List[Int]) =>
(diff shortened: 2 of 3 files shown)


tkroman:
@exoego based on this code, I think you would benefit from `show` on errors:

```scala
@ import cats.implicits._
@ import io.circe.parser._
@ import io.circe.generic.auto._

@ case class A(x: Int, y: String)

@ case class B(a: A, z: Int)

@ decode[A]("""{"x":1}""").left.get.show
res10: String = "DecodingFailure at .y: Attempt to decode value on failed cursor"

@ decode[B]("""{ "z" : 1 }""").left.get.show
res15: String = "DecodingFailure at .a: Attempt to decode value on failed cursor"

@ decode[B]("""{ "z" : 1, "a": { "x": 1 } }""").left.get.show
res16: String = "DecodingFailure at .a.y: Attempt to decode value on failed cursor"
```

exoego:
@tkroman 
Let me confirm your intent.
Do you mean: this PR should modify `show` instead of `tryDecode` ?

tkroman:
I rather meant that this functionality is already covered by `Show` instances on errors, so maybe you just want to use that.

exoego:
> this functionality is already covered by `Show` instances on errors

We could argue that is is covered.
But the point of  the issue #588 is improving error message.
Current error message `DecodingFailure at .x: Attempt to decode value on failed cursor` tell that just something went wrong.
It is better to decipher such message into more user-friendly message like `Field x is missing`.

exoego:
Could I get review on this?

exoego:
Closing this since no interests in this PR.

exoego:
Reopened since #588 is still open for 2 months.

travisbrown:
Sorry for the extremely long delay on review here. I don't think this will make it into 0.12.0, but I'm going to have more time for circe maintenance over the next few weeks, and this will be the basis of a reworking of errors for 0.13.0.

modules/core/shared/src/main/scala/io/circe/Error.scala
```diff
@@ -84,6 +84,10 @@ final object DecodingFailure {
       DecodingFailure(sw.toString, ops)
   }
 
+  case class MissingElement(field: String, history: List[CursorOp])
+      extends DecodingFailure(s"The field $field is missing")
+  case class IncorrectValue(field: String /* ... */, history: List[CursorOp]) extends DecodingFailure("Incorrect value")
```
jeffmay: Why the `/* ... */`?

modules/tests/shared/src/test/scala/io/circe/DecoderSuite.scala
```diff
@@ -561,13 +561,27 @@ class DecoderSuite extends CirceSuite with LargeNumberDecoderTests with TableDri
...
+    case class Bar(nested: String)
+    val decoder: Decoder[Foo] = new Decoder[Foo] {
+      override def apply(c: HCursor): Decoder.Result[Foo] =
+        for {
+          x <- c.downField("x").as[Int]
+          a <- c.downField("bar").downField("nested").as[String]
+        } yield Foo(x, Bar(a))
+    }
+    val Right(fooJson) = parse("""{"x":42, "bar": {}}""")
+    assert(decoder.decodeJson(fooJson).swap.exists(_.message === "The field bar.nested is missing"))
```
jeffmay: I think it would read better if we surrounded the path with `'`s. Also, any reason the path doesn't start with a `.`?

jatcwang:
@exoego thanks the PR. There's this other one https://github.com/circe/circe/pull/1617 which introduce typed errors (including missing field). If you can take a quick look and see if that addresses everything you need then that'll be good :)

exoego:
Closing in favor of #1613

PR #1091, opened 2019-03-07
Added Permutive

Joe8Bit:
Adding Permutive to `README.md`, big fans of Circe!


travisbrown:
Great, thanks @Joe8Bit!

PR #1092, opened 2019-03-07
ISSUE-1088: Adds JsonF pattern functor

Baccata:
Adds a pattern-functor that reflects the Json ADT, as well as
fold/unfold methods.

This can be used in conjunction with recursion schemes libraries.


Baccata:
Force-pushed for scalafix import order.

modules/core/shared/src/main/scala/io/circe/JsonF.scala
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * A pattern-functor reflecting the JSON datatype structure in a
+ * non-recursive way.
+ */
+sealed trait JsonF[+A]
+
+// format: off
+case object JNullF                                  extends JsonF[Nothing]
+case class JBooleanF(b: Boolean)                    extends JsonF[Nothing]
```
kaishh: final everything?

Baccata: Addressed in b6a8a4e

travisbrown:
Thanks, @Baccata! I'm going to go ahead and merge this, and we can decide whether it belongs in core before the 0.12.0 release.

PR #1094, opened 2019-03-10
Move pattern functor to separate module

travisbrown:
@Baccata What do you think? This just moves `JsonF` to a new circe-rs module (for _recursion schemes_, but I'm happy to change the package name). I want to get an 0.12.0 release out as soon as ScalaTest 3.0.7 is here, and we can decide whether this belongs in core after that.


Baccata:
Happy with it :+1:

travisbrown:
Thanks @Baccata!

PR #1096, opened 2019-03-12
More circe-rs stuff

travisbrown:
Some builds have been failing because of the fold-unfold test, and I took a quick look but couldn't reproduce locally and didn't want to wade through the huge `JsonF` values, so I added a `Shrink` instance (and went ahead and threw in a couple of other small changes as well).

### DIFF
--- a/modules/testing/shared/src/main/scala/io/circe/testing/ShrinkInstances.scala
+++ b/modules/testing/shared/src/main/scala/io/circe/testing/ShrinkInstances.scala
@@ -1,6 +1,7 @@
 package io.circe.testing
 
 import io.circe.{ Json, JsonBigDecimal, JsonNumber, JsonObject }
+import io.circe.rs.JsonF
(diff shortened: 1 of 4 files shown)


travisbrown:
(For what it's worth the issue was that the `JsonF` arbitrary instance was producing object values with duplicate keys, which obviously don't round-trip.)

modules/testing/shared/src/main/scala/io/circe/testing/ShrinkInstances.scala
```diff
@@ -58,4 +59,15 @@ private[testing] trait ShrinkInstances {
         }
     )
   )
+
+  implicit def shrinkJsonF[A](implicit A: Shrink[A]): Shrink[JsonF[A]] = Shrink {
```
Baccata: Wow wasn't aware of this, it's great 😄

Issue #797, opened 2017-12-07
Add a method for "canonical" JSON equality


travisbrown:
See #793 for discussion. The idea is that while `{"foo": null}` and `{}` would be represented by `Json` values for which equality doesn't hold (either universal or `Eq`), they could be compared with a new method on `Json` that would indicate that they are canonically the same.

(Relatedly, we may also want to reconsider whether we want `json"""{"a":1,"b":2}""" == json"""{"b":2,"a":1}"""` to remain true—I think we do, but given that they are observably different it's vaguely tempting to change that once we have this new comparison).

PR #1099, opened 2019-03-25
Add canonicallyEqual method to Json

francescoperera:
Created method as per #797 with some tests.
Tests pass but get some errors on the scalaFmtCheck, ran `sbt scalasytle` but got no errors.
I could be missing something very simple but wanted to open the PR regardless to see if anyone could point me out what I am missing/doing wrong.

Any feedback would be very welcomed, thanks.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Json.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Json.scala
@@ -169,6 +169,15 @@ sealed abstract class Json extends Product with Serializable {
     }
 
   /**
+    *Equality method that does not consider field order or null
+    *values.
+    */
+  def canonicallyEqual(that: Any): Boolean = that match {
+    case that : Json => this.dropNullValues.equals(that.dropNullValues)
+    case _ => false
+  }
+
+  /**
    * Drop the entries with a null value if this is an object.
    *
    * Note that this does not apply recursively.
--- a/modules/core/shared/src/main/scala/io/circe/Json.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Json.scala
@@ -169,12 +169,12 @@ sealed abstract class Json extends Product with Serializable {
     }
 
   /**
-    *Equality method that does not consider field order or null
-    *values.
-    */
+   *Equality method that does not consider field order or null
+   *values.
+   */
   def canonicallyEqual(that: Any): Boolean = that match {
-    case that : Json => this.dropNullValues.equals(that.dropNullValues)
-    case _ => false
+    case that: Json => this.dropNullValues.equals(that.dropNullValues)
+    case _          => false
   }
 
   /**
--- a/modules/tests/shared/src/test/scala/io/circe/JsonSuite.scala
+++ b/modules/tests/shared/src/test/scala/io/circe/JsonSuite.scala
@@ -84,6 +84,21 @@ class JsonSuite extends CirceSuite with FloatJsonTests {
 
     assert(merged.asObject.map(_.toList) === Some(fields.reverse))
   }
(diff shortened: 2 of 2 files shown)


francescoperera:
I reused the `dropNullValues` method for this which means that `canonicallyEqual` is not recursive ( since dropNullValues isn't). 
While `{} == {"foo" : null}` is valid , something a bit more involved  like ` {} == {"foo" : { "bar" : null}` or ` {  "foo" : { "baz": "someVal"} == {  "foo" : { "bar" : null, "baz": "someVal"}` will hold ( even though it should IMO).

An alternative path would be to refactor `dropNullValues` to have drop any nested null value if a recursive boolean flag is provided or implement this in a new method.

Any opinions ?

travisbrown:
Thanks @francescoperera! For the Scalafmt errors you should be able to run `sbt scalafmt` to fix everything automatically. I'm traveling at the moment but will take a closer look asap.

darkfrog26:
That's the status of this? I just ran into this situation where I want to compare two JSON objects for equality.

oschrenk:
Any chance of adding this in a new release?

tiphdousset:
I'm also very interested in this feature 🙋 
What is us holding back merging this feature?

armanbilge:
This PR looks stale, so at the very least somebody will have to revive it and merge in the latest main.

tiphdousset:
@armanbilge  I just had a look at the changes @francescoperera is proposing; but I actually don't understand why the "dropNullValues" would cover the problem of ordering? 🤔

armanbilge:
I haven't had a chance to dig into this, but it sounds like it might just need a new PR :)

modules/core/shared/src/main/scala/io/circe/Json.scala
```diff
@@ -168,6 +168,15 @@ sealed abstract class Json extends Product with Serializable {
       case _ => that
     }
 
+  /**
+   *Equality method that does not consider field order or null
```
tiphdousset, line 172: I think the "dropNullValues" has nothing to do with the order of the fields.
In any case (with or without dropNullValues) the equality of JSON does not care about the order.

modules/core/shared/src/main/scala/io/circe/Json.scala
```diff
@@ -168,6 +168,15 @@ sealed abstract class Json extends Product with Serializable {
...
+   *values.
+   */
+  def canonicallyEqual(that: Any): Boolean = that match {
+    case that: Json => this.dropNullValues.equals(that.dropNullValues)
```
tiphdousset, line 176: I think "deepDropNullValues" would make more sense here, WDYT?

modules/core/shared/src/main/scala/io/circe/Json.scala
zarthross: Agreed.  I think the comment is fine either way since its still technically true.  Might be worth adding additional comments to the standard `equals` operator to make it clear.

modules/core/shared/src/main/scala/io/circe/Json.scala
zarthross: Agreed, but at the same time I think we can make a more memory efficient method for doing the equality check instead of building a entirely new json object sans `NULL` to just throw it away.

zarthross (commented):
@tiphdousset I'd be happy to include a `canonicallyEqual` method if you'd like to open an alternative PR.

zarthross:
For now, since this is VERY stale, I'm going to close it.  I'll reopen this if  @francescoperera  requests it, or they can make a new one if they want.

Issue #1024, opened 2018-11-29
Decoding: Configuration.default.withDefaults makes it use the default value when the decoding the field failed


laszlox:
I'm using the Configuration.default.withDefaults feature and lots of my case classes have defaults values for members. However, when decoding a json where there is an error (e.g. type mismatch) it uses the default value and swallows the error:

```scala
import io.circe.generic.extras.Configuration
import io.circe.generic.extras.auto._

implicit val customConfig: Configuration = Configuration.default.withDefaults
final case class Foo(number: Int = 999)
io.circe.parser.decode[Foo]("""{"number": "not a number!"}""")

res0: Either[io.circe.Error,Foo] = Right(Foo(999))
```

Ideally, I'd like the default value to be used only if the field was missing from the json, and get an error when it was there but incorrectly.

Can I change this behavior somewhere central so that I don't need to have custom decoders for all my case classes with defaults?

exoego:
Maybe this PR ?https://github.com/circe/circe/pull/994

laszlox:
I think that doesn't fix it, because it changes the decode behavior only for options, but even for options this PR makes decoding fall back to the default in more cases (missing key), still including type-mismatch or other errors.

Seems like the Decoder.Result should really represent 3 different things: success (right), error (left), missing (?) and the orDefault could fall back to default only in case of "missing", errors would be propagated.

travisbrown:
I agree that this is a bug. Unfortunately it's not trivial to fix, since the default decoding works with results, not cursors, which would make things a lot simpler.

I'm not personally using circe-generic-extras at the moment, and can't really prioritize fixing this, but contributions are very welcome.

froth:
@travisbrown I drafted something:
https://github.com/circe/circe/compare/master...froth:fixWithDefault

Instead of giving the Result to orDefault passing cursor and codec.
This of course changes the interface of ReprDecoder. I believe this is in inevitable. Would that be ok?

If you are ok with my idea of fixing the issue I would invest some time and pull it through for orDefault and orDefaultAccumulating.

froth:
Bonus Question: 
orDefault and orDefaultAccumulating are basically the same code. I think it would be possible to abstract over ApplicativeError. I am not sure if this would be overkill and/or would affect the performance. what do you think?

froth:
Took the time to try the ApplicativeError idea it out and found it was more complicated than helpful. But finished the pull request anyway while doing so. ;) Please have a look.

PR #1109, opened 2019-04-09
Fixes errorHandling for withDefault Configuration

froth:
Changes withDefault semantics to only return defaults when field is
missing and not for parse errors on existing fields.

closes https://github.com/circe/circe/issues/1024

### DIFF
--- a/modules/generic-extras/src/main/scala/io/circe/generic/extras/ConfigurableDeriver.scala
+++ b/modules/generic-extras/src/main/scala/io/circe/generic/extras/ConfigurableDeriver.scala
@@ -38,7 +38,8 @@ class ConfigurableDeriver(val c: whitebox.Context)
 
   protected[this] def decodeField(name: String, decode: TermName): Tree = q"""
     orDefault(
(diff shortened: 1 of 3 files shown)


travisbrown:
Sorry for the long delay in reviewing this, but it looks great—thanks much, @froth.

Falmarri:
@travisbrown This introduced an issue where it now won't use the default value if the field is explicitly `null` in the json (rather than missing).

```
case class OkTest(string: String = "wut")
val testJson = """{
                             |    "string" : null
                             |}""".stripMargin
decode[OkTest](testJson) shouldBe OkTest()
```
fails with `Left(DecodingFailure(String, List(DownField(string))))`

This is wrong IMO

froth:
Yeah, sounds so... sorry about that, I will look into it once I find time. Hopefully without increasing the "complex pattern match" nightmare.

travisbrown:
@Falmarri Agreed. I'm traveling at the moment but we'll get this resolved (and relevant tests in place) before the next milestone.

PR #1111, opened 2019-04-10
Topic/squants compatibility

waxzce:
This is a new module to add acompatibility with typelevel https://github.com/typelevel/squants lib


modules/squants/shared/src/main/scala/io/circe/squants/package.scala
```diff
@@ -0,0 +1,99 @@
...
+/**
+  * Provides codecs for [[https://github.com/typelevel/squants squants]] types.
+  *
+  * @author Quentin ADAM @waxzce
+  */
+
+
+package object squants {
+
+  def getCompagnonObject[A <: Quantity[A]](implicit man: TypeTag[A]) = {
```
travisbrown: `s/Compagnon/Companion/`

travisbrown: Does this need to be public? If it absolutely must be, can you add a type annotation?

modules/squants/shared/src/main/scala/io/circe/squants/package.scala
```diff
@@ -0,0 +1,99 @@
...
+              import _root_.squants.experimental.unitgroups.si.strict.implicits._
+              import _root_.squants.experimental.unitgroups.UnitGroup
+              implicit val dim: Dimension[A] = getCompagnonObject[A]
+              val unitGroup:UnitGroup[A] = _root_.squants.experimental.unitgroups.si.strict.implicits.mkSiUnitGroup
+              val defaultformatter = new DefaultFormatter[A](unitGroup)
+              Some(defaultformatter.inBestUnit(a).toString())
+            } catch {
+              case anyException: Throwable => None
+            }
+            formatedString.getOrElse(a.toString())
```
travisbrown: This makes me really uncomfortable. I've not used quants much, but in what situations are you going to fall back to `toString`? Personally I'd really rather enumerate instances rather than do this kind of thing.

modules/squants/shared/src/test/scala/io/circe/squants/SquantsSuite.scala
```diff
@@ -0,0 +1,160 @@
...
+        |  "readable" : "1,0 m/s",
+        |  "number" : 1.0,
+        |  "unit" : "m/s",
+        |  "name" : "Velocity"
+        |}
+      """.stripMargin.trim
+
+    assert(speedAsserted.equals(speedj))
+    val speedRebuild = decode[Velocity](speedAsserted).toOption
+    assert(speedRebuild.exists(_.equals(speed)))
```
travisbrown: Why not just `decode[...](...) === Right(speed)`?

travisbrown (commented):
I have a few concerns here:

First, the runtime reflection and fallback to `toString` is at odds with the rest of the stuff in this repo. That's not to say it's bad—it really might be more practical than enumerating instances (although I'm skeptical about that). I do think it might fit better in a separate `circe/circe-squants` repo than in this core one.

Second, the encoding seems to make some decisions fairly arbitrarily. What does "readable" mean here? Is this a term that's associated with quants? (I don't think it is.) What if a user wants to localize the encoding, etc.? In general circe tries pretty hard not to make decisions like this, and where it has to it provides alternatives that support configurability. For example: don't like the `{ "Right": 1 }` you get from `disjunctionCodecs`? Just use `Decoder.decodeEither("LEFT", "RIGHT")` instead.

Lastly, I'm reluctant to add new modules with external dependencies in core without confidence that the maintainers of those dependencies provide updates in a timely manner. I don't know much about squants's release cadence, but the fact that it hasn't published artifacts for 2.13.0-M5 yet makes me uncomfortable.

To sum up: I'd happily accept this as a new circe-squants repo in the circe organization today, but my personal feeling is that it doesn't belong in this repo right now.

travisbrown:
Thanks much for putting this together! I hope my comments above don't sound too nitpicky—I'd be very glad to have this functionality available to circe users.

modules/squants/shared/src/main/scala/io/circe/squants/package.scala
waxzce: yep, i'll update it no need to be public indeed

modules/squants/shared/src/main/scala/io/circe/squants/package.scala
waxzce: I'll will be switching toString in a rare situations, like, form my actual tests, never. But I prefer to include it here than break.

modules/squants/shared/src/test/scala/io/circe/squants/SquantsSuite.scala
waxzce: because i'm a copy pasting monkey ;-) you're right

waxzce:
Thanks a lot for all the work. 

the "readable" field is here for easy to read field, in the squants vision I can rename it "inBestUnit" (if you say you have 3000 MB it will print 30 GB, it's very practical) it's a good use case for front end. The normal behavior will be the use of Locale and this kind of use, but I can do another way. Maybe if I add a README to make it clear?

We are using circe and squants together on a important production related software at Clever Cloud, we will be sure to maintain the modules (we have several scala dev internally, it's OK)

Do you prefer I swtich to another repo or we stick here? (seems to be the way other typelevel, like refined, lib where)

travisbrown:
@waxzce Would you be okay with making this a new `circe/circe-squants` repo? I'd be happy to set up the template (similar to e.g. https://github.com/circe/circe-spire).

I see squants as different from refined in that refined has a solid track record of quick releases (generally within days or even hours of its upstream deps), while squants still doesn't have any 2.13.0-X releases, for example.

waxzce:
@travisbrown if it can helps, i'm OK to move it in another repo

waxzce:
@travisbrown do you it on a seperate repo in circe organisation or in my company one?

ryan-richt:
hey folks! Just wondering what happened to this? +1 on this great work from us, we would also use and help contribute to this! We had a similar project at a previous company, but didnt get to get that Squants <-> Circe work out to open source then 😢 🐼

travisbrown:
@ryan-richt I'm still happy to set up a circe-squants repo!

guizmaii:
@waxzce any news?

Lasering:
A much simpler version (doesn't support reading from `String`):

```scala
import squants.{Quantity, UnitOfMeasure}
def squantsCodecIn[Q <: Quantity[Q], T: Decoder: Encoder: Numeric](in: UnitOfMeasure[Q], toT: Double => T = identity): Codec[Q] =
  Codec.from(
    Decoder[T].map(in.apply),
    Encoder[T].contramap[Q](q => toT(q.to(in))),
  )
```

Usage examples:

```scala
import squants.electro.{ElectricPotential, Kilovolts, Volts}
// Expects the json to be an Int in Volts. If the json is `25000` then the decoder will treat it as 25000.volts
given Codec[ElectricPotential] = squantsCodecIn(Volts, _.toInt)
// The same but treats the number in kilovolts, aka, 25000.kilovolts
given Codec[ElectricPotential] = squantsCodecIn(Kilovolts, _.toInt)
// The same but reads from a Float.
given Codec[ElectricPotential] = squantsCodecIn(Kilovolts, _.toFloat)
// The same but reads from a Double.
given Codec[ElectricPotential] = squantsCodecIn(Kilovolts)
```

hamnis:
Closed due to age.

PR #1117, opened 2019-04-29
Strict de-serialization for configured semi-auto decoder derivation

pfcoperez:
Partially addresses https://github.com/circe/circe/issues/374.

This PR adds `strictDeserialization` boolean flag to generic-extras' [`Configuration`](https://github.com/circe/circe/blob/master/modules/generic-extras/src/main/scala/io/circe/generic/extras/Configuration.scala#L18). 

When enabled, `decodeCaseClass` checks whether superfluous fields are present in the input JSON and, if that is the case, reports an error indicating the fields that are present in the decoded JSON but not expected in the target product.


travisbrown:
Thanks for adding this! I'm going to go ahead and merge it, but I'll probably propose a follow-up PR changing the naming to `strictDecoding`, since in general here we use "decoding" for anything involved in transforming well-formed JSON into Scala types, and "deserialization" only in a more hand-wavy way to refer to the whole process of going from Scala types to strings or bytes and back.

pfcoperez:
Great thanks for reviewing and merging this feature 🙇

Issue #1130, opened 2019-05-22
Add an option to fail on duplicate key when parsing an object


blitzkr1eg:
{ "a": 1, "a": 2 }

The above json is invalid (by jsonLint) but the parser succeeds and keeps the last value seen.
Add a mechanism to specify if duplicate keys are allowed or not when parsing.

plokhotnyuk:
It is must have feature and it can be easy implemented by adding a configuration flag which will turn on checking of the result of all `m.put` calls [here](https://github.com/circe/circe/blob/master/modules/jawn/src/main/scala/io/circe/jawn/CirceSupportParser.scala).

blitzkr1eg:
Thanks. I'm trying the fix.

travisbrown:
Fixed in #1131 and #1149. Thanks @blitzkr1eg!

PR #1131, opened 2019-05-22
Add an option to fail on duplicate key when parsing a json object

blitzkr1eg:
Fixes https://github.com/circe/circe/issues/1130
by adding a new flag to the parser: allowDuplicateKeys

Also added a factory to build parsers with custom parameters: ParserFactory


blitzkr1eg:
I see the Travis build could not run. I don't know how to retrigger/fix.

travisbrown:
We’ve been seeing these `openjdk8` failures for the past couple of days (but only on a couple of repos, I think—and restarts don’t seem to help). I’m traveling through Wednesday but will take a closer look when I’m back if nobody beats me to it.

modules/parser/jvm/src/main/scala/io/circe/parser/ParserFactory.scala
```diff
@@ -0,0 +1,8 @@
+package io.circe.parser
+
+import io.circe.Parser
+import io.circe.jawn.JawnParser
+
+case class ParserFactory(maxValueSize: Option[Int], allowDuplicateKeys: Boolean) {
```
travisbrown: Why is this needed? I'd prefer to keep the Scala.js and JVM APIs in circe-parser identical.

modules/jawn/src/main/scala/io/circe/jawn/JawnParser.scala
```diff
@@ -18,16 +18,37 @@ final object JawnParser {
...
+    * digits) may support denial-of-service attacks. For example, the string
+    * constructor for Java's `BigInteger` is quadratic with the length of the
+    * input, and decoding a ten-million digit JSON number into a `BigInteger` may
+    * take minutes.
+    */
+  def apply(maxValueSize: Int, allowDuplicateKeys: Boolean): JawnParser = new JawnParser(Some(maxValueSize), allowDuplicateKeys)
+
+
+  /**
+    * Returns a parser that fails on encountering duplicate keys as per JSONlint.
```
travisbrown: This comment doesn't seem accurate, since the user could provide `allowDuplicateKeys = true`.

modules/jawn/src/main/scala/io/circe/jawn/JawnParser.scala
```diff
@@ -18,16 +18,37 @@ final object JawnParser {
...
+    */
+  def apply(allowDuplicateKeys: Boolean): JawnParser = new JawnParser(None, allowDuplicateKeys)
 }
 
-class JawnParser(maxValueSize: Option[Int]) extends Parser {
-  def this() = this(None)
+class JawnParser(maxValueSize: Option[Int], allowDuplicateKeys: Boolean) extends Parser {
```
travisbrown: Can you add a default `true` value here?

modules/jawn/src/main/scala/io/circe/jawn/CirceSupportParser.scala
```diff
@@ -93,16 +91,24 @@ class CirceSupportParser(maxValueSize: Option[Int]) extends SupportParser[Json]
...
         key = null
       }
       final def finish(index: Int): Json = Json.fromJsonObject(JsonObject.fromLinkedHashMap(m))
       final def isObj: Boolean = true
     }
   }
 
+  private def safePut[K, V](key: K, value: V, map: LinkedHashMap[K, V], allowDuplicateKeys: Boolean): V = {
+    val oldValue = map.put(key, value)
+    if (oldValue != null && !allowDuplicateKeys)
```
travisbrown: Can you use `oldValue.ne(null)`? Reference equality is fine for `null` checks.

modules/jawn/src/main/scala/io/circe/jawn/JawnParser.scala
travisbrown: I think I'd prefer to omit this overload.

modules/jawn/src/main/scala/io/circe/jawn/CirceSupportParser.scala
```diff
@@ -93,16 +91,24 @@ class CirceSupportParser(maxValueSize: Option[Int]) extends SupportParser[Json]
...
+      throw new IllegalArgumentException(s"Invalid json, duplicate key name found: $key")
```
travisbrown: I think it'd be better to use `org.typelevel.jawn.ParseException` here, no?

travisbrown: Nevermind—I agree `IllegalArgumentException` is the right thing to use here.

travisbrown:
The commits here are included in #1149.

PR #1135, opened 2019-05-28
Json.findAllByKey: avoid allocations.

diesalbla:
The existing code of `findAllByKey` was performing several memory allocations, not needed to create the final result.

- It creates a long list with `keyValues` from which it filters   only the pairs with a certain Key.
- Performs several transformations of `Vector` to list, which may  mean copying the whole Vector.
- For each field, it creates a new tuple-pair.

Instead of all that, we use a mutable ListBuffer, and a recursive loop that runs through the Json, picking only the fields and values that it cares about.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Json.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Json.scala
@@ -3,6 +3,7 @@ package io.circe
 import cats.{ Eq, Show }
 import io.circe.numbers.BiggerDecimal
 import java.io.Serializable
+import scala.collection.mutable.ListBuffer
 
 /**
  * A data type representing possible JSON values.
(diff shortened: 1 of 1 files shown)


travisbrown:
Thanks for doing this, @diesalbla! Would you mind merging master?

modules/core/shared/src/main/scala/io/circe/Json.scala
```diff
@@ -202,18 +203,18 @@ sealed abstract class Json extends Product with Serializable {
...
+  final def findAllByKey(key: String): List[Json] = {
+    val hh: ListBuffer[Json] = ListBuffer.empty[Json]
+    def loop(json: Json): Unit = json match {
+      case JObject(obj)  => obj.toIterable.foreach { case (k, v) =>
+        if (k == key) hh += v
+        loop(v)
+      }
+      case JArray(elems) => elems.foreach(loop)
+      case _             => // do nothing
+    }
```
travisbrown: Looks like you need to call `loop(key)` here.

diesalbla: Right. It needs the initial call.

travisbrown:
Looks like it just needs `sbt scalafmt` now.

Issue #1136, opened 2019-05-29
generic encoders ignore "non-constructor" case class members


aSapien:
Example:

```scala
object Example extends App {
  import io.circe.syntax._
  import io.circe.Printer.noSpaces.pretty
  import io.circe.generic.auto._

  case class Bar(foo: Int) {
    val baz: String = "boom"
  }

  println(pretty(Bar(1).asJson))
}
```

Expected: `{"foo":1,"baz":"boom"}`
Actual: `{"foo":1}`

I tried to enable `withDefaults` using `io.circe.generic.extras.Configuration`, but to no avail. 

I believe it is a common use-case in Scala. 
Why is this the default behavior? 
Is there a configuration option I'm missing? 

I'd really like to avoid creating my own `Encoder` for this.

Thanks,
Dima

plokhotnyuk:
@aSapien Have you considered an option to move `baz` to the list of arguments, like here below?

```
case class Bar(foo: Int, baz: String = "boom")
```

aSapien:
@plokhotnyuk that's what I did. For now :) 
However, I'd like to understand if this is a configurable feature, and if not, the reasoning behind it.

travisbrown:
@aSapien All of the functionality in circe-generic-* is built around Shapeless's generic representation of case classes and sealed trait hierarchies, which does not include `val`s defined in the body of the class:

```scala
scala>   case class Bar(foo: Int) {
     |     val baz: String = "boom"
     |   }
defined class Bar

scala> shapeless.LabelledGeneric[Bar]
res0: shapeless.LabelledGeneric[Bar]{type Repr = Int with shapeless.labelled.KeyTag[Symbol with shapeless.tag.Tagged[String("foo")],Int] :: shapeless.HNil} = shapeless.LabelledGeneric$$anon$1@1f39c060
```
You could write your own `LabelledGeneric` instances (either per-class or as a generic macro) that would include things like `baz` as members, and everything in circe would just work as you expect. In my view this is how things should be for circe-generic and circe-generic-extras: `shapeless.LabelledGeneric` underlies everything, and while circe-generic does have some of its own macros, these are only for the purposes of efficiency—they don't try to extend or replace the work done by Shapeless.

[...]

PR #1138, opened 2019-05-30
Fix Travis failed to install JDK

exoego:
Recent PRs like #1131, #1132, #1133, #1135 and #1136 due to JDK installation failure on Travis.

```
install-jdk.sh 2019-05-02
Expected feature release number in range of 9 to 13, but got: 8
The command "~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"" failed and exited with 3 during .
```

This PR fixes it.

[As of  2019-04-15, `xenial` is now default and recommended on Travis](https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment), so let's move on instead of staying trusty. 
This PR explicitly specifies the build environment, so it wont be automatically updated to latest environments and break builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: xenial
 language: scala
 
 scala:
@@ -6,7 +7,7 @@ scala:
   - 2.12.8
 
 jdk:
-  - oraclejdk8
+  - openjdk8
 
 cache:
   directories:


travisbrown:
Excellent, thanks for looking at this!

PR #1140, opened 2019-05-30
EnumerationEncoder/Decoder should respect Configuration

nadavwr:
Currently, `EnumerationEncoder`/`Decoder` will use case object names as-is even when a `Configuration` with snakecase constructor names is in scope.

This PR makes `EnumerationEncoder`/`Decoder` rely on `Configuration.transformConstructorNames` when a configuration is available.

I made the configuration parameter default to `Configuration.defaults`—this isn't consistent with `deriveDecoder` and `deriveEncoder`, but can reduce disruption should this PR be accepted.


travisbrown:
This looks reasonable to me, thanks! I'll leave it open for a bit in case anyone else has concerns, since I'm not personally an active circe-generic-extras user, but assuming nobody speaks up it'll be in the next 0.12 milestone.

Issue #1142, opened 2019-06-03
Defaults are no longer used for null-valued fields


travisbrown:
The behavior of the following code changed in 0.12.0-M1:

```scala
import io.circe.generic.extras.Configuration
import io.circe.generic.extras.auto._
import io.circe.jawn.decode

implicit val config: Configuration = Configuration.default.withDefaults

case class Foo(x: String = "abc")

decode[Foo]("""{"x":null}""")
```
In 0.11.x and earlier versions decoding would succeed, but now it fails. I think it's pretty clear that the previous behavior was correct. See [this thread](https://github.com/circe/circe/pull/1109#issuecomment-497992407) for more discussion.

0.12.0-M3 will go back to the old behavior. We'll need a fix and tests, which I probably wouldn't be able to get to until this Friday at the earliest, so hopefully someone will beat me to it.

Thanks to @Falmarri for reporting this in #1109.

travisbrown:
Note that normally I'd rush a fix for an issue like this, but this is `-extras`, which is explicitly experimental, and these are milestone releases, so I think a couple of days after the report isn't too terrible.

froth:
I have a question. There is a test that ensures that for Optional Values the default should not be used in case of null. 
   "Option[T] with default" should "be None if null decoded" in {
      val json = json"""{ "a": null }"""
      assert(Decoder[FooWithDefault].decodeJson(json) === Right(FooWithDefault(None, "b")))
    }
So the wanted behaviour is the following?
a) if key null and Option decoder -> None
b) if key missing and Option decoder -> Default value (?)
c) if key null or missing and not Option decoder -> default value (?)

froth:
If that is the intended behaviour I hope I got it right this time :)

PR #1143, opened 2019-06-03
Fixes defaults for null-valued fields

froth:
fixes https://github.com/circe/circe/issues/1142


froth:
> Thanks. The match is getting kind of messy but this looks right to me.

Yeah, I was unhappy with it even before this additional fix :(
The whole "if the decoding fails and we are on special cursor conditions then do some special handling" logic is not very nice in my opinion. But no real good solution came to my mind, yet.

PR #1148, opened 2019-06-09
Follow-up to #1117

travisbrown:
This makes the `strictDeserialization`-to-`strictDecoding` name change I mentioned in #1117, and also rearranges the implementation a little. The only substantive change is that if `maybeUnexpectedErrors` ends up being `None`, the new implementation does not fail. It should only ever be `None` in the case where the focus isn't a JSON object, and in that case we wouldn't have gotten even this far.

r? @pfcoperez


pfcoperez:
Thanks for the improvements! It's great too that you moved the class definition so the access modifiers for the auxiliary methods don't need to change.

PR #1150, opened 2019-06-10
Rename XEncoder to Encoder.AsX

travisbrown:
We currently have `Encoder`, `ObjectEncoder`, `ArrayEncoder`, and `RootEncoder` all at the top level of the `io.circe` package—not to mention `KeyEncoder`, which only adds to the confusion, since it's a totally different kind of thing.

I'm in the process of adding a new `Codec` type class (in a separate branch), and I don't want to clutter up the package even more with `ObjectCodec`, etc.

This PR implements a proposal that would rename e.g. `ObjectEncoder` to `Encoder.AsObject`. Instead of `ObjectCodec` we can then have `Codec.AsObject` without being inconsistent, and we actually end up with a smaller `io.circe` package footprint than we had before.

The old `ObjectEncoder`, `ArrayEncoder`, etc. would be available but deprecated. I think we could remove `ArrayEncoder` and `RootEncoder` in 0.13.x, but we probably want to keep `ObjectEncoder` around a little longer.


lorandszakacs:
@travisbrown, should I revamp this PR if Codec is back in discussion?
https://github.com/circe/circe/pull/811

travisbrown:
@lorandszakacs I'm taking a different approach—the main motivation is to make semi-automatic derivation more efficient, so having `deriveCodec` rely on resolving separate encoders and decoders won't work. I'm also intentionally and explicitly not providing any implicit summoning for `Codec` instances given encoders and decoders (the idea is that `Codec` is a convenience for definitions and generally should not be used as a constraint).

I'll open a PR soon and we can discuss the merits of my approach vs. the one in your PR. Thanks for bringing it up.

lorandszakacs:
@travisbrown if you can squeeze efficiency out of your approach then it does sound better.

At the very least I'll rebase my branch and bring it up to date.

PR #1151, opened 2019-06-10
Codec

travisbrown:
I've resisted adding this thing for years, but I've finally been convinced that having `Codec` around for more convenient definitions (and faster semi-automatic derivation) is a good idea. This PR builds on #1150 (which is motivated by this change but should stand on its own) and includes @OlivierBlanvillain's change from #301. It doesn't build on @lorandszakacs's #811 at the moment, but I'll probably pull some tests from there.

Note that the intention is that `Codec` should only be used to make defining instances more convenient, and not as a constraint. The core module provides no `Codec` instances for any standard library types, and there is no machinery for summoning `Codec` instances given an encoder-decoder pair. Fully-automatic generic derivation does not (and probably will not) provide `Codec` instances, although semi-automatic derivation is supported in both the vanilla circe-generic form and with configuration in circe-generic-extras.

The motivation is to speed up semi-automatic generic derivation (it should be not too far off twice as fast as deriving separate encoders and decoders, ~~but I've not actually done any benchmarking yet~~) and to avoid forcing users to write out member names twice when using `forProductN`.

Update: I put together a file with 200 case classes and either `deriveCodec` or a `deriveDecoder` / `deriveEncoder` pair in the companion class, and after half a dozen runs of each on Scala 2.12.8, the new `deriveCodec` approach consistently compiles in 20-21 seconds, while the old approach takes 35-36 seconds. So about 42% faster for that example case, which is actually a little better than I expected.

The implementation currently supports writing this:

```scala
import io.circe.Codec
import io.circe.generic.semiauto.deriveCodec

case class Foo(a: String, b: List[Boolean], c: Double)

object Foo {
  implicit val fooCodec: Codec[Foo] = deriveCodec
}
```
…or this:

```scala
import io.circe.Codec

case class Foo(a: String, b: List[Boolean], c: Double)

object Foo {
  implicit val fooCodec: Codec[Foo] =
    Codec.forProduct3("a", "b", "c")(Foo(_, _, _))(foo => (foo.a, foo.b, foo.c))
}
```
…instead of this:
```scala
import io.circe.{ Decoder, ObjectEncoder }

case class Foo(a: String, b: List[Boolean], c: Double)

object Foo {
  implicit val decodeFoo: Decoder[Foo] = deriveDecoder
  implicit val encodeFoo: ObjectEncoder[Foo] = deriveEncoder
}
```
…or this:
```scala
import io.circe.{ Decoder, Encoder, ObjectEncoder }

case class Foo(a: String, b: List[Boolean], c: Double)

object Foo {
  implicit val decodeFoo: Decoder[Foo] =
    Decoder.forProduct3("a", "b", "c")(Foo(_, _, _))

  implicit val encodeFoo: ObjectEncoder[Foo] =
    Encoder.forProduct3("a", "b", "c")(foo => (foo.a, foo.b, foo.c))
}
```

It still needs tests and better docs and probably some clean-up.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Codec.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Codec.scala
@@ -1,0 +1,17 @@
+import io.circe._
+
+/**
+ * A type class that provides back and forth conversion between values of type `A`
+ * and the [[Json]] format. Must obey the laws defined in [[io.circe.tests.CodecLaws]].
+ */
+trait Codec[A] extends Encoder[A] with Decoder[A]
+
+object Codec {
+  def apply[A](implicit instance: Codec[A]): Codec[A] = instance
+
+  implicit def fromEncoderDecoder[A](implicit e: Encoder[A], d: Decoder[A]): Codec[A] =
+    new Codec[A] {
+      def apply(c: HCursor): Decoder.Result[A] = d(c)
+      def apply(a: A): Json = e(a)
+    }
+}
--- a/modules/core/shared/src/main/scala/io/circe/Codec.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Codec.scala
@@ -1,17 +1,52 @@
-import io.circe._
+package io.circe
 
 /**
  * A type class that provides back and forth conversion between values of type `A`
- * and the [[Json]] format. Must obey the laws defined in [[io.circe.tests.CodecLaws]].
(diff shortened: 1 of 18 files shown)


tpolecat:
doobie has `Meta` that works exactly the same way. You can use it to introduce a symmetric `Get`/`Put` pair but you never demand it as a evidence. It's bad and I feel bad but users like it.

PR #1169, opened 2019-06-16
Better caching for Travis CI

exoego:
This PR updates the Travis CI cache settings by following the [Travis CI sample settings](https://www.scala-sbt.org/1.x/docs/Travis-CI-with-sbt.html#Sample+setting) updated for sbt 1.x.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,6 +41,9 @@ script:
     else echo "Skipping Scala.js build";
     fi
 
-  # See http://www.scala-sbt.org/0.13/docs/Travis-CI-with-sbt.html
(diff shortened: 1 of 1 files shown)


travisbrown:
When we updated to sbt 1.3.0-RC1 I had to remove some similar code because `find` was returning no results and builds were failing (because of the switch from Ivy to Coursier, I guess?): https://github.com/circe/circe/pull/1127

exoego:
Oops, you are correct.
This change seems no longer required since
sbt 1.3.0 switched from Ivy to Coursier internally...

Issue #1168, opened 2019-06-15
Replace annonymous class with lambda expression


exoego:
There are lots (100+) of annonymous class instantiation of SAM (Single Abstract Method) types in circe code base.
https://github.com/circe/circe/blob/23af4778f1bf5b8ba8cfc93a6c32b83350a1dffc/modules/shapes/src/main/scala/io/circe/shapes/CoproductInstances.scala#L7-L9

If I understand correctly, replacing them with lambda expression can eliminate `.class` files for anonymous classes in Scala 2.12, since Scala 2.12 can use bytecode `invokedynamic` for lambdas.
Therefore refactoring with lamda expression will result in 

* Faster compilation
* Smaller JAR file
* Faster runtime performance

Example of refactoring:

```scala
// Before
implicit final val decodeCNil: Decoder[CNil] = new Decoder[CNil] {
  def apply(c: HCursor): Decoder.Result[CNil] = Left(DecodingFailure("CNil", c.history))
}

// After
implicit final val decodeCNil: Decoder[CNil] =
  (c: HCursor) => Left(DecodingFailure("CNil", c.history))
```

travisbrown:
Unfortunately the compiler won't actually use `LambdaMetafactory` for instances of `Decoder` defined with the SAM syntax, so even if we were willing to use the experimental SAM support for 2.11 in circe (which I'm not sure about—I seem to remember running into problems with it), it wouldn't really help us out in this respect, since we'd still get the anonymous class representation.

At some point a couple of years ago I spent some time working through what the 2.12 compiler does in cases like this, but it seemed like a confusing mess and to be honest I don't really remember why `Decoder` doesn't get the LMF treatment. Even this simplified version doesn't

```scala
trait Decoder[A] {
  def apply(input: String): Option[A]
  final def orElse(input: String, a: A): A = apply(input).getOrElse(a)
}

object Test {
  val stringDecoder: Decoder[String] = (input: String) => Some(input)
}
```
This will still end up with an anonymous class for the `stringDecoder` implementation. I don't know why—from my reading of the (scant) documentation it seems like this `Decoder` is pretty clearly a "simple interface" that should qualify. Remove the `orElse` and you'll get the LMF version, but that's not really an option for us.

[...]

PR #1170, opened 2019-06-16
Add benchmarks for new vs. SAM vs. instance

travisbrown:
See #1168. I'm running the benchmarks locally now.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -508,6 +508,10 @@ lazy val benchmark = circeModule("benchmark", mima = None)
     scalacOptions ~= {
       _.filterNot(Set("-Yno-predef"))
     },
+    scalacOptions ++= (CrossVersion.partialVersion(scalaVersion.value) match {
+      case Some((2, 11)) => Seq("-Xexperimental")
+      case _             => Nil
+    }),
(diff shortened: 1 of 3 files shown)


travisbrown:
Results for `Java HotSpot(TM) 64-Bit Server VM (build 25.92-b14, mixed mode)` and 2.12 on my Linux desktop:
```
Benchmark                                    Mode  Cnt         Score         Error  Units
InstantiationBenchmark.decoderFromInstance  thrpt   50  29130370.899 ±  184188.680  ops/s
InstantiationBenchmark.decoderFromNew       thrpt   50  26918856.742 ±   83697.771  ops/s
InstantiationBenchmark.decoderFromSAM       thrpt   50  29945941.246 ±  379555.540  ops/s
InstantiationBenchmark.encoderFromInstance  thrpt   50  23388206.168 ±   57265.405  ops/s
InstantiationBenchmark.encoderFromNew       thrpt   50  20393834.074 ±  509535.100  ops/s
InstantiationBenchmark.encoderFromSAM       thrpt   50  20920077.835 ± 1030026.931  ops/s
```
This don't make a lot of sense to me, but we should look closer at some point.

PR #1180, opened 2019-06-23
Experimenting with by-name implicit parameters for generic derivation

travisbrown:
This PR introduces a 2.13-only rewrite of circe-generic (as a new circe-generic-simple module, for now at least) that passes the same tests as circe-generic but completely removes `DerivationMacros`; the only macros are for the export mechanism and the `@JsonCodec` macro annotation. All of the `shapeless.Lazy` occurrences are gone—most have been replaced by by-name implicit parameters, but a few don't seem to be needed at all with 2.13.

In addition to having a less terrifying implementation, circe-generic-simple also provides much faster compilation (up to ~30% for some of my examples).


travisbrown:
Note that we'll need https://github.com/scalameta/scalameta/pull/1880 for Scalafmt to work here (it seems like you can set `runner.dialect = dotty` as a workaround).

modules/generic-simple/src/main/scala/io/circe/generic/simple/codec/DerivedAsObjectCodec.scala
```diff
@@ -0,0 +1,23 @@
...
+
+abstract class DerivedAsObjectCodec[A] extends Codec.AsObject[A]
+
+final object DerivedAsObjectCodec {
+  implicit def deriveCodec[A, R](
+    implicit
+    gen: LabelledGeneric.Aux[A, R],
+    codec: => ReprAsObjectCodec[R]
+  ): DerivedAsObjectCodec[A] = new DerivedAsObjectCodec[A] {
+    final def apply(c: HCursor): Decoder.Result[A] = codec.apply(c) match {
```
johnynek: Do you not need to cache code in a lazy val to avoid executing it again and again for each call to apply?

travisbrown: Yeah, I meant to go back and do a pass for runtime performance stuff like this after getting it correct. I've just opened #1183. Thanks for catching.

PR #1182, opened 2019-06-24
Update Sigma to Merit in README

JSchwerberg:
Sigma is now Merit! (and is still using Circe)
https://www.merits.com/blog/sigma-rebrands-as-merit


travisbrown:
Thanks @JSchwerberg!

PR #1186, opened 2019-06-27
Cursor clean-up

travisbrown:
I'm trying to pick up [circe-algebra](https://github.com/circe/circe-derivation) again, and I'm working backwards toward it a bit by minimizing the cursor API in circe-core.

I'm removing most of the cursor modification operations, and the navigation operations that take a predicate. I've literally never used any of these outside of the circe test suite, or seen anyone else use them, and all of them can be replaced if necessary by working at the `Json` level.

I'm also removing `lefts` and `rights`, which I've also never used or seen used. Both can be replaced pretty easily with `values` and `drop` or `take`.

I've deprecated a few navigation operations that I _have_ used or seen used but that are easily replaced by more primitive operations.

I was tempted to get rid of `DownArray` (which is just `DownN(0)`) and `Field` (`Up` and `DownField`) as well, but `DownArray` is used in `SeqDecoder` and saves some allocations, and `Field` _could_ be used similarly.

The full list of what's gone:

* `DeleteGoLeft`
* `DeleteGoRight`
* `DeleteGoFirst`
* `DeleteGoLast`
* `DeleteGoField`
* `DeleteLefts`
* `DeleteRights`
* `SetLefts`
* `SetRights`
* `ACursor#lefts`
* `ACursor#rights`

And what's deprecated:

* `MoveFirst`
* `LeftN`
* `RightN`


agourlay:
Responding to https://twitter.com/travisbrown/status/1144230565727617026 

My project uses `rightN` to implement `JsonPath` on top of `Circe`.
https://github.com/agourlay/cornichon/blob/master/cornichon-core/src/main/scala/com/github/agourlay/cornichon/json/JsonPath.scala#L70

What would my migration path be like?

travisbrown:
@agourlay For any instance of `rightN` immediately after `downArray`, both can be replaced by a single `downN`.

agourlay:
This is neat 👍 

It is shorter to write and allocates less.

Thanks a lot!

PR #1190, opened 2019-06-30
A few optimizations

travisbrown:
In particular in `MapDecoder` we can avoid checking for JSON object and key presence with `downField` for every key.


travisbrown:
For the `MapDecoder` change, here's before:
```
MapDecoderBenchmark.decodeNestedMap  thrpt   20  1580.659 ± 21.327  ops/s
```
And after:
```
MapDecoderBenchmark.decodeNestedMap  thrpt   20  2119.333 ± 8.666  ops/s
```
So for large maps it's a pretty big win.

PR #1192, opened 2019-07-02
Updated README with panaseer

TobiasRoland:
Hey, I'm a dev at this company, and we've been using circe for about year now. It's made handling and maintaing our yaml and json configuration files so much easier than it was before. It's been the gateway drug for cats as well within our organisation, so I have only thanks to give to all the maintainers.


travisbrown:
Thanks much @TobiasRoland!

PR #1196, opened 2019-07-09
Add docs about Plated (circe-optics)

note:


docs/src/main/tut/optics.md
```diff
@@ -118,6 +118,23 @@ val modifiedJson = doubleQuantities(json)
...
+## Recursively modifying JSON
+
+Sometimes you may need to recursively modify JSON. Let assume you need to transform all numbers into
+strings in the example JSON:
+
+```tut:book
+import io.circe.optics.JsonOptics._
+import monocle.function.Plated
+
+Plated.transform[Json] { json =>
```
travisbrown: What do you think about changing this to something like `j` or `value` to avoid shadowing the example `json` in the outer scope?

note: Sure, good idea

PR #1201, opened 2019-07-12
remove redundant final object.

takayahilton:


travisbrown:
For a long time the `final` modifier [did make a difference in this context](https://stackoverflow.com/q/26079591/334519), both in terms of semantics and byte code, but I don't think that's the case any longer, so thanks, @takayahilton!

PR #1199, opened 2019-07-10
Fix StackOverflow of Codec[SelfRecursiveWithOption]

ashawley:
Found in the Scala Community Build, and likely caused by a seemingly innocuous change in ScalaCheck to `Arbitrary.arbOption`.

https://github.com/rickynils/scalacheck/issues/485

PR #1204, opened 2019-07-16
Fix StackOverflow of recursive Options tests

ashawley:
Travis,

I believe I better understand this code.  The depth argument is incremented, now.  I also added `Gen.delay` to protect from infinite recursion, as well.

Follow-up to #1199.


travisbrown:
Looks good to me, thanks again!

PR #1208, opened 2019-07-29
Remove redundant collection conversion in scalajs

exoego:


travisbrown:
Thanks! I think we tried this before but it broke the build on 2.11.

PR #1214, opened 2019-08-06
ValidatedNec should be used instead of ValidatedNel for error accumulation.

takayahilton:
List's concat is O(N), but Chain's concat is O(1).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,4 +1,3 @@
-import ReleaseTransformations._
 import microsites.ExtraMdFileConfig
 import microsites.ConfigYml
 import sbtcrossproject.{ CrossProject, CrossType }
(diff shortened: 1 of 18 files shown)


travisbrown:
I'm inclined not to think this is the right thing to do. In my view the `cats.data` collections types are still kind of a mess, and the biggest virtue of `NonEmptyList` is that it can be treated pretty straightforwardly as a wrapper for `List`.

I'm also just not convinced that the differences in the performance characteristics of `NonEmptyList` vs. `NonEmptyChain` justify the change. In most cases there will only be a single error, and in the cases where there might be enough that the performance of the collection they're stored in might matter (decoding maps and sequences), we're already using `List.newBuilder` behind the scenes for error accumulation.

If you have benchmarks that suggest that the difference is non-trivial in use cases you care about, I'm happy to reconsider for 0.13.

takayahilton:
@travisbrown 
Thank you for your review comments.
Now that I understand your thoughts, I will close this PR.

Issue #1220, opened 2019-08-13
[generic-extras] configured derivation prone to using wrong import


felixbr:
Hi,

My colleagues and I use `circe-generic-extras` in several projects and it's a really useful. There is, however, a recurring scenario where someone is using it and the code just doesn't seem to do the expected thing.

```
sealed trait Foo
object Foo {
  case object Bar extends Foo
  case object Baz extends Foo

  // sometimes imported from somewhere else
  implicit val derivationConfig: Configuration = 
    Configuration.default.withDiscriminator("__typename")

  implicit val encodeFoo: Encoder[Foo] = deriveEncoder
}
```

Most of the time people think the implicit `Configuration` is not being picked up correctly, so they move it around and try all sorts of things.

The real problem usually is that they imported
`io.circe.generic.semiauto._` instead of `io.circe.generic.extras.semiauto._`. The code still compiles and runs but it does the wrong thing.

In some cases it can help to activate `-Ywarn-unused:implicits` but only if the `Configuration` wasn't used somewhere else already. (And it's still a mental leap from "unused implicit" to "my circe imports are wrong")

I know this is a long shot as it would be a breaking change, but I think renaming `deriveEncoder` and `deriveDecoder` in `circe-generic-extras` to `deriveConfiguredEncoder` and `deriveConfiguredDecoder` respectively would make this much harder to get wrong (especially for people new to circe).
If a breaking change will not be considered I'd also be fine with deprecating it while introducing the new alias.

I'd love to hear your thoughts on this :)

Cheers
~ Felix

travisbrown:
I agree that this is confusing, and personally I'd be happy to consider a PR that adds new `deriveConfiguredX` methods in generic-extras and deprecates the `deriveX` ones (I think we'll want to get more feedback before merging something like this, but there's a chance it could still happen in 0.12.0).

travisbrown:
I've been thinking a little more about this, and I'm convinced we should do it. Would you be able to open a PR?

felixbr:
I'll try to write one in the next days 👍

travisbrown:
Fixed in #1221 and will be released in an 0.12.0-RC3 this morning (probably).

PR #1221, opened 2019-08-14
Rename generic-extras configured derive functions

felixbr:
This is a PR addressing the usability problems explained in #1220.

I tried to rename all occurrences of `deriveEncoder`, `deriveDecoder`, `deriveCodec` and `deriveFor` to start with `deriveConfigured*`.

The original public functions in `io.circe.generic.extras.semiauto._` have been deprecated.

I didn't rename `deriveEnumeration*` and `deriveUnwrapped*` functions as they don't use the configuration afaik.

The methods in `ConfigurableDeriver` have been renamed without deprecation as it doesn't seem like a public API to me. Please tell me if I should add the deprecation there as well or if not renaming them is a better solution (the class is already named  `ConfigurableDeriver`, so the method names are a bit redundant).

I had to parameterize the function name prefix for the macro annotation in `JsonCodecMacros`. I'm not very experienced with macros but I tried a wrong prefix and it gave me expected compile-errors, so I think it works as it should.

Hope this is what you expected 🙂

### DIFF
--- a/modules/generic-extras/src/main/scala/io/circe/generic/extras/ConfiguredJsonCodec.scala
+++ b/modules/generic-extras/src/main/scala/io/circe/generic/extras/ConfiguredJsonCodec.scala
@@ -15,6 +15,7 @@ private[generic] class ConfiguredJsonCodecMacros(val c: blackbox.Context) extend
   import c.universe._
 
   protected[this] def semiautoObj: Symbol = symbolOf[semiauto.type].asClass.module
+  protected[this] def deriveFunctionPrefix: String = "deriveConfigured"
 
   def jsonCodecAnnotationMacro(annottees: Tree*): Tree = constructJsonCodec(annottees: _*)
 }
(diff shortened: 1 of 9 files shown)


modules/generic-extras/src/main/scala/io/circe/generic/extras/ConfiguredJsonCodec.scala
```diff
@@ -15,6 +15,7 @@ private[generic] class ConfiguredJsonCodecMacros(val c: blackbox.Context) extend
   import c.universe._
 
   protected[this] def semiautoObj: Symbol = symbolOf[semiauto.type].asClass.module
+  protected[this] def deriveFunctionPrefix: String = "deriveConfigured"
```
travisbrown: This is super nitpicky but I'd probably call this `deriveMethodPrefix`, since it's a method, not a function. I'm happy either to leave it or to change it in a follow-up, though.

felixbr: I considered "derive*" functions, as they are not bound to a class instance (in the OOP sense). I'll change it real quick 🙂

felixbr:
Should I do something about the code coverage failure? The only new untested code is the parameterization of the method name in the macro, but that's covered by the compiler imo.

travisbrown:
@felixbr No, it's fine. I should set that threshold a little higher.

PR #1243, opened 2019-09-07
Accumulation errors

ChristopherDavenport:
A conjecture of an approach, I suspect most of the accumulator code can instead be removed and rewritten in terms of this mechanic. 

Needs a bit more of the replication of a case class machinery that the SingleDecodingFailure has, but as you can see in the added test, makes it very easy to do error accumulation.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -158,7 +158,7 @@ trait Decoder[A] extends Serializable { self =>
       case r @ Right(_) => r
       case Left(e)      => e match {
         case e: SingleDecodingFailure => Left(e.withMessage(message))
-        case e: AggregatedDecodingFailure => Left(e.withMessage(message)) // Not sure I like this
+        case e: AggregatedDecodingFailure => Left(e.withMessage(message))
       }
     }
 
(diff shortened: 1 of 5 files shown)


travisbrown:
I want to go ahead and publish 0.12.0, but we'll work through this in the 1.0.0 milestones.

ChristopherDavenport:
That sounds good. `.mapN` works as well as `.parMapN` for my decoders presently. I do feel this is a major optimization to get all the errors though, which will help with client debugging etc in a faster iteration cycle.

modules/core/shared/src/main/scala/io/circe/Error.scala
```diff
@@ -42,20 +42,24 @@ object ParsingFailure {
  * An exception representing a decoding failure and (lazily) capturing the
  * decoding history resulting in the failure.
  */
-sealed abstract class DecodingFailure(val message: String) extends Error {
+sealed trait DecodingFailure extends Error {
+  val message: String
```
jeffmay: Why not add the `def withMessage(): DecodingFailure` method here?

modules/core/shared/src/main/scala/io/circe/Decoder.scala
```diff
@@ -156,11 +156,25 @@ trait Decoder[A] extends Serializable { self =>
...
 
     override def decodeAccumulating(c: HCursor): Decoder.AccumulatingResult[A] =
-      self.decodeAccumulating(c).leftMap(_.map(_.withMessage(message)))
+      self
+        .decodeAccumulating(c)
+        .leftMap(
+          _.map(
+            e =>
+              e match {
+                case e: SingleDecodingFailure     => e.withMessage(message)
```
jeffmay: You could just flatten this to:
```
_.map {
  case e: SingleDecodingFailure     => e.withMessage(message)
  case e: AggregatedDecodingFailure => e.withMessage(message)
}
```

PR #1246, opened 2019-09-08
Make most implicit {enc,dec}oders lazy

### LINKED ISSUE #1245: Scala.JS linking errors with 0.12.0-RC4
Using 0.12.0-RC4 with Scala.JS, everything compiles but then it fails at link-time with lots of linking errors. They all seem to be related to `java.time`. Some of `java.time` is implemented for Scala.JS under `"org.scala-js" %%% "scalajs-java-time" % "0.2.5"` but not all of it, and not enough to prevent these linking errors.

The sad thing is that this will affect all Scala.JS users and not just those who attempt to use `java.time` decoders. The `Decoder` object has a bunch of `implicit val`s and Scala.JS currently doesn't eliminate them as dead-code yet, the workaround for now is to make them all `lazy val`s instead which *does* get DCE'd. Or drop back to Circe 0.11.1 but then one loses Scala 2.13 support.

Are you happy to make the out-of-the-box implicits lazy? If so, I'm happy to submit a PR.

Sample errors:

```
[error] Referring to non-existent method java.time.format.DateTimeFormatter$.ISO$undOFFSET$undTIME()java.time.format.DateTimeFormatter
[error]   called from io.circe.Decoder$$anon$70.parseUnsafe(java.lang.String)java.time.OffsetTime
[error]   called from io.circe.Decoder$$anon$70.parseUnsafe(java.lang.String)java.lang.Object
[error]   called from io.circe.Decoder$JavaTimeDecoder.apply(io.circe.HCursor)scala.util.Either
[error]   called from io.circe.Decoder.tryDecode(io.circe.ACursor)scala.util.Either
[error]   called from io.circe.Decoder$JavaTimeDecoder.tryDecode(io.circe.ACursor)scala.util.Either
[error]   called from io.circe.ACursor.as(io.circe.Decoder)scala.util.Either

[error] Referring to non-existent method java.time.Period$.parse(java.lang.CharSequence)java.time.Period
[error]   called from io.circe.Decoder$$anon$54.parseUnsafe(java.lang.String)java.time.Period
[error]   called from io.circe.Decoder$$anon$54.parseUnsafe(java.lang.String)java.lang.Object
[error]   called from io.circe.Decoder$JavaTimeDecoder.apply(io.circe.HCursor)scala.util.Either
[error]   called from io.circe.Decoder.tryDecode(io.circe.ACursor)scala.util.Either
[error]   called from io.circe.Decoder$JavaTimeDecoder.tryDecode(io.circe.ACursor)scala.util.Either
[error]   called from io.circe.ACursor.as(io.circe.Decoder)scala.util.Either

```

japgolly:
Closes #1245.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -604,7 +604,7 @@ object Decoder
    *
    * @group Decoding
    */
-  implicit final val decodeJavaBoolean: Decoder[java.lang.Boolean] = decodeBoolean.map(java.lang.Boolean.valueOf)
+  implicit final lazy val decodeJavaBoolean: Decoder[java.lang.Boolean] = decodeBoolean.map(java.lang.Boolean.valueOf)
 
   /**
    * @group Decoding
@@ -621,7 +621,7 @@ object Decoder
    *
    * @group Decoding
    */
-  implicit final val decodeJavaCharacter: Decoder[java.lang.Character] = decodeChar.map(java.lang.Character.valueOf)
+  implicit final lazy val decodeJavaCharacter: Decoder[java.lang.Character] = decodeChar.map(java.lang.Character.valueOf)
 
   /**
    * Decode a JSON value into a [[scala.Float]].
@@ -630,7 +630,7 @@ object Decoder
    *
    * @group Decoding
    */
-  implicit final val decodeFloat: Decoder[Float] = new DecoderWithFailure[Float]("Float") {
+  implicit final lazy val decodeFloat: Decoder[Float] = new DecoderWithFailure[Float]("Float") {
     final def apply(c: HCursor): Result[Float] = c.value match {
       case Json.JNumber(number) => Right(number.toFloat)
       case Json.JString(string) =>
@@ -648,7 +648,7 @@ object Decoder
    *
    * @group Decoding
    */
-  implicit final val decodeJavaFloat: Decoder[java.lang.Float] = decodeFloat.map(java.lang.Float.valueOf)
+  implicit final lazy val decodeJavaFloat: Decoder[java.lang.Float] = decodeFloat.map(java.lang.Float.valueOf)
 
   /**
    * Decode a JSON value into a [[scala.Double]].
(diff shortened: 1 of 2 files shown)


modules/core/shared/src/main/scala/io/circe/Decoder.scala
```diff
@@ -604,7 +604,7 @@ object Decoder
    *
    * @group Decoding
    */
-  implicit final val decodeJavaBoolean: Decoder[java.lang.Boolean] = decodeBoolean.map(java.lang.Boolean.valueOf)
+  implicit final lazy val decodeJavaBoolean: Decoder[java.lang.Boolean] = decodeBoolean.map(java.lang.Boolean.valueOf)
```
travisbrown: Is there a reason to make these lazy? There is some overhead, and since as far as I know all of these non-`java.time` Java types are fully supported by Scala.js, I'd rather not make this change where it's not necessary.

travisbrown:
Thanks, if it really is this easy I'll be thrilled!

modules/core/shared/src/main/scala/io/circe/Decoder.scala
japgolly: I was thinking that making these lazy means that Scala.JS can remove its `java.lang.*` shims from output if otherwise unused, resulting in a smaller output JS size. I can revert this though if you like, no prob.

japgolly:
> It would be great to have something in place to verify that it's possible to run some Circe operations on Scala.js with this change but without either java.time implementation.

I'm not sure how feasible that is.... Erm, what've you got in mind?

> Thanks, if it really is this easy I'll be thrilled!

You're welcome! I'm looking forward to the day when Scala.JS starts performing purity-analysis in its DCE consideration because then we wouldn't need changes like this. It's on the post 1.0 roadmap and could literally be years away though.

travisbrown:
> I'm not sure how feasible that is.... Erm, what've you got in mind?

I think this could be as simple as some tests in a module that depends on the circe-core module but not on scala-java-time or not-java-time.

Note that the Scala.js tests aren't run by CI by default (they take an incredibly long time, so I just do them before releases and for changes that look potentially problematic), so you'd have to run this manually.

japgolly:
Yeah good idea, that would work fine. Ok cool, I'll get that done in a few days - my eyes had enough computer tonight.

(And thanks, noted, re tests & CI.)

modules/core/shared/src/main/scala/io/circe/Decoder.scala
travisbrown: Hmm, we should verify that, but sounds reasonable. I'm merging this now, but will change the non `java.lang` ones back in a follow-up.

travisbrown:
Commits included in #1249.

PR #1257, opened 2019-09-13
~2x speed up in serialization of `java.time.*` instances by default encoders

plokhotnyuk:
Bellow are results of benchmarks which writes an array of 128 `java.time.*` instances as a JSON array of strings:

##Before
```
[info] Benchmark                            (size)   Mode  Cnt      Score      Error  Units
[info] ArrayOfLocalDateTimesWriting.circe      128  thrpt    5  17828.304 ± 2224.497  ops/s
[info] ArrayOfLocalTimesWriting.circe          128  thrpt    5  28257.429 ± 2637.814  ops/s
[info] ArrayOfOffsetDateTimesWriting.circe     128  thrpt    5  14812.471 ±  745.845  ops/s
[info] ArrayOfOffsetTimesWriting.circe         128  thrpt    5  21762.077 ± 2176.789  ops/s
[info] ArrayOfZonedDateTimesWriting.circe      128  thrpt    5  12016.890 ± 1583.515  ops/s
```
##After
```
[info] Benchmark                            (size)   Mode  Cnt      Score      Error  Units
[info] ArrayOfLocalDateTimesWriting.circe      128  thrpt    5  29355.954 ± 2920.523  ops/s
[info] ArrayOfLocalTimesWriting.circe          128  thrpt    5  58263.162 ± 8294.110  ops/s
[info] ArrayOfOffsetDateTimesWriting.circe     128  thrpt    5  25150.080 ± 2462.175  ops/s
[info] ArrayOfOffsetTimesWriting.circe         128  thrpt    5  41273.528 ± 7187.940  ops/s
[info] ArrayOfZonedDateTimesWriting.circe      128  thrpt    5  20173.697 ± 2366.075  ops/s
```

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Json.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Json.scala
@@ -198,7 +198,7 @@ sealed abstract class Json extends Product with Serializable {
   /**
    * Use implementations provided by case classes.
    */
-  override def hashCode(): Int
+  def hashCode(): Int
 
   // Alias for `findAllByKey`.
   final def \\(key: String): List[Json] = findAllByKey(key)
@@ -510,8 +510,7 @@ object Json {
    * The result is a JSON null if the argument cannot be represented as a JSON
    * number.
    */
-  final def fromFloatOrNull(value: Float): Json =
-    if (isReal(value)) JNumber(JsonFloat(value)) else Null
+  final def fromFloatOrNull(value: Float): Json = if (isReal(value)) JNumber(JsonFloat(value)) else Null
 
   /**
    * Create a `Json` value representing a JSON number or string from a `Double`.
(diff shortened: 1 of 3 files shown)


travisbrown:
Thanks much, @plokhotnyuk! Will take a closer look tomorrow.

modules/core/shared/src/main/scala/io/circe/Json.scala
```diff
@@ -198,7 +198,7 @@ sealed abstract class Json extends Product with Serializable {
   /**
    * Use implementations provided by case classes.
    */
-  override def hashCode(): Int
+  def hashCode(): Int
```
travisbrown: Why remove the `override`? IIRC having it here means if an implementer doesn't provide an implementation they'll get a compiler error, which is a nice check to have.

modules/core/shared/src/main/scala/io/circe/Json.scala
```diff
@@ -544,18 +543,14 @@ object Json {
   final def fromBigDecimal(value: BigDecimal): Json = JNumber(JsonBigDecimal(value.underlying))
 
   /**
-   * Calling `.isNaN` and `.isInfinity` directly on the value boxes; we
-   * explicitly avoid that here.
+   * Calling `.isFinite` directly on the value boxes; we explicitly avoid that here.
    */
-  private[this] def isReal(value: Double): Boolean =
-    (!java.lang.Double.isNaN(value)) && (!java.lang.Double.isInfinite(value))
+  private[this] def isReal(value: Double): Boolean = java.lang.Double.isFinite(value)
```
travisbrown: Good catch, thanks—now that we've dropped 2.11 this is definitely better.

modules/core/shared/src/main/scala/io/circe/Json.scala
plokhotnyuk: It is redundant at least in Scala 2.13:

```
$ scala
Welcome to Scala 2.13.0 (OpenJDK 64-Bit Server VM, Java 1.8.0_222).
Type in expressions for evaluation. Or try :help.

scala> abstract class A { def hashCode: Int }
defined class A

scala> class B extends A
             ^
       error: class B needs to be abstract. No implementation found in a subclass for deferred declaration
       def hashCode: Int (defined in class A)

scala> case class B() extends A
defined class B
```

plokhotnyuk:
Benchmarks above shows improvement for serialization only.

Parsing speed doesn't changed, but using of formatters was removed with hope that default `parse` functions will be revised in future versions of Java for better performance...

travisbrown:
Thanks @plokhotnyuk!

Issue #1271, opened 2019-09-21
`.handleErrorWith` on `decodeOption` breaks its handling of missing fields


LeifW:
Normally the decoder for `Option` sets missing fields to `None`.
If I want Optional fields to still succeed if they're mal-formatted, I can write a new `Decoder[Option]` in terms of the existing one by applying `.handleErrorWith` to it and returning `None`. Unfortunately, missing fields will now error instead of being set to `None`.

Circe 2.12.1, Scala 2.12.10.
Example:
```scala
import io.circe.Decoder.Result
import io.circe.{Decoder, Json}
import io.circe.generic.semiauto.deriveDecoder

case class Person(name: Option[String], age: Option[Int])
object Person {
  implicit def optionDec[A: Decoder]: Decoder[Option[A]] = Decoder.decodeOption[A].handleErrorWith { err =>
    println("Error parsing Optional field; setting to None: " + err)
    Decoder.const(None)
  }
  implicit val dec: Decoder[Person] = deriveDecoder[Person]
}

object Main {
  val badValue = Json.obj(
    "name" -> Json.fromString("Bob"),
    "age" -> Json.fromString("foo")
  )

  val missingField = Json.obj(
    "name" -> Json.fromString("Bob")
  )

  val nullField = Json.obj(
    "name" -> Json.fromString("Bob"),
    "age" -> Json.Null
  )


  def decodedBadValue: Result[Person] = Person.dec.decodeJson(badValue)
  def decodedMissingField: Result[Person] = Person.dec.decodeJson(missingField)
  def decodedNullField: Result[Person] = Person.dec.decodeJson(nullField)

  def main(args: Array[String]) = {
    println("bad value:     " + decodedBadValue)
    println("missing field: " + decodedMissingField)
    println("null field:    " + decodedNullField)
  }

}
```
The desired output is:
```
Error parsing Optional field; setting to None: DecodingFailure(Int, List(DownField(age)))
bad value:     Right(Person(Some(Bob),None))
missing field: Right(Person(Some(Bob),None))
null field:    Right(Person(Some(Bob),None))
```
The actual output is:
```
Error parsing Optional field; setting to None: DecodingFailure(Int, List(DownField(age)))
bad value:     Right(Person(Some(Bob),None))
missing field: Left(DecodingFailure(Attempt to decode value on failed cursor, List(DownField(age))))
null field:    Right(Person(Some(Bob),None))
```

LeifW:
[...]

PR #1272, opened 2019-09-21
Preserve FailedCursor handling when using or/either/handleErrorsWith

LeifW:
Don't just override apply, override tryApply and have it use the
underlying impl's tryApply.
Fixes https://github.com/circe/circe/issues/1271


travisbrown:
This looks reasonable to me, thanks. Would you want to add a failing test? I could also do it in a follow-up.

travisbrown:
I've got some failing tests lined up, will open an new PR shortly.

PR #1275, opened 2019-09-23
Allow different modes for deepMerge

halfvim:
When performing a deepMerge, allow user to specify a merge mode, which can be
- Default: treats Array as value, similar to Null, Boolean, String or Number
- Index: replace the elements in this JSON array with the elements in the argument JSON at corresponding position.
- Concat: concatenate the elements in this JSON array and the argument JSON array.

This PR handles issue #271

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Json.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Json.scala
@@ -1,8 +1,10 @@
 package io.circe
 
+import java.io.Serializable
+
 import cats.{ Eq, Show }
 import io.circe.numbers.BiggerDecimal
-import java.io.Serializable
+
 import scala.collection.mutable.ListBuffer
 
 /**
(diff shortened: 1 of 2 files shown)


modules/core/shared/src/main/scala/io/circe/Json.scala
```diff
@@ -228,6 +257,14 @@ sealed abstract class Json extends Product with Serializable {
 
 object Json {
 
+  /**
+   * Merge strategy used for deepMerge operation when merging two arrays.
+   */
+  object MergeMode extends Enumeration {
```
travisbrown: I think it'd be better to use the `sealed trait` + `case object` pattern here, for consistency.

halfvim: > I think it'd be better to use the `sealed trait` + `case object` pattern here, for consistency.

done.

halfvim:
not sure what exactly to be done for moving stuff to `io.circe.extras`, but i've changed enum to case objects.

JD557:
Sorry for commenting out of nowhere, but wouldn't it be better to have something like `def deepMerge(that: Json, mergeLeafs: (Json, Json) => Json = _._2)`?

The current implementation only cares about arrays, but there are other cases where setting a merging strategy can be useful (e.g. when merging two numbers, sum the result).

ShahOdin:
@nrthyrk Just curious why this was closed?

jcallin:
Yeah would love to have this merged

PR #1279, opened 2019-09-30
Add deepMerge to JsonObject

bastewart:
Also change `Json.deepMerge` to use `toIterable` rather than `toList` internally.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/JsonObject.scala
+++ b/modules/core/shared/src/main/scala/io/circe/JsonObject.scala
@@ -166,6 +166,21 @@ sealed abstract class JsonObject extends Serializable {
    */
   final def filterKeys(pred: String => Boolean): JsonObject = filter(field => pred(field._1))
 
+  /**
+   * Perform a deep merge of this JSON object with another JSON object.
+   *
+   * Objects are merged by key, values from the argument JSON take
+   * precedence over values from this JSON. Nested objects are
+   * recursed.
+   */
+  def deepMerge(that: JsonObject): JsonObject =
+    toList.foldLeft(that) {
+      case (acc, (key, value)) =>
+        that(key).fold(acc.add(key, value)) { r =>
+          acc.add(key, value.deepMerge(r))
+        }
+    }
+
   private[circe] def appendToFolder(folder: Printer.PrintingFolder): Unit
 
   /**
--- a/modules/core/shared/src/main/scala/io/circe/JsonObject.scala
+++ b/modules/core/shared/src/main/scala/io/circe/JsonObject.scala
@@ -174,7 +174,7 @@ sealed abstract class JsonObject extends Serializable {
    * recursed.
    */
   def deepMerge(that: JsonObject): JsonObject =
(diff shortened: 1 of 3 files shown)


modules/core/shared/src/main/scala/io/circe/JsonObject.scala
```diff
@@ -166,6 +166,21 @@ sealed abstract class JsonObject extends Serializable {
    */
   final def filterKeys(pred: String => Boolean): JsonObject = filter(field => pred(field._1))
 
+  /**
+   * Perform a deep merge of this JSON object with another JSON object.
+   *
+   * Objects are merged by key, values from the argument JSON take
+   * precedence over values from this JSON. Nested objects are
+   * recursed.
```
bastewart: Just thought I should maybe link to `Json.deepMerge` here to explain for the behaviour of values being merged?

modules/core/shared/src/main/scala/io/circe/JsonObject.scala
```diff
@@ -166,6 +166,21 @@ sealed abstract class JsonObject extends Serializable {
...
+   */
+  def deepMerge(that: JsonObject): JsonObject =
+    toList.foldLeft(that) {
```
travisbrown: One small thing: can you use `toIterable` here? It should behave identically but would avoid some allocations.

modules/core/shared/src/main/scala/io/circe/JsonObject.scala
travisbrown: Sounds good to me. I can never remember the exact syntax for linking methods in API docs—if you can't find it easily `[[Json]]` would definitely be fine.

bastewart: I think what I've added is the right syntax... At least IntelliJ knows what's going on!

travisbrown: Oh, that was easier than I thought!

modules/core/shared/src/main/scala/io/circe/JsonObject.scala
```diff
@@ -166,6 +166,23 @@ sealed abstract class JsonObject extends Serializable {
...
   final def filterKeys(pred: String => Boolean): JsonObject = filter(field => pred(field._1))
 
+  /**
+   * Perform a deep merge of this JSON object with another JSON object.
+   *
+   * Objects are merged by key, values from the argument JSON take
+   * precedence over values from this JSON. Nested objects are
+   * recursed.
+   *
+   * See [[Json.deepMerge]] for behavior of merging values that are not objects.
```
travisbrown: We don't need to do it in this PR, but Scaladoc actually has a `@see` tag for cases like this.

bastewart: Ah nice, wasn't aware of that as I've never actually ended up building any scaladoc docs myself...

travisbrown:
Thanks much @bastewart!

PR #1291, opened 2019-10-11
Make value of the implicit class private

implmnt:
Since circe has dropped support for scala 2.10, there is no reason to use public val's in AnyVal implicit classes. 
This should fix potential collisions with user defined fields (key, wrappedEncodeable)

[See](https://github.com/typelevel/cats/issues/2514) a similar issue in cats


travisbrown:
@implmnt Thanks for catching this! Will get it merged soon.

PR #1293, opened 2019-10-11
Fix #1292 by restoring using of formatters for serialization of LocalDateTime, LocalTime, OffsetDateTime, OffsetTime, ZonedDateTime

### LINKED ISSUE #1292: java8 LocalDateTime is dropping 00 seconds value when parsing date string
Currently, I'm experiencing an issue with LocalDateTime conversion.

I saw the latest changes in circe-0.12.2 here: [Commit](https://github.com/circe/circe/commit/631b299f175edf4aa6fbe4735ab34f588322ddbb
)
This change is leading to break everything related to LocalDateTime now in my company. 

I also read a similar problem on Stackoverflow and this is considered as a future. Here: [SO Answer](https://stackoverflow.com/questions/50786482/java-8-localdatetime-dropping-00-seconds-value-when-parsing-date-string-value-wi)

[You can see a live example here](https://scastie.scala-lang.org/Bunyod/ercAv3ewQHe7yJlHENfzIQ/25)

plokhotnyuk:


travisbrown:
Thanks for fixing this. You don't think we should give `LocalDate` the same treatment for consistency?

PR #1298, opened 2019-10-15
Adds decoders for numerics which doesn't coerce strings, use them to build lenient legacy decoders

### LINKED ISSUE #1297: Default `Decoder[T]` for numeric types coerces strings to numbers and there are no other alternatives
According to the JSON format grammar published at http://json.org, JSON numbers are not contained in strings:

image

Circe decodes numeric values contained in strings and that's super convenient in most of the cases, however there are some services which need to be non-lenient in this regard.

Even though I think Circe should keep coercing valid strings into numeric values, it is still useful to have non-lenient decoders for these living under `Decoder` object.

I am opening a PR just in case this idea is accepted by the Circe community and maintainers.

pfcoperez:
Closes https://github.com/circe/circe/issues/1297

This PR adds non string coercing numeric decoders to `Decoder` object umbrella, this helps users building non-lenient APIs to enforce the types while keeping the default behaviour.

This default behaviour is now provided by composing the non-coercing decoders.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -473,6 +473,28 @@ object Decoder
     final def fail(c: HCursor): Result[A] = Left(DecodingFailure(name, c.history))
   }
 
+  private[this] class NumberWithStringCoercion[A](name: String)(
+    numericTarget: JsonNumber => Option[A]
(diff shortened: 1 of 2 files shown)


travisbrown:
Thanks for the implementation! I think we'll need to take some time to gather feedback on this change, but I definitely appreciate that you've provided a concrete proposal.

hamnis:
Closed due to age.

Issue #1281, opened 2019-10-01
recommend increasing stack size for JVM (add `.sbtopts`)


mr-git:
A year ago I had the issue with circe automatic codec derivation (https://github.com/sbt/sbt/issues/4311) - at some point the stack trace becomes too small to make a compilation. Possibly, it happens due to specific way how Shapeless parses the structure of JSON and/or case classes.

It would be nice to mention in README, near required dependencies, that it is recommended to increase stack size for SBT when using auto-derived codecs in project.

I got mine project under control with `.sbtopts` file in root directory, which contains:
```scala
-J-Xss2M
```

Though in other cases, people allocate even more, like 258M in https://github.com/circe/circe-magnolia/blob/master/.sbtopts :)

travisbrown:
This is discussed briefly in [the "Warnings and known issues" section](https://circe.github.io/circe/codecs/known-issues.html):

> For large or deeply-nested case classes and sealed trait hierarchies, the generic derivation provided by the `generic` subproject may stack overflow during compilation, which will result in the derived encoders or decoders simply not being found. Increasing the stack size available to the compiler (e.g. with `sbt -J-Xss64m` if you’re using SBT) will help in many cases, but we have at least one report of a case where it doesn’t.

I'd happily review a PR adding more detail to the docs, either here or somewhere else where it would be useful.

mr-git:
Thank you! I missed that page somehow...
Maybe with https://github.com/circe/circe/pull/1301 it will be easier to discover.

Please feel free to ignore and close this issue and PR, if they do not make sense.

PR #1301, opened 2019-10-21
#1281

mr-git:
I am not sure if this is the best way, but it would have helped me a lot.
Sadly, I didn't find `known-issues` page, when I was setting up project, nor when I was looking for solution.


travisbrown:
Thanks @mr-git!

PR #1315, opened 2019-11-13
Fix bug in optics example

GrumpySage:
Changed ```json.asNumber``` to ```j.asNumber``` in Recursively modifying JSON. Now correctly converts the numbers to string.


travisbrown:
Good catch, thanks. I guess this is a good reminder that just compiling your docs isn't always enough.

PR #1323, opened 2019-11-28
List Criteo as a circe user.

warrenseine:
An article was written about the migration from Jackson: https://medium.com/criteo-labs/migrate-a-service-getting-200kqps-from-jackson-to-circe-a475b2718206


warrenseine:
Please review the article. I do believe it's a nice feedback for future users, but I'll be totally fine removing it if it is not deemed appropriate as part of the README document.

travisbrown:
@warrenseine The article looks great, thanks for sharing it!

PR #1349, opened 2019-12-21
Put find back into HCursor

mdedetrich:
As discussed at https://gitter.im/circe/circe?at=5dfa18d549314a1d459c0a1c this PR puts the `find` method back into `HCursor`.


travisbrown:
Thanks. We should have a test for this but I can do that later.

mdedetrich:
Should I just write a test straight into this PR?

travisbrown:
@mdedetrich Up to you! I'm not currently planning an 0.12.4 but could do it if there's demand—otherwise there's no rush on this.

mdedetrich:
Ah understood, I will try and dig up an older test (or write my own)

PR #1375, opened 2020-01-26
Dotty

travisbrown:
This includes a few changes that are only partly Dotty related—for example I was getting errors I didn't really want to investigate from the `AllInstances` / `AllSyntax` stack in the tests on Dotty, so I just switched to using imports, which I've been wanting to do for a while both here and [in Cats](https://github.com/typelevel/cats/issues/3262), anyway.

The most exciting thing here is the `derived` machinery, which means that users on Dotty get generic derivation with nothing but circe-core. Here's an example translated from [the current docs](https://circe.github.io/circe/codecs/adt.html):

```scala
import io.circe.Codec

sealed trait Event derives Codec

case class Foo(i: Int) extends Event derives Codec
case class Bar(s: String) extends Event derives Codec
case class Baz(c: Char) extends Event derives Codec
case class Qux(values: List[String]) extends Event derives Codec
```
And then:
```scala
scala> import io.circe.jawn.decode, io.circe.syntax._

scala> val events = List[Event](Qux(List("a", "b", "c")), Foo(101), Bar("xyz"), Baz('-'))
val events: List[Event] = List(Qux(List(a, b, c)), Foo(101), Bar(xyz), Baz(-))

scala> events.asJson.noSpaces
val res0: String = [{"Qux":{"values":["a","b","c"]}},{"Foo":{"i":101}},{"Bar":{"s":"xyz"}},{"Baz":{"c":"-"}}]

scala> decode[List[Event]](res0)
val res1: Either[io.circe.Error, List[Event]] = Right(List(Qux(List(a, b, c)), Foo(101), Bar(xyz), Baz(-)))
```

I'm impressed with how little overhead the generic representation brings. I've included a benchmark that compares the derived codecs and `Codec.forProductN`, and the derived ones are faster for both encoding and decoding. I've not tried to measure compile times yet, but playing around in the REPL feels less slow than with circe-generic. 

The current `derived` implementations are a mess, ~~and they don't support some basic things, like generic or recursive case classes,~~ [update: these are mostly fixed] but I think these should be pretty easy to add. I've also been experimenting with configuration (snake-case keys, type fields, etc.).


Sebruck:
Looks good! Does that mean one can only use this dotty way of generic derivations for classes/types under your control?

travisbrown:
@Sebruck Thanks! And for `derives Codec`, yes, that's Dotty-only (for now, at least—I think I've seen talk about backporting it to Scala 2), but you can still get the equivalent of `semiauto` by calling `derived` directly:

```scala
scala> case class Foo(i: Int, s: String)
// defined case class Foo

scala> import io.circe._, io.circe.syntax._

scala> implicit val fooCodec: Codec[Foo] = Codec.derived
val fooCodec: io.circe.Codec[Foo] = anon$1@21055e79

scala> Foo(123, "abc").asJson
val res0: io.circe.Json = {
  "i" : 123,
  "s" : "abc"
}
```

unoexperto:
@travisbrown Is this part of 0.13.0 ? [None of the codecs unfortunately work](https://gitter.im/circe/circe?at=5e6a8ff962a5aa6a2305df9c) for me with Dotty 0.22.0-RC1.

Issue #1121, opened 2019-05-06
Printer - Exclude empty array


leobenkel:
Hello,

Is it possible to exclude empty list from the output json string representation ? 

Same way we have `dropNullValues` ?

Thanks

travisbrown:
We're unlikely to support this at the printer level. It's pretty common (and at least recommended by some versions of the JSON spec, IIRC) to treat missing fields and null-valued fields as semantically equivalent, and while you sometimes see the same thing for `{}` and `[]`, it's much less common in my experience.

If someone wanted to add a convenience method for doing this at the `Json` level, I'd be happy to review.

PR #1399, opened 2020-02-20
Add dropEmptyValues for Json

droptheplot:
#1121
Since objects could be empty also I think that it would be more convenient to drop them as well.


leobenkel:
I would love this feature ! What is the status ?

travisbrown:
I think we can add this in 0.14, although it might be in an `extras` package (sorry about the delay—I've been putting off working through what this would look like). In any case I'm happy to merge this when green. I think it just needs an `sbt +scalafmtAll`.

travisbrown:
Thanks!

leobenkel:
Thanks !

Issue #1113, opened 2019-04-19
Example for value classes in docs


mat646:
Hi, 
I've recently tried to handle value classes in circe without nested structure. 
Only found issue https://github.com/circe/circe/issues/469 which lacks in full example.
I've figured it out, but as circe beginner I wish I found it in docs (e.g. in semi-automatic-derivation).

Would you mind accepting such docs pull request?

```scala
import io.circe.Encoder
import io.circe.syntax._
import io.circe.generic.extras.semiauto._
import io.circe.generic.extras.defaults._

implicit val UserEncoder: Encoder[User]         = deriveEncoder
implicit val UserIdEncoder: Encoder[UserId]     = deriveUnwrappedEncoder
implicit val UserNameEncoder: Encoder[UserName] = deriveUnwrappedEncoder

case class UserId(id: Int) extends AnyVal
case class UserName(name: String) extends AnyVal
case class User(userId: UserId, userName: UserName)

val usr = User(UserId(1446), UserName("User Name"))

usr.asJson
//  res0: io.circe.Json =
//  {
//    "userId" : 1446,
//    "userName" : "User Name"
//  }
```

GMadorell:
Thanks for putting this issue out there! For me it was the only way to find it in google 👍 .
Also, personally I would like for being able to automatically use this as the default option (maybe by using a different import?) for automatically derivating the codecs. In the project I'm currently working on 100% of the encoders and decoders are like this.

PR #1427, opened 2020-03-15
Added example for value classes

mtsokol:
Related to #1113 

Hi! 

After some time I thought about adding it to the docs.

What do you think about it?


travisbrown:
Thanks, @mtsokol, this looks good to me. I think you may need to wrap the value class definitions in an object to satisfy mdoc on 2.12, though?

mtsokol:
@travisbrown Thank you for a clue.

I've tried to solve it but wrapping value classes in `object` fails with same error: `value class may not be a member of another class`

I recreated it in Scastie (it always fails in Worksheet mode):
1. without object: https://scastie.scala-lang.org/eeEKCLZaRhehI4hiWqFwTA
2. wrapped in object: https://scastie.scala-lang.org/RzFmo8ihTLuysSPbKJDgDQ

I asked about it on gitter and got solution to switch off worksheet mode: https://gitter.im/scalacenter/scastie

Right now my guess it that `mdoc` executes snippets in some kind of worksheet mode.

So right now I don't have any ideas to solve it. What do you think about it?

PR #1432, opened 2020-03-27
More verbose error messages when decoding refined types

vitaliihonta:
Hi there! I've got an issue with error message verbosity when using circe with refined types.
This PR provides API to chose the verbosity level:
- default which (I suppose) doesn't break source compatibility 
- more verbose classTag based (both ScalaJS, JVM)
- the most verbose type tag based implementation (JVM only)

If it looks fine I'll provide some scaladoc =)

### DIFF
--- a/modules/refined/jvm/src/main/scala/io/circe/refined/info/impl/TypeTagImplicits.scala
+++ b/modules/refined/jvm/src/main/scala/io/circe/refined/info/impl/TypeTagImplicits.scala
@@ -1,0 +1,23 @@
+package io.circe.refined.info.impl
+
+import io.circe.refined.info.TypeInfo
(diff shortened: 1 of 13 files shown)


travisbrown:
Sorry about the delay in reviewing this—I haven't had much time to work on Circe lately.

My main concern here is that it sets a precedent for using runtime reflection for error messages, which is something I want to avoid in circe-core.

I'd be happy to accept it as a separate circe-refined-info module, and possibly even as optional machinery in circe-refined itself, but I don't think it should be used in the default instances. I know circe-refined isn't necessarily one of the more popular modules, but it still gets 35-40k Maven Central downloads per month, and I'm hesitant to introduce a new runtime reflection-based mechanism given even those numbers.

Thanks for the implementation, though—it looks very good to me. We just need to find the right place to put it.

PR #1438, opened 2020-04-09
Add imap and iemap to Codec

ybasket:
Adding an instance of cats.Invariant gives access to `imap`, further `iemap` is added on Codec itself for cases where "decoding" can fail. Handy for `AnyVal`s and similar wrapper types.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Codec.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Codec.scala
@@ -1,5 +1,6 @@
 package io.circe
 
+import cats.Invariant
 import cats.data.Validated
 
 /**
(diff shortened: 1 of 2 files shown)


travisbrown:
Sorry for the long delay in responding. I've been reluctant to add something like this, because it felt like a dangerous step down a slippery slope toward thinking of `Codec` as a type class, and not just a thing that makes defining `Encoder` and `Decoder` instances a little more convenient.

I've been thinking about it a little more, though, and I'm not sure it's that big a deal. I'm inclined to go ahead and merge this.

ybasket:
No problem. I know you're busy working on awesome stuff like dhallj, so I'm fine waiting :)

I definitely see your point and wouldn't be disappointed if the PR is declined for that - these methods can be useful in some cases, but it's also limited as you often don't have a "base" `Codec` at hand, so you don't get to use `imap`. 

There's some prior art in doobie, `Meta` is similar to `Codec` (with differences, for example it has primitive instances itself) and has `imap`: https://github.com/tpolecat/doobie/blob/master/modules/core/src/main/scala/doobie/util/meta/meta.scala

PR #1450, opened 2020-04-28
Update broken link in optics documentation

valydia:
The `JsonPath` link on this page https://circe.github.io/circe/optics.html return a `404`.
Update the link with the correct location


travisbrown:
Good catch, thanks!

PR #1451, opened 2020-04-29
Polymorphic Encoding

diesalbla:
This WIP Draft makes the `Encoder` trait polymorphic also on the representation type of Json.  

`circe` is built around the purely-functional `Json` data type, which represents syntax trees in the [Json **grammar](https://tools.ietf.org/html/rfc7159#page-4). An _Encoding_ is just a function from a data type `A` to a `Json` value.  `circe` then provides many implicit standard encoders for base data types, and utilities to derive encoders generically. 

Having a representation of Json as a AST data structure helps to query it, or modify it in some ways. However, an application may only use Json to generate an HTTP response. In the case of `http4s` with `circe`, the application may encodes an internal response into a `Json` object, only to immediately encode that `Json` value into a `Chunk[Byte]` that goes in the response. [code](https://github.com/http4s/http4s/blob/master/circe/src/main/scala/org/http4s/circe/CirceInstances.scala#L94-L95). In this case, it may be more efficient to directly encode the `A` value into the chuk, by-passing the `Json` value.

One solution to this problem, of course, is to use a library like [`jsoniter`](https://github.com/plokhotnyuk/jsoniter-scala), that directly generate the bytearray. However, this functionality can be incorporated into `circe` itself.

We can do this by making `Encoder[A]` polymorphic on the target representation, so we may directly define encoders from base types. Notably, most of the encoders provided by `circe`, as well as the derivation, are not tied to to the `Json` data type itself: they work by _mapping_ or _contramapping_ other encoders.


travisbrown:
Thanks for this proposal. My current plan is to publish a 1.0 release using the current model at some point over the next few months, and then to immediately start 2.0 series milestones to explore less AST-dependent approaches for both encoding and decoding. I'm not sure what that will look like, but I'm not convinced that the encoder or decoder type classes should be parameterised on the representation. In any case it's useful to have this as a concrete starting point on the encoder side to evaluate.

PR #1454, opened 2020-05-05
Docs: Fix link to spray

MasseGuillaume:


MasseGuillaume:
Also, could you publish circe-spray for 2.12?
https://search.maven.org/search?q=circe-spray

Thanks!

travisbrown:
Thanks, and right, we can't publish this for 2.12 because Spray does support 2.12 (and probably never will).

Issue #1342, opened 2019-12-18
Add laws for key codecs, check these for the existing instances


nigredo-tori:
We can make a `KeyCodecLaws` and `KeyCodecTests` analogous to, respectively, `CodecLaws` and `CodecTests`. There seems to only be one law (round-trip).

satorg:
Looks like this one can be closed as resolved.

PR #1472, opened 2020-05-25
Laws for key codecs

tmccarthy:
This PR adds laws testing for the `KeyEncoder` and `KeyDecoder` classes, as suggested by #1342. The only law added is the "round trip" law, although I have followed the lead of the [`CodecTests`](https://github.com/circe/circe/blob/6d6f70e/modules/testing/shared/src/main/scala/io/circe/testing/CodecTests.scala#L48) and added tests for serialisation too.

I have added checks for the `KeyCodecLaws` for all existing instances.


travisbrown:
Thanks, this is great!

PR #1477, opened 2020-06-02
Make `Decoder#validate` generators provide error accumulation.

pfcoperez:
This PR is aimed to make `validate` methods keep processing the input and report all potential errors when the `Decoder[T]` instances they generate are used via `decodeAccumulating`.

That means the comments stating:

```
   * Note that this condition is checked before decoding with the current	
   * decoder, and if it does not hold, decoding does not continue. This means	
   * that if you chain calls to this method, errors will not be accumulated	
   * (instead only the error of the last failing `validate` in the chain will be	
   * returned).
```

Could be removed and the _fail-fast_ behaviour would only apply when the resulting decoders are used via `apply`.

The motivation for this came for me writing almost this in parallel for https://github.com/circe/circe-generic-extras/issues/72 and thinking _oh well, this would make sense as part of the existing validation methods._

I thought about splitting `validate` methods into offering this and not but, if I understand it correctly: Isn't that the motivation behind having two decoding methods (`apply`, `decodeAccumulating`)?

**Note** I didn't open an issue for this but I can do it if required.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -207,12 +207,6 @@ trait Decoder[A] extends Serializable { self =>
   /**
    * Build a new instance that fails if the condition does not hold for the
    * input.
-   *
-   * Note that this condition is checked before decoding with the current
-   * decoder, and if it does not hold, decoding does not continue. This means
-   * that if you chain calls to this method, errors will not be accumulated
-   * (instead only the error of the last failing `validate` in the chain will be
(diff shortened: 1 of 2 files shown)


travisbrown:
Thanks! This looks reasonable to me—I'm trying to remember what the original motivation was. I'll look more closely asap but we'll definitely get this merged before the next release.

pfcoperez:
I think this is related: https://github.com/circe/circe/issues/907#issuecomment-406968725

travisbrown:
Thanks again!

PR #1478, opened 2020-06-06
Use different language for approved list of JSON keys.

kevinmeredith:


travisbrown:
👍, thank you for doing this, @kevinmeredith.

PR #1479, opened 2020-06-08
Add decoder for untagged ArraySeq

tmccarthy:
Currently, Circe does not implicitly resolve a decoder for the immutable [`ArraySeq`](https://www.scala-lang.org/api/current/scala/collection/immutable/ArraySeq.html) where a `ClassTag` is not available. The [`CollectionDecoders.decodeIterable`](https://github.com/circe/circe/blob/v0.13.0/modules/core/shared/src/main/scala-2.13+/io/circe/CollectionDecoders.scala#L28) decoder is resolved where a `ClassTag` is available. But `ArraySeq` has no implicit `Factory` without an implicit `ClassTag`, so that decoder is not implicitly resolved when none is available.

An example of code that does not currently compile is available in the [`decodeUntaggedArraySeq` method](https://github.com/circe/circe/pull/1479/files#diff-d3ec9f95b90a735f8f1698aa3d2a5238R18) in the Spec for this PR.

This PR adds a dedicated decoder for `ArraySeq` in the case where a `ClassTag` is present, and a lower-priority decoder that uses the [`ArraySeq.untagged`](https://www.scala-lang.org/api/current/scala/collection/immutable/ArraySeq$.html#untagged:scala.collection.SeqFactory[scala.collection.immutable.ArraySeq]) factory in the case where there is no `ClassTag`.

I probably should have raised an issue to discuss this solution first, so please feel free to ask for a complete rework.

### DIFF
--- a/modules/tests/shared/src/test/scala-2.13/io/circe/ArraySeqSuite.scala
+++ b/modules/tests/shared/src/test/scala-2.13/io/circe/ArraySeqSuite.scala
@@ -1,0 +1,44 @@
+package io.circe
+
+import cats.Eq
+import io.circe.tests.CirceSuite
+import io.circe.syntax.EncoderOps
+import cats.instances.list.catsKernelStdEqForList
+import cats.instances.int.catsKernelStdOrderForInt
+import cats.instances.string.catsKernelStdOrderForString
+import io.circe.testing.CodecTests
+
+import scala.collection.immutable.ArraySeq
+
+class ArraySeqSuite extends CirceSuite {
+
+  // TODO this can be removed once Cats 2.2.0 is used
+  implicit def eqForArraySeq[A: Eq]: Eq[ArraySeq[A]] = Eq.by(_.toList)
+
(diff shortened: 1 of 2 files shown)


modules/tests/shared/src/test/scala-2.13/io/circe/ArraySeqSuite.scala
```diff
@@ -0,0 +1,44 @@
...
+import io.circe.tests.CirceSuite
+import io.circe.syntax.EncoderOps
+import cats.instances.list.catsKernelStdEqForList
+import cats.instances.int.catsKernelStdOrderForInt
+import cats.instances.string.catsKernelStdOrderForString
+import io.circe.testing.CodecTests
+
+import scala.collection.immutable.ArraySeq
+
+class ArraySeqSuite extends CirceSuite {
```
tmccarthy, line 13: I'm open to feedback on the location of these tests. Most other decoders don't seem to get a dedicated suite, but these are version-specific and have a little complexity, so I thought it would be justified.

PR #1480, opened 2020-06-08
Codecs for tagged types

vincentdehaan:
This PR adds automatic codecs for tagged types, for example, used in case classes, for example:

```scala
trait Name
trait Age
case class Person(name: String @@ Name, age: Int @@ Age)
```

As Shapeless is always available in the generic module, this seems the appropriate place to implement this feature.


travisbrown:
I like this, but I'm not sure I'm convinced about the placement—it could also go in circe-shapes, which is much more minimal and seems to me more suitable, although I'm definitely happy to discuss the two options (circe-generic or circe-shapes). In either case I think it'd be better as its own import, rather than bundled with `io.circe.generic.auto`.

vincentdehaan:
Thanks for your comments! I do agree that circe-shapes would be a better place. Had not thought about that in the first place. I'll push a new commit tonight or tomorrow to move it.

However, I think we can just mix in the trait with the `shapes` packages object. Otherwise the import might be hard to find for end users.

travisbrown:
Looks good to me, thanks!

PR #1487, opened 2020-06-14
Remove GC options from .jvmopts

joroKr21:
* ConcMarkSweepGC is deprecated on newer JVMs
  * GraalVM doesn't support it
  * It clashes with locally defined `SBT_OPTS`
  * It's not the same as cats
  * Link is outdated

If it's important for CI, could add it in `SBT_OPTS` environment variable.


travisbrown:
Thanks, these were copy-pasted a long time ago and haven't caused any problems, so I haven't touched them.

PR #1495, opened 2020-06-22
Add method to resolve Codec when Encoder/Decoder in implicit scope

tmccarthy:
This PR adds `Codec.instance`, which resolves a `Codec` where there is an `Encoder` and a `Decoder` in scope. I do not believe there was a way to do this previously other than:

```scala
Codec.from(Decoder[A], Encoder[A])
```

I'm open to feedback on whether `instance` is a good name for this method. It doesn't feel entirely consistent with (for example) [`Decoder.instance`](https://github.com/circe/circe/blob/v0.14.0-M1/modules/core/shared/src/main/scala/io/circe/Decoder.scala#L493), which takes an argument. It's more consistent with `apply[A]`, but obviously that's already taken here.


ybasket:
I found myself several times in the need of this, would be cool to get it in. Code looks good to me as well (just needs a rebase)!

Lasering:
I don't see a problem overloading the apply method.

zarthross:
I personally don't like this in that in introduces a possible foot gun for users.  The problem I see is the temptation for people to try and write 
```scala
implicit def fooCodec: Codec[Foo] = Codec.instance[Foo]
```
 and not realizing they created a runtime infinite implicit loop. 

I also want to note, that using a Codec for anything but `derives Codec` or `io.circe.generic.semiauto.deriveCodec` is NOT intended:

https://github.com/circe/circe/blob/e794df62aeea04bc008dd37232963d9801b0bb9e/modules/core/shared/src/main/scala/io/circe/Codec.scala#L28-L29

@hamnis  Do you have additional thoughts on this?  Am I wrong here?

hamnis:
It is true that this is a potential footgun.

However, personally I have found it very useful whenever I need to wrap a type which I know has been previously derived or the type is part of the default implicit set of types.

We can note in the scaladoc that it might cause a runtime infinite loop if used incorrectly.

hamnis:
Another suggestion could be naming it `unsafeSummon` or something like that.

modules/core/shared/src/main/scala/io/circe/Codec.scala
```diff
@@ -55,6 +55,12 @@ trait Codec[A] extends Decoder[A] with Encoder[A] {
 object Codec extends ProductCodecs with ProductTypedCodecs with EnumerationCodecs with CodecDerivationRelaxed {
   def apply[A](implicit instance: Codec[A]): Codec[A] = instance
 
+  /**
+   * Summons a `Codec` where a `Decoder` and an `Encoder` are in implicit scope.
+   */
+  def instance[A](implicit decoder: Decoder[A], encoder: Encoder[A]): Codec[A] =
```
hamnis:
```suggestion
  def unsafeSummon[A](implicit decoder: Decoder[A], encoder: Encoder[A]): Codec[A] =
```

modules/core/shared/src/main/scala/io/circe/Codec.scala
```diff
@@ -55,6 +55,12 @@ trait Codec[A] extends Decoder[A] with Encoder[A] {
 object Codec extends ProductCodecs with ProductTypedCodecs with EnumerationCodecs with CodecDerivationRelaxed {
   def apply[A](implicit instance: Codec[A]): Codec[A] = instance
 
+  /**
+   * Summons a `Codec` where a `Decoder` and an `Encoder` are in implicit scope.
```
hamnis, line 59:
```suggestion
   * Summons a `Codec` where a `Decoder` and an `Encoder` are in implicit scope.
   * This can cause a runtime infinite loop if you use it and the encoder and decoder is not in implicit scope, and this causes them to become in implicit scope.
```

modules/tests/shared/src/test/scala/io/circe/CodecSuites.scala
```diff
@@ -209,9 +209,9 @@ class CirceCodecSuite extends CirceMunitSuite {
 }
 
 class InvariantCodecSuite extends CirceMunitSuite {
-  val wubCodec = Codec.from(Decoder[Long], Encoder[Long]).imap(Wub(_))(_.x)
-  val wubCodecE = Codec.from(Decoder[Long], Encoder[Long]).iemap(l => Right(Wub(l)))(_.x)
-  val wubCodecT = Codec.from(Decoder[Long], Encoder[Long]).iemapTry(l => Success(Wub(l)))(_.x)
+  val wubCodec = Codec.instance[Long].imap(Wub(_))(_.x)
```
hamnis:
```suggestion
  val wubCodec = Codec.unsafeSummon[Long].imap(Wub(_))(_.x)
```

modules/tests/shared/src/test/scala/io/circe/CodecSuites.scala
```diff
@@ -209,9 +209,9 @@ class CirceCodecSuite extends CirceMunitSuite {
...
+  val wubCodecE = Codec.instance[Long].iemap(l => Right(Wub(l)))(_.x)
```
hamnis:
```suggestion
  val wubCodecE = Codec.unsafeSummon[Long].iemap(l => Right(Wub(l)))(_.x)
```

modules/tests/shared/src/test/scala/io/circe/CodecSuites.scala
```diff
@@ -209,9 +209,9 @@ class CirceCodecSuite extends CirceMunitSuite {
...
+  val wubCodecT = Codec.instance[Long].iemapTry(l => Success(Wub(l)))(_.x)
```
hamnis:
```suggestion
  val wubCodecT = Codec.unsafeSummon[Long].iemapTry(l => Success(Wub(l)))(_.x)
```

hamnis:
Closing in favour of #2348

PR #1502, opened 2020-07-09
Adding missing operations in the JawnParser

BalmungSan:
This is a simple PR which adds a couple of extra operations to the JawnParser.

-----

# Edit.

I am not sure if I should or not add tests for the new methods.
For what I see [here](https://github.com/circe/circe/blob/master/modules/jawn/src/test/scala/io/circe/jawn/JawnParserSuite.scala) it didn't include tests for the file and bytes methods since those would mostly be redundant; some would also require creating some json files in the resources folder.


travisbrown:
Thanks, this looks reasonable to me.

PR #1512, opened 2020-08-04
Fix taggedBooleanCodec

vincentdehaan:
I discovered a typo in my previous PR https://github.com/circe/circe/pull/1480


travisbrown:
Thanks for catching this. It makes me kind of uncomfortable about the `asInstanceOf` cast, which I didn't flag before. Couldn't `a.asInstanceOf[U]` be rewritten as a safe upcast (`(a: U)`)?

vincentdehaan:
I agree with you and pushed a new commit.

travisbrown:
Thanks!

PR #1537, opened 2020-09-16
Add Encoder.at

### LINKED ISSUE #1536: Encoder.at
Similarly to [Decoder.at](https://circe.github.io/circe/api/io/circe/Decoder.html#at(field:String):io.circe.Decoder[A]) there should also exist a method `at` in `Encoder`

```scala
trait Encoder[A] extends Serializable { self =>
  (...)
  final def at(field: String): Encoder[A] = mapJson(j => Json.obj(field -> j))
}
```

Lasering:
Closes #1536

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Encoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Encoder.scala
@@ -61,6 +61,12 @@ trait Encoder[A] extends Serializable { self =>
   final def mapJson(f: Json => Json): Encoder[A] = new Encoder[A] {
     final def apply(a: A): Json = f(self(a))
   }
+  
+  /**
+   * Create a new [[Encoder]] by creating a JsonObject with a single `field` with the output of this one.
+   */
+  final def at(field: String): Encoder.AsObject[A] =
+    Encoder.AsObject.instance(a => JsonObject.singleton(field, apply(a)))
 }
 
 /**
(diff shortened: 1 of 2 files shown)


travisbrown:
Thanks, looks good to me, but do you mind adding a test [here](https://github.com/circe/circe/blob/master/modules/tests/shared/src/test/scala/io/circe/EncoderSuite.scala)?

Lasering:
Added. Not sure if `m` should have a more complicated type.

Lasering:
Class Codec no longer compiles because it inherits final member `at` from both `Encoder` and `Decoder`, and JVM does not have reification so we cannot overload a method with different return types.

How do you want to solve this problem? Rename Encoder.at to something else? Suggestions:

- `enclosed`
- `inside`
- `atField`
- `atObject`

travisbrown:
Oh, right. Another alternative would be to add an `implicit x: Dummy` argument. Let's see if other people have preferences.

Lasering:
:+1: on the dummy implicit.

Lasering:
@travisbrown have you reached a decision on whether to use a different name or a dummy implicit? Meanwhile I changed my opinion. I think using a different, carefully chosen name is simpler and conveys the programmers objective better.

Lasering:
@travisbrown implemented the DummyImplicit

Lasering:
The approach of DummyImplicit fails if we want to implement an `at` for Codec (which would just return a `Codec.from(this.at(field), this.at(field))`)

PR #1545, opened 2020-10-01
Support for KeyCodec[Double]

an-tex:


travisbrown:
I guess I'm okay with this but it makes me kind of nervous (because of the general problems with mapping between floating point numbers and strings, and also using floating point numbers as map keys).

an-tex:
Agree!

Lasering:
This would be really helpful.

zarthross (approved):

PR #1552, opened 2020-10-16
Link directly to `io.circe` package in API documentation

x3ro:
This is, in my opinion, a small UX improvement over linking to the
root package, as it saves a couple of clicks that one always needs
to make when using the `Documentation` link on the microsite.

In addition I took the liberty of changing "Documentation" to
"API documentation" in order to make clearer what's being linked to
(this was inspired by the [Cats documentation](https://typelevel.org/cats/))

This is the API documentation start package after the change:

Screenshot 2020-10-17 at 01 14 26


x3ro:
I'm not entirely sure why this build fails, can you help me out 🤔

travisbrown:
This looks reasonable to me, thanks! I think running `sbt scalafmt` should fix the CI failure.

x3ro:
Mmh, unfortunately it still complains about lines in the `packages.scala` comment being too long. There doesn't seem to be a way to make them shorter than 120 characters though, or at least not without e.g. link shorteners which doesn't seem like a good idea :D Any ideas?

travisbrown:
Hmm, I think changing the Scalafmt config to accommodate the longer comment is the best choice, then.

x3ro:
I didn't find a way to disable this limitation for comments specifically, at least not [here](http://www.scalastyle.org/rules-1.0.0.html) so I just disabled it locally for the comment.

x3ro:
The build is passing now :) What do you think, can this be merged or is more work needed?

travisbrown:
Merging now, thanks again. Will try to get around to publishing this weekend.

x3ro:
Thanks :) I'll potentially have time to submit another 1 or 2 PRs that touch the docs this weekend, so (in case they get accepted) maybe they can be published all in one go 🎉

PR #1558, opened 2020-10-24
Document how to update the documentation in contributing guide

x3ro:
Since I've been doing some work on the documentation, and did not find a place where "how to generate the docs" is documented, I'm adding it to the contribution guide.

Let me know if this make sense or if it should be put elsewhere :)


travisbrown:
Thanks, this is great!

PR #1560, opened 2020-10-27
WIP: Add a more elaborate example to quickstart guide

x3ro:
I recently started using Circe, and while I was able to piece together usage for my use case, it would've been nice to have a more elaborate example in the documentation, something that brings together the features that are explained in the sub-sections of the docs. 

I don't expect this to be mergeable right away, more as a starting point to discuss what such an example could look like and would have to include.

### DIFF
--- a/docs/src/main/tut/quickstart.md
+++ b/docs/src/main/tut/quickstart.md
@@ -4,7 +4,7 @@ circe is published to [Maven Central][maven-central] and cross-built for Scala 2
 so you can just add the following to your build:
 
 ```scala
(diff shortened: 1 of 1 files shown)


x3ro:
Even though it says `WIP`, feedback is very welcome though :) Also, I know this is a bit of and odd request, but in order to meet my Hacktoberfest quota it'd be positively wonderful if you could add a `hacktoberfest-accepted` label to this PR, since I don't expect it to be merged until the month is over 😨

PR #1578, opened 2020-11-13
Add Instances ISO 4217 Currency Values

isomarcte:
This commit adds an Encoder and Decoder instance for ISO 4217 currency values, as modeled by the JRE as `java.util.Currency`.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -13,6 +13,7 @@ import cats.data.{
   Validated,
   ValidatedNel
 }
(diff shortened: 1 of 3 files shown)


travisbrown:
Thanks. I don't think I've ever actually used this type, but this seems reasonable. I'll take a closer look as soon as I can.

isomarcte:
@travisbrown I'm just circling through my old open PRs. Anything I can do to move this guy forward?

travisbrown:
@isomarcte Sorry about the delay—was meaning to get back to this before 0.14.0 but then dropped the ball. I think we can go ahead and merge it.

PR #1587, opened 2020-11-25
Add github actions

erwan:
Add Github Action.

Not removing Travis CI just yet, let's see how GH actions goes before.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,105 @@
+name: CI
+
+on:
+  push:
+    branches:
+        - master
+  pull_request:
+    branches:
+        - master
+
+jobs:
+  main:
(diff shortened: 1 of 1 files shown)


travisbrown:
Thanks! I'll take a closer look at this tomorrow—at a glance I don't see anything specific on the admin side that needs to be done for this to run. I guess we could just merge it and see if that works, if you think it's ready enough to try that, @erwan.

erwan:
@travisbrown I've opened a PR on my fork and it seems to be working: https://github.com/erwan/circe/pull/1 I'm a bit surprised the codecov publication worked considering I didn't give it any secret 🤷  not sure it published anything.

I think it's good to merge, I haven't removed travis-ci from it anyway so it shouldn't break anything.

PR #1591, opened 2020-11-28
Update Jawn to 1.0.2

travisbrown:


rossabaker:
Note: jawn-1.0.2 is incompatible with circe-literal-0.13.0.  It passed MiMa, but I still should have at least bumped the minor version when I released that change to jawn.  Sorry. :disappointed:

```
scala> json"""{"oops": "sorry"}"""
       ^
       error: Invalid JSON in interpolated string, java.lang.reflect.InvocationTargetException
...
       Caused by: scala.MatchError: (add,[Ljava.lang.Class;@58390643,[Ljava.lang.Object;@29e32886) (of class scala.Tuple3)
       	at io.circe.literal.JsonLiteralMacros$ObjectContextHandler.$anonfun$invokeWithArgs$3(JsonLiteralMacros.scala:118)
       	at io.circe.literal.JsonLiteralMacros$Handler.invoke(JsonLiteralMacros.scala:46)
```

It works with these changes:

```
scala> json"""{"hooray": "it's fixed"}"""
res0: io.circe.Json =
{
  "hooray" : "it's fixed"
}
```

travisbrown:
Yeah, circe-literal is doing some pretty terrible stuff with reflection to avoid forcing Scala.js users to need extra configuration (also to avoid imposing a runtime Jawn dependency on JVM users, IIRC). We could probably revisit that at some point, but since this will always be a compile-time problem and Jawn doesn't change very often it's not a priority for me.

rossabaker:
Yeah, it's fine once the versions line up.  And if we could get a circe-0.13.1, then Scala Steward updates won't keep sending people to this.

PR #1592, opened 2020-11-28
Remove Travis CI

travisbrown:


travisbrown:
For some reason the Codecov comment isn't being posted, but the [report](https://codecov.io/gh/circe/circe/pull/1592) is being posted and is available from the list of successful checks, so I don't really mind.

Issue #1582, opened 2020-11-22
Migrate tests to MUnit


travisbrown:
I've been using [MUnit](https://github.com/scalameta/munit) in all of my new Scala projects and have been slowly migrating others from ScalaTest. Getting away from ScalaTest is something I've wanted to do for a long time, and [Cats](https://github.com/typelevel/cats/pull/3538) for example has also chosen MUnit as a replacement.

One of the primary motivations for MUnit is that it tracks Dotty / Scala 3 releases more closely than ScalaTest, but it's also more closely aligned with the style and principles of Circe, and some of its features (like clues) are really nice.

This migration doesn't have to be done for all at once for all modules in this repository, or even for all tests within a module.

As far as I'm concerned, all other projects in the Circe organization would also benefit from this migration, but from my perspective the following are the highest priority:

* [circe-generic-extras](https://github.com/circe/circe-generic-extras)
* [circe-derivation](https://github.com/circe/circe-derivation)
* [circe-jackson](https://github.com/circe/circe-jackson)

PR #1595, opened 2020-11-28
Refactor benchmark to use MUnit

joe-warren:
Refactor tests in the `benchmark` project to use MUnit

See https://github.com/circe/circe/issues/1582


modules/benchmark/src/test/scala/io/circe/benchmark/FoldingBenchmarkSpec.scala
```diff
@@ -1,19 +1,20 @@
 package io.circe.benchmark
 
-import org.scalatest.flatspec.AnyFlatSpec
+import munit.FunSuite
```
joe-warren, line 3: I'm not sure if `FunSuite` is the best thing to use here, but it works

travisbrown:
I think you might just need to run `scalafmtAll` from sbt? Otherwise looks good to me.

travisbrown:
Excellent, looks good to me!

Issue #1519, opened 2020-08-19
Add Encoder/Decoder for `Nothing`


LukaJCB:
Ran into a problem with an ADT with multiple type parameters that was due to there being no `Nothing` codec in circe.  This may be controversial, but here's a motivating example:

```scala
sealed trait ADT[+A, +B]

case class Foo[+A, +B](val a: A, val b: B) extends ADT[A, B]
case class Bar[+B](val b: B) extends ADT[Nothing, B]

implicit def adtEncoder[A: Encoder, B: Encoder]: Encoder[ADT[A, B]] = ...
implicit def adtDecoder[A: Decoder, B: Decoder]: Decoder[ADT[A, B]] = ...
```

Now if I create a `Bar`, I don't get to have a codec for it. I know I could introduce a second type parameter for `Bar` to mitigate this, but assume it comes from a third party library.

hardmettle:
@travisbrown picking this up as part of LSUG meetup.

PR #1598, opened 2020-11-29
Add codec for Nothing

hardmettle:
Changes part of https://github.com/circe/circe/issues/1519

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -616,12 +616,7 @@ object Decoder
    * @group Decoding
    */
   implicit final val decodeNothing: Decoder[Nothing] = new Decoder[Nothing] {
-    final def apply(c: HCursor): Result[Nothing] = c.value match {
-      case Json.JObject(obj) if obj.isEmpty => Right(())
-      case Json.JArray(arr) if arr.isEmpty  => Right(())
-      case other if other.isNull            => Right(())
-      case _                                => Left(DecodingFailure("Nothing", c.history))
-    }
+    final def apply(c: HCursor): Result[Nothing] = Left(DecodingFailure("Nothing", c.history))
   }
 
   /**
--- a/modules/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -615,6 +615,18 @@ object Decoder
   /**
    * @group Decoding
    */
+  implicit final val decodeNothing: Decoder[Nothing] = new Decoder[Nothing] {
+    final def apply(c: HCursor): Result[Nothing] = c.value match {
+      case Json.JObject(obj) if obj.isEmpty => Right(())
+      case Json.JArray(arr) if arr.isEmpty  => Right(())
+      case other if other.isNull            => Right(())
+      case _                                => Left(DecodingFailure("Nothing", c.history))
+    }
+  }
+
+  /**
+   * @group Decoding
+   */
   implicit final val decodeBoolean: Decoder[Boolean] = new Decoder[Boolean] {
     final def apply(c: HCursor): Result[Boolean] = c.value match {
       case Json.JBoolean(b) => Right(b)
--- a/modules/tests/shared/src/main/scala/io/circe/tests/examples/package.scala
+++ b/modules/tests/shared/src/main/scala/io/circe/tests/examples/package.scala
@@ -149,57 +149,37 @@ package examples {
   }
   sealed trait ADT[+A, +B]
 
-  case class ADTFoo(a: String, b: Int) extends ADT[String, Int]
+  case class ADTFoo[+A, +B](a: A, b: B) extends ADT[A, B]
 
-  object ADTFoo {
-    implicit val eqADTFoo: Eq[ADTFoo] = Eq.fromUniversalEquals
-    implicit val arbitraryADTFoo: Arbitrary[ADTFoo] = Arbitrary(
-      for {
-        a <- Arbitrary.arbitrary[String]
-        b <- Arbitrary.arbitrary[Int]
-      } yield ADTFoo(a, b)
-    )
-    val decodeADTFoo: Decoder[ADTFoo] = Decoder.forProduct2("a", "b")(ADTFoo.apply)
-    val encodeADTFoo: Encoder[ADTFoo] = Encoder.forProduct2("a", "b") {
-      case ADTFoo(a, b) => (a, b)
-    }
-  }
-
-  case class ADTBar(b: Int) extends ADT[Nothing, Int]
-
-  object ADTBar {
-    implicit val eqADTBar: Eq[ADTBar] = Eq.fromUniversalEquals
-    implicit val arbitraryADTBar: Arbitrary[ADTBar] = Arbitrary(
-      for {
-        b <- Arbitrary.arbitrary[Int]
-      } yield ADTBar(b)
-    )
-    val decodeADTBar: Decoder[ADTBar] = Decoder.forProduct1("b")(ADTBar.apply)
-    val encodeADTBar: Encoder[ADTBar] = Encoder.forProduct1("b") {
-      case ADTBar(b) => b
-    }
-  }
+  case class ADTBar[+B](b: B) extends ADT[Nothing, B]
 
   object ADT {
-    implicit val eqADT: Eq[ADT[String, Int]] = Eq.fromUniversalEquals
 
-    implicit val arbitraryADT: Arbitrary[ADT[String, Int]] = Arbitrary(
-      Gen.oneOf(
-        Arbitrary.arbitrary[ADTFoo],
-        Arbitrary.arbitrary[ADTBar]
-      )
-    )
-
-    val encodeADT: Encoder[ADT[String, Int]] = Encoder.instance {
-      case foo @ ADTFoo(_, _) => Json.obj("ADTFoo" -> ADTFoo.encodeADTFoo(foo))
-      case bar @ ADTBar(_)    => Json.obj("ADTBar" -> ADTBar.encodeADTBar(bar))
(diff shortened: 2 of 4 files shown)


travisbrown:
Thanks! A couple of points:

* I mentioned the `Unit` instances to point out where these should go, but the semantics are slightly different. The decoder, for example, will have to always fail (for the encoder the behavior doesn't really matter, so matching the unit instance seems reasonable to me).
* We should think about what a test looks like (probably just exercising the kind of use case @LukaJCB notes in #1519).

hardmettle:
@travisbrown made changes based on your comments and added tests for the same. The test is failing at the moment, so might need your help here.

travisbrown:
@hardmettle Thanks, I'll take a closer look this evening.

modules/core/shared/src/main/scala/io/circe/Decoder.scala
```diff
@@ -612,6 +612,15 @@ object Decoder
     }
   }
 
+  /**
+   * @group Decoding
+   */
+  implicit final val decodeNothing: Decoder[Nothing] = new Decoder[Nothing] {
+    final def apply(c: HCursor): Result[Nothing] = c.value match {
```
travisbrown, line 619: You can avoid some syntactic / runtime overhead here by just ignoring the cursor and returning the `Left` directly.

modules/tests/shared/src/main/scala/io/circe/tests/examples/package.scala
```diff
@@ -147,4 +147,62 @@ package examples {
...
+    )
+
+    val encodeADT: Encoder[ADT[String,Int]] = Encoder.instance {
+      case foo @ ADTFoo(_, _) => Json.obj("ADTFoo" -> ADTFoo.encodeADTFoo(foo))
+      case bar @ ADTBar(_)    => Json.obj("ADTBar" -> ADTBar.encodeADTBar(bar))
+    }
+
+    val decodeADT: Decoder[ADT[String,Int]] = Decoder.instance { c =>
+      c.keys.map(_.toVector) match {
+        case Some(Vector("ADTFoo")) => c.get("ADTFoo")(ADTBar.decodeADTBar.widen)
```
LukaJCB, line 201: Pretty sure this should use the `ADTFoo` decoder :)

LukaJCB (dismissed):
Looks good to me, only issue seems to be test coverage of the actual implementation which can't be called reasonably anyway, so it's good from my side 👍  thanks!

travisbrown:
I'm sorry, I haven't had time to take a closer look, but this test compiles without the `Nothing` instances, right? I want to be sure we're properly actually exercising the fact that they exist, even though they're never actually used.

hardmettle:
@travisbrown @LukaJCB The example has `case class ADTBar(b: Int) extends ADT[Nothing, Int]` , do want it to be explicit in any other way ?

travisbrown:
I'm unlikely to be able to spend more time with this until the weekend, but the basic idea is that the test for this code should be something that didn't compile before it was introduced.

travisbrown:
The goal would be to be able to have a test like this:

```scala
checkAll("Codec[ADT[Nothing, Int]]", CodecTests[ADT[Nothing, Int]].codec)
```

…which wouldn't have compiled without the `Nothing` instances.

Unfortunately this doesn't seem to work even with them, for some reason I haven't had a chance to look at more closely. It could very well be some Scala bug or other nonsense.

Alternatively you could have a couple of tests like this:

```scala
def decodeTagged[A: Decoder, B: Decoder]: Decoder[Either[A, B]] = Decoder[A].at("L").either(Decoder[B].at("R"))

assertEquals(io.circe.jawn.decode("""{"L": 1}""")(decodeTagged[Int, Nothing]), Right(Left(1)))
```
Which again wouldn't have compiled without these instances, but has the advantage of actually working now.

That wouldn't resolve #1519 specifically (that would take more investigation into why the `Decoder[ADT[Nothing, Int]]` examples don't work), but it would be a reasonable thing to add.

hardmettle:
@travisbrown what I did locally was 
`def nothing:Nothing = throw new NotImplementedError`
and then 

```
implicit val arbitraryADTFoo: Arbitrary[ADTFoo[Nothing, Int]] = Arbitrary(
    for {
      b <- Arbitrary.arbitrary[Int]
    } yield ADTFoo(nothing,b)
  )
```
This `checkAll("Codec[ADT[Nothing, Int]]", CodecTests[ADT[Nothing, Int]].codec)`  is giving 
```
 diverging implicit expansion for type io.circe.Decoder[A]
[error] starting with value decodeInt in object Decoder
[error]   checkAll("Codec[ADT[Nothing, Int]]", CodecTests[ADT[Nothing, Int]].codec)
```

travisbrown:
@hardmettle Right, that shows the issue, although we wouldn't want to use an arbitrary instance that generates `Nothing` values in the tests.

The problem is that the `Decoder[ADT[Nothing, Int]]` instance isn't resolved, even though there's a `Decoder[Nothing]` in scope and `ADT.adtDecoder[Nothing, Int]` compiles just fine. This seems like a Scala bug, but I don't have any idea what's going on, and am not likely to have time to take a closer look.

There are two ways we could go from here: you could add a test with something like my `decodeTagged` example above, where these instances do provide value in a way that works, and we could merge it but not close #1519, since it still doesn't support that particular use case. Alternatively if someone wants to dig into why the `Decoder[ADT[Nothing, Int]]` instance isn't being resolved given the current implementation and example code in the tests, that'd be great—I'm just not going to be able to help.

kubukoz:
`Nothing` gets special treatment from the compiler, I'm not sure it could be called a bug (might actually be a feature, for all I know). We can use a workaround similar to what fs2 does with `Pure` - define a type alias for a subtype of Nothing and add a lower bound on the derivations:

```diff
   object ADT {
+    type Nope <: Nothing

-    implicit def adtEncoder[A: Encoder, B: Encoder]: Encoder[ADT[A, B]] = Encoder.instance {
+    implicit def adtEncoder[A >: Nope: Encoder, B: Encoder]: Encoder[ADT[A, B]] = Encoder.instance {
       case foo @ ADTFoo(_, _) =>
         Json.obj(
           "ADTFoo" -> Json.obj(
@@ -167,7 +168,7 @@ package examples {
         Json.obj("ADTBar" -> Json.obj("b" -> Json.fromString(Option(bar.b.toString).getOrElse(""))))
     }

-    implicit def adtDecoder[A: Decoder, B: Decoder]: Decoder[ADT[A, B]] = Decoder.instance { c =>
+    implicit def adtDecoder[A >: Nope: Decoder, B: Decoder]: Decoder[ADT[A, B]] = Decoder.instance { c =>
       c.keys.map(_.toVector) match {
         case Some(Vector("ADTFoo")) =>
           for {
```

Then the encoder/decoder instances get picked up correctly:

image

kubukoz:
I remember @smarter explained how that part of type inference works in Scala 2 and how it changes in Scala 3 in this talk: https://www.youtube.com/watch?v=lMvOykNQ4zs

kubukoz:
There's one remaining problem with the Arbitrary instance (there's no `Arbitrary[Nothing]`), but I think we can pass a `Gen.fail` for it, so that the ADTFoo branch will effectively be ignored if Nothing is used as its left type. Here's the code I used to get it running (`PropertyCheckConfiguration(minSuccessful = 1000000)` didn't give me any failures):

```diff
diff --git a/modules/tests/shared/src/test/scala/io/circe/CodecSuites.scala b/modules/tests/shared/src/test/scala/io/circe/CodecSuites.scala
index defbf7e5..5b15148e 100644
--- a/modules/tests/shared/src/test/scala/io/circe/CodecSuites.scala
+++ b/modules/tests/shared/src/test/scala/io/circe/CodecSuites.scala
@@ -183,28 +183,31 @@ class EitherCodecSuite extends CirceSuite {
 
 class NothingCodecSuite extends CirceSuite {
 
-  implicit val eqADTFoo: Eq[ADTFoo[String, Int]] = Eq.fromUniversalEquals
-  implicit val arbitraryADTFoo: Arbitrary[ADTFoo[String, Int]] = Arbitrary(
+  implicit def eqADTFoo[A]: Eq[ADTFoo[A, Int]] = Eq.fromUniversalEquals
+  implicit def arbitraryADTFoo[A: Arbitrary]: Arbitrary[ADTFoo[A, Int]] = Arbitrary(
     for {
-      a <- Arbitrary.arbitrary[String]
+      a <- Arbitrary.arbitrary[A]
       b <- Arbitrary.arbitrary[Int]
     } yield ADTFoo(a, b)
   )
+
   implicit val eqADTBar: Eq[ADTBar[Int]] = Eq.fromUniversalEquals
   implicit val arbitraryADTBar: Arbitrary[ADTBar[Int]] = Arbitrary(
     for {
       b <- Arbitrary.arbitrary[Int]
     } yield ADTBar(b)
   )
-  implicit val eqADT: Eq[ADT[String, Int]] = Eq.fromUniversalEquals
+  implicit def eqADT[A]: Eq[ADT[A, Int]] = Eq.fromUniversalEquals
 
-  implicit val arbitraryADT: Arbitrary[ADT[String, Int]] = Arbitrary(
+  implicit val arbNothing: Arbitrary[Nothing] = Arbitrary(Gen.fail)
+  implicit val arbitraryADT: Arbitrary[ADT[Nothing, Int]] = Arbitrary(
     Gen.oneOf(
-      Arbitrary.arbitrary[ADTFoo[String, Int]],
+      arbitraryADTFoo[Nothing](arbNothing).arbitrary,
       Arbitrary.arbitrary[ADTBar[Int]]
     )
   )
-  checkAll("Codec[ADT[String, Int]]", CodecTests[ADT[String, Int]].codec)
+
+  checkAll("Codec[ADT[String, Int]]", CodecTests[ADT[Nothing, Int]].codec)
 
 }
 class ValidatedCodecSuite extends CirceSuite {
```

hardmettle:
@travisbrown your thoughts on @kubukoz 's suggestion? Can I close it with those changes ?

hamnis:
@hardmettle I think with the changes from @kubukoz this will work. Are you willing to make the required changes?

hamnis:
closing, will reopen if this was done in error.

PR #1617, opened 2020-12-20
Introduce reason for DecodingFailure, improve error messages

AL333Z:
This PR is an attempt to offer a way to interpret `DecodingFailure` outcomes and let the user build a human-readable error message.
I've been a user of circe for years and I never needed this feature before. Now I have a use-case in which a good error message is required (a public REST endpoint).

Basically, I'm introducing an explicit encoding of the reason to complement the message field in DecodingFailure:

```scala
  sealed trait Reason
  case object MissingField extends Reason
  case class WrongTypeExpectation(expectedJsonFieldType: String, jsonValue: Json) extends Reason // the JSON contained the field, but the field has the wrong JSON type
  case class CustomReason(message: String) extends Reason // a validation failed while decoding or a custom string error message
```

And a default implementation for the `message` field which should be good enough for most cases.

```scala
  def history: List[CursorOp]

  val message: String = reason match {
    case WrongTypeExpectation(expJsType, v) => s"Got value '${v.noSpaces}' with wrong type, expecting $expJsType"
    case MissingField                                        => "Missing required field"
    case CustomReason(message)                 => message
  }
  ...
}
```

I then updated all the instances available, so that the `DecodingFailure`s they return will be reflecting what I believe is the right `Reason`.

This predates mostly from these issues/PR, so most of the credits should go to the original authors/participants: https://github.com/circe/circe/issues/188, https://github.com/circe/circe/issues/588, https://github.com/circe/circe/pull/1081, https://github.com/circe/circe/issues/306.

I'd be more than open to any adjustment needed for a smoother migration or for a better encoding.
Hope you may find this useful :)

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -39,6 +39,9 @@ import java.time.{
 import java.time.format.DateTimeFormatter
 import java.util.Currency
 import java.util.UUID
+
+import io.circe.DecodingFailure.Reason.{ MissingField, WrongTypeExpectation }
+
 import scala.annotation.tailrec
 import scala.collection.immutable.{ Map => ImmutableMap, Set, SortedMap, SortedSet }
 import scala.collection.mutable.Builder
@@ -70,7 +73,7 @@ trait Decoder[A] extends Serializable { self =>
     case hc: HCursor => apply(hc)
     case _ =>
       Left(
-        DecodingFailure("Attempt to decode value on failed cursor", c.history)
+        cursorToDecodingFailure(c)
       )
   }
 
@@ -78,8 +81,20 @@ trait Decoder[A] extends Serializable { self =>
     case hc: HCursor => decodeAccumulating(hc)
     case _ =>
       Validated.invalidNel(
-        DecodingFailure("Attempt to decode value on failed cursor", c.history)
+        cursorToDecodingFailure(c)
       )
+  }
+
+  private[this] def cursorToDecodingFailure(cursor: ACursor) = {
+    val history = cursor.history
+    val historyToFailedCursor = history.takeWhile(_ != CursorOp.DeleteGoParent)
+    val field = CursorOp.opsToPath(historyToFailedCursor).replaceFirst("^\\.", "")
+    val down = cursor.downField(field)
+    if (down.succeeded) {
+      DecodingFailure(s"Couldn't decode $field", history)
+    } else {
+      DecodingFailure(MissingField, history)
+    }
   }
 
   /**
@@ -575,7 +590,7 @@ object Decoder
   implicit final val decodeJsonObject: Decoder[JsonObject] = new Decoder[JsonObject] {
     final def apply(c: HCursor): Result[JsonObject] = c.value.asObject match {
       case Some(v) => Right(v)
-      case None    => Left(DecodingFailure("JsonObject", c.history))
+      case None    => Left(DecodingFailure(WrongTypeExpectation("object", c.value), c.history))
     }
   }
 
@@ -585,7 +600,7 @@ object Decoder
   implicit final val decodeJsonNumber: Decoder[JsonNumber] = new Decoder[JsonNumber] {
     final def apply(c: HCursor): Result[JsonNumber] = c.value.asNumber match {
       case Some(v) => Right(v)
-      case None    => Left(DecodingFailure("JsonNumber", c.history))
+      case None    => Left(DecodingFailure(WrongTypeExpectation("number", c.value), c.history))
     }
   }
 
@@ -595,7 +610,7 @@ object Decoder
   implicit final val decodeString: Decoder[String] = new Decoder[String] {
     final def apply(c: HCursor): Result[String] = c.value match {
       case Json.JString(string) => Right(string)
-      case _                    => Left(DecodingFailure("String", c.history))
+      case json                 => Left(DecodingFailure(WrongTypeExpectation("string", json), c.history))
     }
   }
 
@@ -607,7 +622,8 @@ object Decoder
       case Json.JObject(obj) if obj.isEmpty => Right(())
       case Json.JArray(arr) if arr.isEmpty  => Right(())
       case other if other.isNull            => Right(())
-      case _                                => Left(DecodingFailure("Unit", c.history))
+      case json =>
+        Left(DecodingFailure(WrongTypeExpectation("'null' or '[]' or '{}'", json), c.history))
     }
   }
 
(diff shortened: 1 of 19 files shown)


travisbrown:
Thanks for doing this! It'd be great if something like this could make it into 0.14.0. I don't have a lot of time for Circe at the moment and am currently focusing on 0.13.1, but I'll review as soon as I can.

kell18:
Hi! Is there some news about this one? 
For me the main pain-point with Circe are error messages, it'd awesome to improve that!

modules/core/shared/src/main/scala/io/circe/Error.scala
```diff
@@ -96,6 +123,13 @@ object DecodingFailure {
     val path = CursorOp.opsToPath(failure.history)
     s"DecodingFailure at ${path}: ${failure.message}"
   }
+
+  sealed trait Reason
+  case class MissingField(targetType: String) extends Reason
+  case class UnsupportedType(targetType: String) extends Reason
+  case class WrongTypeExpectation(expectedJsonFieldType: String, jsonValue: Json, targetType: String) extends Reason
```
jatcwang, line 130: I'm wondering whether `expectedJsonFieldType` should be a `NonEmptyList[JsonType]` where JsonType is an enum.

Also instead of `targetType` which leaks implementation details, should we have a `message` instead? As a user of an API I probably have no idea what a `NonEmptyList` is (and nor should I need to know). A message like "The array must have at least 1 element" is more actionable.

Similar sentiments about IncorrectValue's `targetType` field. If we remove `targetType` field there we can remove IncorrectValue altogether and I don't think we're really losing any expressiveness here.

Some examples error cases I can think of:

* Wrong primitive type => `WrongTypeExpectation`
* Empty array for NonEmptyString => CustomReason
* `java.time.Period` where `start` cannot be after `end` => CustomReason
* Unrecognized discriminator when trying to decode to a ADT => CustomReason (`"unrecognized discriminator 'not_a_valid_one'. Possible values are ..."`)

TBH, I'm starting to doubt whether we need WrongTypeExpectation at all...

modules/core/shared/src/main/scala/io/circe/Error.scala
```diff
@@ -96,6 +123,13 @@ object DecodingFailure {
     val path = CursorOp.opsToPath(failure.history)
     s"DecodingFailure at ${path}: ${failure.message}"
   }
+
+  sealed trait Reason
+  case class MissingField(targetType: String) extends Reason
+  case class UnsupportedType(targetType: String) extends Reason
```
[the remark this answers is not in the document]
AL333Z: No, IIRC the only instance of `UnsupportedType` is in CNil decoder.
I'd be more than happy to get rid of it, but I find `WrongTypeExpectation` not to be the right place since I don't know what the json type for a CNil should be (I don't even know why we need a decoder for CNil :))

modules/core/shared/src/main/scala/io/circe/Error.scala
AL333Z: About targetType, I'd be 100% to just remove that. The only reason I added that was to keep a sort of continuity with the former error messages (e.g. `"[A]Option[A]"` when failing to decode an `Option[A]`, "string" for `String`, etc..).

About the other points:
-`WrongTypeExpectation` has been thought to hold the failures related to wrong primitive types in the JSON
- `IncorrectValue` to hold failures due to validations (e.g. empty list when a NEL is required, refinements, etc...)
- `CustomReason` to hold all the custom messages.

 I think that we can get rid of `targetType` and collapse `IncorrectValue` to `CustomReason` without losing any expressivity, but I think that `WrongTypeExpectation` with a proper list of expected `JsonType` may still be helpful.

That said, I'm happy to apply any amendments to this regardless :).

modules/core/shared/src/main/scala/io/circe/Error.scala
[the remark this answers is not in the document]
jatcwang: I feel like we don't if the derivation implicits work differently. @travisbrown might know exactly why, but if we provide a lower priority implicit for `H :+: CNil` (and not CNil by itself) it should still wrk right?

Either way, in reality CNil decoder's body will never be called so it doesn't matter what we put in it

modules/core/shared/src/main/scala/io/circe/Error.scala
AL333Z: Hi @jatcwang ,
I just removed `targetType` and collapsed `IncorrectValue` to `CustomReason` in 6f4a601.

Do you have any suggestion about the `JsonType` enum to show the expected json type? Maybe some representation we can reuse without redefining a new one?

jatcwang: Cool :) I don't think there's anything we can reuse for JsonType within the circe codebase, no. IDM leaving it as a String since all I really need is the ability to distinguish between missing or invalid fields. Let's see what others say.

I also still see UnsupportedType, are you removing it atm?

AL333Z:
Thanks @jatcwang, as a last thing I updated the hierarchy to be an abstract class + companion object.
I'm happy to address any further request for amendment.

kell18:
Hi, @vkostyukov @alexarchambault @travisbrown! Can somebody please have a look at the PR?

etspaceman:
+1 to this. One of our biggest issues with Circe today is that we have cryptic error messages that make it hard to determine issues. This PR solves it well.

modules/core/shared/src/main/scala/io/circe/Decoder.scala
```diff
@@ -962,23 +981,24 @@ object Decoder
...
     }
     final override def tryDecode(c: ACursor): Decoder.Result[None.type] = c match {
       case c: HCursor =>
         if (c.value.isNull) rightNone
-        else Left(DecodingFailure("None", c.history))
+        else Left(DecodingFailure(WrongTypeExpectation("null", c.value), c.history))
       case c: FailedCursor =>
-        if (!c.incorrectFocus) keyMissingNone else Left(DecodingFailure("None", c.history))
+        if (!c.incorrectFocus) keyMissingNone
+        else Left(DecodingFailure(MissingField, c.history))
```
AL333Z, line 992: I'm not sure about this one and the next one. Thoughts?

evbo:
@AL333Z especially for the use-case you're intending this for, how might line numbers fit into this? I've found a hacky way to implement line numbers:

https://github.com/circe/circe/issues/1464

AL333Z:
@evbo I'm definitely not against having something like that, but I won't be committing to that now, in the context of this PR which is already big enough.

etspaceman:
Do we plan to merge this soon? This PR is very useful and has been open for quite some time.

AL333Z:
I obviously don't know :)
What I can say is that every once in a while a rebase this against the master, so when/if the time comes.. it should be ready.

travisbrown:
I've been putting off making a real start on the 0.15.x series, but I should probably do that. Give me a few more days. In any case I promise it'll be merged by next Wednesday—if I don't have time to do more substantive thinking about this before then we can just merge it anyway and handle any changes in follow-up PRs. Thanks again.

etspaceman:
I think another helpful thing to do here would be to set the `getMessage` of the `Errors` class. Right now, if `Errors` is used, the stack-trace that is generated gives you a null error message. Instead we should be able to concatenate each Error in that list.

Issue #1540, opened 2020-09-21
Encoding/decoding `Option[Option[A]]` loses data (is not injective)


neko-kai:
Example (https://scastie.scala-lang.org/jpXxtlT1Rk6SERNsMqz3IQ):

```scala
import io.circe.generic.auto._
import io.circe.syntax._

case class Odd(a: Option[Option[Int]])

val js = Odd(Some(None)).asJson

js
// { "a" : null }: io.circe.Json

js.as[Odd]
// Odd(None)
```

Decoding `Odd(Some(None)).asJson` results in `Odd(None)`.

tg44:
What would be the expected behaviour here? Missing param makes the outer option None, null makes the inner? As I understand the json generator can be configured to miss out the Nones keys or make the properties null, so the serialization part and the serialize and deserialize will probably still lose data.

neko-kai:
> Missing param makes the outer option None, null makes the inner?

This strategy will fail as you nest more options: `Option[Option[Option[Option[Option[....`

> What would be the expected behaviour here?

Falling back to explicit `{ "Some": {...} }` & `{ "None": {} }` ADT encoding whenever there are nested options is the only thing that can reasonably work, I think, but I'm not sure how reasonable that is.

The problem is only that `Codec` is no longer injective in the presence of nested Options (`decode(encode(a)) != a`) which isn't a nice property, but this situation may be not seen often enough in the real world to not treat it as morally injective anyway.

erikvanoosten:
Expected behavior: _does not compile_.

With Circe's interpretation of Option, no type `Option[Option[A]]` can be be expressed in JSON. Therefore it should not be supported and therefore it should not even compile.

This could be achieved by adding a second implicit that only matches on `Option[Option[A]]` with the appropriate annotation to document that this type is not supported.

travisbrown:
Circe doesn't make any guarantee that every codec will round-trip all values through the JSON representation, even for codecs provided in implicit scope (although the only other examples I can think of off the top of my head that don't round-trip are `Double.PositiveInfinity` and friends).

In this case we probably could provide ambiguous implicits for `Option[Option[A]]`, so that anyone explicitly asking for an `Encoder` or `Decoder` for an `Option[Option[A]]` would get a compiler error, but this still wouldn't solve the problem, since `Option[Option[A]]` would still be fine in a context where you have an `Option[A]` / `A: Encoder` constraint and `A` is `Option[B]`.

[...]

PR #1619, opened 2020-12-20
no default codec for option of option, a partial solution for #1540

erikvanoosten:
Unfortunately `sbt +validateJVM` fails on an unrelated issue. 

This code is therefore just provided as a suggestion to be used in any way you wish.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -952,6 +952,22 @@ object Decoder
     }
   }
 
+  implicit final def decodeOptionOption[A](implicit d: Decoder[A]): Decoder[Option[Option[A]]] =
+    new Decoder[Option[Option[A]]] {
+      private val message = "Option[Option[_]] has no default Decoder. Please write a custom Decoder."
+
+      final def apply(c: HCursor): Result[Option[Option[A]]] = tryDecode(c)
+
+      final override def tryDecode(c: ACursor): Decoder.Result[Option[Option[A]]] =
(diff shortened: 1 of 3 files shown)


erikvanoosten:
`codecov/patch — 0.00% of diff hit (target 85.36%) `

Since the new code is written for the purpose of _not_ being used, it will be impossible to pass this check 😄

travisbrown:
Thanks, but this doesn't actually prevent compilation (it's not very well-documented, but generic-simple is an experimental 2.13-only module, and its tests shouldn't be used for testing circe-core changes).

For example, from this branch:

```scala
scala> import io.circe.{Decoder, Encoder}
import io.circe.{Decoder, Encoder}

scala> Decoder[Option[Option[Int]]]
val res0: io.circe.Decoder[Option[Option[Int]]] = io.circe.Decoder$$anon$40@4c74c8f7

scala> Encoder[Option[Option[Int]]]
val res1: io.circe.Encoder[Option[Option[Int]]] = io.circe.Encoder$$anonfun$encodeOptionOption$2@282b48bd
```

It would be possible to have these cases fail by defining two identical (and therefore ambiguous) implicits, but as I noted in #1540 I don't personally think the trade-offs are worth it, especially because of cases like generic `Option[A]` members in a case class, where you can still easily end up with `Option[Option[A]]` being treated as having instances.

If someone wants to add ambiguous instances behind a special import in circe-extras, I wouldn't be opposed, but I'm not convinced this belongs in circe-core.

erikvanoosten:
Its your call, you know the project best. Thanks for looking into it though.

PR #1622, opened 2020-12-22
Tests: migrate from ScalaTest to MUnit.

diesalbla:
Migrate several of the tests suites in `circe` from Scalatest to MUnit.  We introduce a `CirceMunitSuite` trait, similar to `CirceSuite`. We adapt most of the packages and tests suites but some are still pending to be migrated (preferably before 0.14 is released)


travisbrown:
Wow, thanks, this looks great!

PR #1639, opened 2021-01-15
Encode Long into string when long is larger than 2^53 - 1 or smaller than -(2^53 - 1)

exoego:
`number` in JavaScript/JSON may loose precision when it is larger than `2^53 - 1` or smaller than `-(2^53 - 1)`.
Therefore, if one want to have `int64`/`long` value in JSON without loosing precision, it is safe to store the value as `string` in JSON.

The default `Decoder[Long]` can decode both `number` and `string` into `Long`.
https://github.com/circe/circe/blob/582f35d1068b1b2f45df77851fb52c2554d0407f/modules/core/shared/src/main/scala/io/circe/Decoder.scala#L808-L821
However, the default `Encoder[Long]` seems always encode `Long` into `number`.
I think circe users get surprised because the encoded value loose precision silently.

After this PR, the default `Decoder[Long]` transforms `Long` into `number` only if the value is safe integer in JSON, otherwise transforms `Long` into `string` which can preserve precision.

I assume new behavior is "safer", but it could be a breaking change.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Encoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Encoder.scala
@@ -266,7 +266,16 @@ object Encoder
    * @group Encoding
    */
   implicit final val encodeLong: Encoder[Long] = new Encoder[Long] {
-    final def apply(a: Long): Json = Json.fromLong(a)
+    // 2^53 - 1
+    private final val MAX_SAFE_INTEGER: Long = 45035996273704950L
+
+    final def apply(a: Long): Json = {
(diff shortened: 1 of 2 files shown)


travisbrown:
Thanks for the PR. I'm inclined to think this belongs in circe-extras, since we generally try to stick to the JSON spec rather than accommodate platforms that impose further restrictions (like JavaScript). I'd be interested in hearing feedback from others, though.

travisbrown:
Also if we were to decide to go this route, we'd need to make the same change for the `BigInt` and `BigDecimal` encoders.

PR #1642, opened 2021-01-21
Drops documentation concerning Scala 2.10

froth:
Support of circe for 2.10 has been dropped since Sep 24, 2018


travisbrown:
Thanks!

PR #1648, opened 2021-01-31
Add Json.traverse* methods

sergeykolbasov:
It's similar to `Json.map*` transformations but with a drop of `F[_] : Applicative`

We mangle a lot with JSON AST, often in some monadic context. This collection of methods would come in handy.


zarthross:
Thanks for the PR.  Love the idea.  What do you think of changing the signature though to be a little more flexible.

From:
`def traverseBoolean[F[_]](f: Boolean => F[Boolean])(implicit F: Applicative[F]): F[Json]`

To:
`def traverseBoolean[F[_]](f: Boolean => F[Json])(implicit F: Applicative[F]): F[Json]`

I think this would allow more transformation within the monadic context.  This would make it more akin to the `with*` methods than the `map*` methods, but you get a greater flexibility.

hamnis:
I think this would a nice set of combinators to have, @sergeykolbasov are you willing to make the changes suggested by @zarthross ?

The boolean case is an example, it would be the same for all the new traverseX functions

hamnis:
mimaErrors are genuine. We do not have these in previous releases, so we need to add exclusions rules here.

hamnis (dismissed):

modules/core/shared/src/main/scala/io/circe/Json.scala
```diff
@@ -350,6 +365,14 @@ object Json {
     final def mapString(f: String => String): Json = this
     final def mapArray(f: Vector[Json] => Vector[Json]): Json = this
     final def mapObject(f: JsonObject => JsonObject): Json = this
+
+    final def traverseBoolean[F[_]](f: Boolean => F[Json])(implicit F: Applicative[F]): F[Json] =
+      F.map(f(value))(identity)
```
zarthross: I think we might be overly complicating the traverse methods here that are not just pure.  Here is my suggestion:

```suggestion
    final def traverseBoolean[F[_]](f: Boolean => F[Json])(implicit F: Applicative[F]): F[Json] = f(value)
```

This change would need to be applied to each of there other sum types here.

zarthross: Pushed a fix for this.

zarthross (approved):

PR #1650, opened 2021-02-03
Replace actions/setup-ruby with ruby/setup-ruby

sloshy:
See: https://github.com/typelevel/cats/pull/3767

`actions/setup-ruby` is being officially deprecated in favor of `ruby/setup-ruby` and building the microsite without specifying a version can possibly cause errors. This fixes both issues, to use the new action as well as set a stable version of ruby.


travisbrown:
Thanks much! This looks good to me.

PR #1683, opened 2021-02-27
Remove 0.12 Deprecations

diesalbla:
WE remove all the items that have been marked as deprecated since versions 0.12.0 and 0.12.3, from September and October 2019 resp. This include the following elements:

- The CursorOp move operations to move N items to the right or left,   (in a Json array) or to move the the first item in the array.
- The `ACursor` methods to do those same operations. 
- The `pretty` method for printing.
- The `AccumulatingDecoder` class.


travisbrown:
Looks good to me! Will merge when green.

travisbrown:
Looks good to me, thanks!

Issue #1680, opened 2021-02-27
Clean up build warnings


travisbrown:
There are currently a lot of build warnings, which fall into at least four categories:

* Stuff we should just fix right now, like unused imports and "Auto-application to `()`".
* Stuff that needs further investigation (like a new-in-2.13.5 inexhaustive match warning in Error.scala).
* Our own deprecations; 0.14.0 is not far off so we could go ahead and clean these up.
* Standard library deprecations in 2.13 (`Stream`, etc.); we probably can't get rid of these without splitting things out into version-specific code, which is likely not worth the trouble.

travisbrown:
Done in #1683 and #1685.

PR #1685, opened 2021-02-27
Decreased Build Warnings

nilsmahlstaedt:
Picked some low hanging fruits from #1680 

Removed most to all of the following warnings when using `sbt compile`:
- Auto-application to () is depricated
- Unused Imports
- non exhaustive pattern matches
- changed Streams to LazyLists where used internally
- changed Map.mapValues to Map.view.mapValues(.map where needed)

based on local testing this should decrease the number of build output lines with `[warn]` in them by around 40%

There are still a lot of build warnings coming the usage of streams in the api, these could not really be adjusted

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Error.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Error.scala
@@ -84,8 +84,8 @@ object DecodingFailure {
       DecodingFailure(sw.toString, ops)
   }
 
-  implicit final val eqDecodingFailure: Eq[DecodingFailure] = Eq.instance {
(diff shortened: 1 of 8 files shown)


travisbrown:
This looks good to me, but won't the use of `LazyList` fail to compile on 2.12?

travisbrown:
To validate the 2.12 build you can run `++2.12.12` from inside an sbt console and then `validateJVM`.

nilsmahlstaedt:
Yeah, you are right.
Did not think about that.

travisbrown:
Thanks again!

PR #1692, opened 2021-03-08
Migrate tests from ScalaTest to MUnit

diesalbla:
Add a "group" construct, copied from `fs2` tests.

The last suite pending to be migrated is `MemoizedPiecesSuite`, which uses the `ScalaFutures` trait from `scalatest`.


travisbrown:
Thanks! Will take a closer look tomorrow.

travisbrown:
I did these out of order but I'll resolve the conflict.

zmccoy:
@diesalbla Looks like io.circe.JsonObjectSuite is failing on 2.13.6 each time.

zmccoy:
It looks like the issue is when the fields List contains multiple pairs of the same key.  Our expected takes the first, but both of our implementations use the last K,V pair.  This is reproducible via a test like the below on main, but it's interesting that we don't hit this use case in our current `main` while we do here with the change over to munit.  Doing `findLast` works of course and represents how both of our from methods work, but I'm curious as to why the seeding hits this case on this PR compared to main.  

`  it should "tester" in { 
    val fields = List(("",io.circe.Json.True), ("", io.circe.Json.Null))
    val key = ""
    val result1 = JsonObject.fromIterable(fields)
    val result2 = JsonObject.fromFoldable(fields)
    val expected = fields.find(_._1 == key).map(_._2)

    println("ITER: " + result1)
    println("FOLD " + result2)
    println("Expected " + expected)
    assert(result1(key) === expected)
    assert(result2(key) === expected)
  }`

zarthross (changes requested):
Looks like the Scala3 tests aren't running in CI, but are failing to compile. I confirmed they are working on `main`.  We'll need to get those tests building/passing.  In the meantime, I'll do some work on getting Scala3 added to the CI.

zmccoy:
@zarthross With https://github.com/circe/circe/pull/1878 being in I restarted these tests which should run the Scala 3 tests :)

zmccoy (approved):

modules/tests/shared/src/test/scala/io/circe/JsonObjectSuite.scala
```diff
@@ -2,200 +2,237 @@ package io.circe
...
+      assertEquals(result2.nonEmpty, expected)
+    }
   }
 
-  "kleisli" should "find fields if they exist" in { (fields: List[(String, Json)], key: String) =>
-    val result1 = JsonObject.fromIterable(fields)
-    val result2 = JsonObject.fromFoldable(fields)
-    val expected = fields.find(_._1 == key).map(_._2)
+  property("kleisli should find fields if they exist") {
+    forAll { (fields: List[(String, Json)], key: String) =>
```
zmccoy, line 128: @zarthross This will have the same issues as the one you fixed where the generator isn't generating a ton of 'good' test cases.  Do you think just using that one here as well would be good?  One thing with that generator is that it doesn't create cases where it wouldn't exist at all, if we expect that base case to be here as well.

zarthross: > @zarthross This will have the same issues as the one you fixed where the generator isn't generating a ton of 'good' test cases. Do you think just using that one here as well would be good?

Yes, I think that's a good idea to use a similar generator.

> One thing with that generator is that it doesn't look like it will create cases where the key wouldn't exist at all in the list of (key,vals) since it's always appended onto the list, if we expect that base case to be here as well.

You maybe right... The generator I wrote might just be putting in `JNull` instead of it simply not existing.  It shouldn't be hard to fix that generator so that sometimes its `Null` and othertimes its completely missing.  I think that would make for a better test.

zmccoy: 35b30d7 adds in the tests to cover those edge cases and pulls out the generators for use in the tests.

zarthross (approved):

PR #1696, opened 2021-03-21
Docs - Update note about @JsonCodec use

LLCampos:
In the [documentation](https://circe.github.io/circe/codecs/semiauto-derivation.html#jsoncodec) it is written that, to use  `@JsonCodec` you need the Macro Paradise plugin.

But as of Scala 2.13.x, [the plugin has been included in the compiler](https://github.com/scalamacros/paradise#macro-paradise-plugin) and what you need is the `"-Ymacro-annotations"` flag.


travisbrown:
Thanks for catching this!

PR #1715, opened 2021-04-07
Migrate build.sbt to slash syntax.

aumann:
As discussed yesterday @travisbrown 

Straightforward, there should be no need for additional parentheses according to the official migration guide here:
https://www.scala-sbt.org/1.x/docs/Migrating-from-sbt-013x.html#Migrating+to+slash+syntax


travisbrown:
Thanks much, looks good to me!

PR #1716, opened 2021-04-07
remove sbt-dotty from build

SethTisue:
its functionality is built in to sbt 1.5. but some details are a little different


SethTisue:
it looks like CI only tests 2.12 and 2.13, but not 3.0?

travisbrown:
Thanks, and right, there are still enough issues with dependencies missing for Scala 3, etc., that I'm only enabling the Scala 3 build when publishing. If anyone wants to fix this that'd be great but it's not a priority for me.

PR #1726, opened 2021-04-20
Make None.type decoder consistent with the Option decoder

### LINKED ISSUE #1725: None.type decoder is inconsistent with the Option decoder
[With Circe 0.14.0-M5](https://scastie.scala-lang.org/WPVpHLc1Tpy3woTM5ZsG9Q):

```scala
import io.circe._

val n = Json.obj()

// Produces Right(None)
n.as(Decoder[Option[Int]].at("foo"))
// Produces Left
n.as(Decoder[None.type].at("foo"))
```
`Decoder[Option[Int]]` overrides `tryDecode` and `tryDecodeAccumulating` to handle `FailedCursor`, but `Decoder[None.type]` does not.

nigredo-tori:
Fixes #1725

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -910,11 +910,11 @@ object Decoder
     }
   }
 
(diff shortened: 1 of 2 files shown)


travisbrown:
Looks good to me, thanks!

PR #1749, opened 2021-05-14
Scala 3 Support

mpilquist:
Still waiting on:
- scala-java-time (https://github.com/cquiroz/scala-java-time/pull/286)
- ~circe-optics~ (+update told me this was needed but I don't think it really is)


travisbrown:
@mpilquist Thanks! I don't think we actually need scala-java-time for a milestone release, since it's only a test dependency, and we're not currently committing to running any tests for Scala 3 on Scala.js on the milestones. We'll need it before 0.14.0, but I doubt that'll happen in the next few weeks anyway.

PR #1753, opened 2021-05-17
Provide CirceCodecRefined trait

DenisNovac:
Sometimes IDEA removes imports with implicits as not used so it is useful to have trait with them. I made this trait by example from TapirCodecRefined: https://github.com/softwaremill/tapir/blob/master/integrations/refined/src/main/scala/sttp/tapir/codec/refined/TapirCodecRefined.scala


travisbrown:
Looks good to me, thanks! Will merge when green.

PR #1788, opened 2021-06-30
Derive Decoder and Codec under -Yexplicit-nulls

### LINKED ISSUE #1786: `derives Decoder` and `derives Codec.AsObject` fail with -Yexplicit-nulls compiler option
Scala version 3.0.0
Circe version 0.14.1

When the `-Yexplicit-nulls` compiler flag is enabled, `Decoder` derivations like the following fail to compile:
```scala
case class Foo(bar: String) derives io.circe.Decoder
```

I get the following as the first error:
```
[error] -- [E007] Type Mismatch Error: /Users/barnardb/some-project/src/main/scala/c.scala:1:45
[error] 1 |case class Foo(bar: String) derives io.circe.Decoder
[error]   |                                             ^
[error]   |                               Found:    Null
[error]   |                               Required: Left[io.circe.DecodingFailure, ?]
[error]   | This location contains code that was inlined from Derivation.scala:77```
```

As one might expect with the derivation for `Decoder` not working, the following `Codec.AsObject` derivation also fails:
```scala
case class Foo(bar: String) derives io.circe.Codec.AsObject
```

## Background

Scala 3 has a [`-Yexplicit-nulls` compiler option that enables explicit nulls](https://dotty.epfl.ch/docs/reference/other-new-features/explicit-nulls.html). When this flag is enabled, `AnyRef` types such a `String` no longer accept `null` as a value, requiring values to be explicitly marked as being in a union with the `Null` type in order to be nullable. E.g. `val s: String = null` fails to typecheck, but `val s: String | Null = null` does. This allows more safety around the use of nulls.

Note that if explicit nulls are not being used then the type `String | Null` is equivalent to `String`, so by using explicit `… | Null` types for the rare Scala types that are actually nullable we can produce code that works regardless of whether the `-Yexplicit-nulls` option is in use or not.

There is an [`unsafeNulls` language feature](https://dotty.epfl.ch/docs/reference/other-new-features/explicit-nulls.html#unsafenulls) that can be enabled in a particular scope by importing `scala.language.unsafeNulls` to suppress most of the effects of `unsafeNulls` in code that needs a bit of an escape hatch.

## Workaround

Here are some workarounds using `unsafeNulls` that can be used by user of `-Yunsafe-nulls` until this issue is addressed in Circe. The workarounds use `Codec.AsObject`, but they also work with the appropriate adjustments for `Decoder`.

The code compiles if we enable the `unsafeNulls` language feature by importing `scala.language.unsafeNulls`:
```scala
import scala.language.unsafeNulls

case class Foo(bar: String) derives io.circe.Codec.AsObject
```

For top-level declarations though, this prevents us from taking advantage of most of the effects of explicit nulls in the entire file.

If we're willing to repeat code for every type that derives the typeclass, we can move the import to the companion object and reduce the `unsafeNulls` scope to the companion object:
```scala
case class Foo(bar: String) derives io.circe.Codec.AsObject

object Foo {
  import scala.language.unsafeNulls
}
```

But in that case we may as well stop using `derives`, and declare the given typeclass instances manually, limiting the `unsafeNulls` scope to their definitions:
```scala
case class Foo(bar: String)

given io.circe.Codec.AsObject[Foo] = {
  import scala.language.unsafeNulls
  io.circe.Codec.AsObject.derived
}
```

Alternatively, we can define an extension to the trait, and have its derivation use `unsafeNulls`. This is what I'm currently doing in my codebase, because it lets me keep the `derives` style, most of the workaround lives in one place, and it should be easy to remove the shim later with a trivial find-and-replace:
```scala
/** Shim until `dervives Codec.AsObject` works with explicit nulls */
trait CodecAsObject[A] extends io.circe.Codec.AsObject[A]

object CodecAsObject {
  inline def derived[A: scala.deriving.Mirror.Of]: CodecAsObject[A] = {
    new CodecAsObject[A] {
      val derived = {
        import scala.language.unsafeNulls
        io.circe.Codec.AsObject.derived[A]
      }
      def encodeObject(a: A): io.circe.JsonObject = derived.encodeObject(a)
      def apply(c: io.circe.HCursor): io.circe.Decoder.Result[A] = derived.apply(c)
    }
  }
}

case class Foo(bar: String) derives CodecAsObject
```

barnardb:
Fixes #1786.

Since testing this requires a new compiler flag, I had to come up with some way to test it. I looked into setting this flag for the tests in the tests module, but this would have made a lot of code that is shared with Scala 2 stop compiling, including where JDK APIs return values that are never null but not annotated as such. I opted instead to add a dependency on scala3-staging and use the mechanism documented at [Runtime Multi-Stage Programming](https://dotty.epfl.ch/docs/reference/metaprogramming/staging.html) in the Scala 3 documentation to compile a code snippet at test run-time. One drawback of this approach is that if the test fails, it is not clear which line of the macro code is the offender.

Feedback is welcome.


rleibman:
@travisbrown.... could we have a review? it's been 13 months, please.

rleibman:
@zmccoy, @zarthross? Or who owns this repo now?

zarthross:
@rleibman Sorry, @zmccoy and I are working through PRs when we are able too.  I took a glance at this one and its probably fine, but before I give it a stamp, I'm going to need it rebased on the latest main so CI can check it..   @barnardb When you have time to rebase, it would be much appreciated.

lqhuang:
@barnardb Any progress here? I could do a favor if it's helpful.

build.sbt
```diff
@@ -535,6 +535,9 @@ lazy val testsBase = circeCrossModule("tests", mima = None)
       "org.typelevel" %%% "discipline-scalatest" % disciplineScalaTestVersion,
       "org.typelevel" %%% "discipline-munit" % disciplineMunitVersion
     ),
+    libraryDependencies ++= {
+      if (isScala3.value) Seq("org.scala-lang" %% "scala3-staging" % scalaVersion.value) else Nil
```
rossabaker, line 539: Is this only used for tests?  Can it be put into the `Test` config and avoid a transitive dependency?

If not, how stable is that dependency?  It looks like it depends on the full version of Scala?  Would the same circe-core_3 still work on Scala 3 of at least the same minor version as Circe was released with?

hamnis:
Closing due to seemingly lack of interest, Reopen if this not the case.

Issue #1063, opened 2019-01-16
Rounding error when parsing `Float` values


plokhotnyuk:
Here is an example of mismatch with the standard Java parser for floats:
```
scala> java.lang.Float.parseFloat("1.199999988079071")
res0: Float = 1.1999999

scala> io.circe.parser.decode[Float]("1.199999988079071")
res1: Either[io.circe.Error,Float] = Right(1.2)

scala> java.lang.Float.parseFloat("7.038531E-26")
res2: Float = 7.038531E-26

scala> io.circe.parser.decode[Float]("7.038531E-26")
res3: Either[io.circe.Error,Float] = Right(7.0385313E-26)
```

Affected are only numbers which in a string of base-10 representation exceed the precision of the `Float` type, except `7.038531E-26f`.

While almost all differences are ~1ulp (one mantissa bit in base-2 representation) but it can imply greatly on base-10 representation when converting back to a string and so it can affect subsequent actions based on that representation like validation, decimal rounding, etc.

It would be better to specify that rounding error in docs at least.

PR #1800, opened 2021-07-24
Derivation support for `transformNames`, `useDefaults`, `discriminator`, and `strictDecoding` in Scala 3

### LINKED ISSUE #1777: Discriminant configuration for Scala 3's `derives` syntax
Hi!

It'd be useful to have a way of configuring how `Encoder` works for ADTs in Scala 3, in particular with the `derives` syntax.

I have an ADT like this:
```scala
sealed trait Event derives Codec.AsObject

case class Foo(i: Int) extends Event
case class Qux(values: List[String]) extends Event

println(List[Event](Qux(List("a", "b", "c"))).asJson)
```
which produces a JSON like this:
```
[{"Qux":{"values":["a","b","c"]}}]
```
however, I'd like to have an output like this:
```
[{"values":["a","b","c"]}]
```
or like this:
```
[{ "type": "Qux", "values":["a","b","c"]}]
```

From my understanding, this is not possible with the current mechanism so my fallback solution is a custom  ` given Encoder` with a huge `match` for each of the statements. Unfortunately, this is pretty boilerplateish.

Are there any plans to support something like this for Scala 3 derivation?

Lasering:
Fixes #1777.

Code duplication has been refactored.

When decoding a sum type the previous code would emit a DecodingFailure if the json contained more than one key, the new code does not.

`useDefaults=true` only works if the compiler flag `-Yretain-trees` is present.

Usage:

- Using default configuration (names are not transformed, defaults are used, no discriminator):

```scala
case class Foo(sTop: String = "bar", i: Int, d: Double = Math.PI) derives Encoder.AsObject, Decoder
case class Foo(sTop: String = "bar", i: Int, d: Double = Math.PI) derives Codec.AsObject
```

- Using a global configuration for every derivation:
```scala
// Somewhere in the global implicit scope
given Configuration = Configuration.default.withSnakeCaseNames
// Then same as above
```

- Overriding the default/global configuration:

```scala
object Bar:
  given Configuration = Configuration.default.withDiscriminator("_type")
enum Bar derives Encoder.AsObject, Decoder:
  case Batata(s: String = "bar")
  case Cenoura(i: Int)
  case Nabo
```

or passing the configuration explicitly:

```scala
object Bar:
  given Encoder.AsObject[Bar] = Encoder.AsObject.derive(discriminator = Some("_type"))
  given Decoder[Bar] = Decoder.derived(using Configuration.default.withDiscriminator("_type"))
enum Bar:
  case Batata(s: String = "bar")
  case Cenoura(i: Int)
  case Nabo
```


Lasering:
Looking at [Pureconfig EnumConfigReader](https://github.com/pureconfig/pureconfig/blob/master/core/src/main/scala-3/pureconfig/generic/derivation/EnumConfigReaderDerivation.scala) it seems we can be cleaver when encoding enums whose cases are all just singletons. Maybe create the objects `EnumEncoder`, `EnumDecoder`, `EnumCodec` which just contain the `derived` method, they would encode to string and decode from string. And to make it more ergonomic would also allow transforming the strings.

Lasering:
Derivation for `EnumEncoder`, `EnumDecoder`, and `EnumCodec` works as follows:

```scala
package io.circe.derivation.EnumEncoder

enum Bar derives EnumEncoder:
  case Batata(s: String = "bar")
  case Cenoura
  case Nabo
```
Will fail compilation with error: `Enum Bar contains non singleton case Batata`. The same applies for the `EnumDecoder` and `EnumCodec`.

- Using default configuration (names are not transformed):

```scala
enum Bar derives EnumCodec:
  case Batata
  case Cenoura
  case Nabo
```

- Using a global configuration for every derivation:
```scala
// Somewhere in the global implicit scope
import io.circe.derivation.EnumConfiguration
given EnumConfiguration = EnumConfiguration.default.withEncodeTransformNames(_.toUppercase)
// Then same as above
```

- Overriding the default/global configuration:

```scala
object Bar:
  given EnumConfiguration = EnumConfiguration(identity, renaming.snakeCase)
enum Bar derives EnumEncoder, EnumDecoder:
  case Batata
  case Cenoura
  case Nabo
```

or passing the configuration explicitly:

```scala
object Bar:
  given Decoder[Bar] = EnumDecoder.derive(identity)
  given Encoder[Bar] = EnumEncoder.derive(renaming.snakeCase)
enum Bar:
  case Batata
  case Cenoura
  case Nabo
```

felixbr:
Hi,

First of all, thank you so much for working on this! ❤️
It's a big blocker for Scala 3 at my company and I wanted to address it myself but haven't found the time so far.

I think it's great to have manual configuration (via explicit arguments) as well as global configuration for the whole project. That said, I can see some usability problems, we should think about. In particular I'm worried if different parts of a project require different configurations, which is a commons use-case at my company.

The current solution in `circe-generic-extras` had the issue in the past where it was not obvious if an implicit configuration was used at all. See my previous issue and PR here: https://github.com/circe/circe/issues/1220

Back then we renamed `deriveEncoder` to `deriveConfiguredEncoder` to make it clear that an implicit configuration was used. This also doesn't compile if the `Configuration` was not in scope.

With your implementation here it seems like there will be no error at all if someone puts the given `Configuration` in the wrong scope (or forgets to import it). In practice this would mean that you'd have to write tests for everything just to be sure the right configuration was in scope.

In my mind it would be ideal to have something like
```scala
case class Foo(...) derives Codec.AsObject(explicitlyPassedConfiguration)
```
but I don't think Scala 3 allows this kind of syntax. So maybe we can go a similar route like in `circe-generic-extras` and at least have two types:
```scala
case class Foo(...) derives Codec.AsObject // doesn't use implicit Configuration

case class Bar(...) derives ConfiguredCodec.AsObject // requires implicit Configuration and doesn't compile if nothing is in scope
```

While this doesn't prevent someone from using the _wrong_ config, it at least makes sure you have something in scope and didn't forget. It also shows the reader of the code that this doesn't use the default derivation.

I also think it would be nice to be able to pass a `Configuration` object explicitly instead of having to pass all parameters one-by-one each time. It gets annoying if you have multiple case classes which share one config:
```scala
given Encoder.AsObject[Bar] = ConfiguredEncoder.AsObject.derive(config = myExplicitConfiguration)
```

Thanks again and please let me know what you think 🙂

travisbrown:
@Lasering This is looking great! I do think @felixbr is right about keeping configuration distinct via something like a separate `ConfiguredCodec` type, which could in principle live in a separate module, which would allow easier iteration.

Lasering:
@felixbr I also think this is the correct way to go:

```scala
case class Foo(...) derives Codec.AsObject(explicitlyPassedConfiguration)
```

But unfortunately Scala 3 does not support it. Maybe we could open a discussion on the contributors forum.

```scala
case class Foo(...) derives Codec.AsObject // doesn't use implicit Configuration

case class Bar(...) derives ConfiguredCodec.AsObject // requires implicit Configuration and doesn't compile if nothing is in scope
```

So the method `Codec.AsObject.derived` would not receive a given Configuration, and would be implemented with:

```scala
inline final def derived[A](using inline A: Mirror.Of[A]): Codec.AsObject[A] =
  given Configuration = Configuration(transformNames = identity, useDefaults = false, discriminator = None)
  ConfiguredCodec.AsObject.derived[A]
```

Or following @travisbrown idea the implementation for `Codec.AsObject.derived` and `ConfiguredCodec.AsObject.derived` would be completely distinct?

The new classes would be:

- ConfiguredCodec.AsObject
- ConfiguredDecoder
- ConfiguredEncoder.AsObject
- ConfiguredEnumCodec
- ConfiguredEnumDecoder
- ConfiguredEnumEncoder


> I also think it would be nice to be able to pass a `Configuration` object explicitly instead of having to pass all parameters one-by-one each time. It gets annoying if you have multiple case classes which share one config:
> 
> ```scala
> given Encoder.AsObject[Bar] = ConfiguredEncoder.AsObject.derive(config = myExplicitConfiguration)
> ```

This one can be easily achieved with:
```scala
object Bar:
  given Configuration = myExplicitConfiguration
  given Encoder.AsObject[Bar] = ConfiguredEncoder.AsObject.derived
  given Decoder[Bar] = ConfiguredDecoder.derived
enum Bar:
  case Batata(s: String = "bar")
  case Cenoura(i: Int)
  case Nabo
```
Or even more streamlined
```scala
object Bar:
  given Configuration = myExplicitConfiguration
enum Bar derives ConfiguredEncoder.AsObject, ConfiguredDecoder:
  case Batata(s: String = "bar")
  case Cenoura(i: Int)
  case Nabo
```

My idea behind the `derive` methods which receive the settings directly was for the classes/traits/enums which do not follow the rest of the project convention (or conventions) and need special care, and not creating the configuration explicitly would make things more streamlined. You could even mix derivation with config and without config at the same time:

```scala
object Bar:
  given Configuration = myExplicitConfiguration
  given Decoder[Bar] = ConfiguredDecoder.derive(_.toUpperCase)
enum Bar derives ConfiguredEncoder.AsObject:
  case Batata(s: String = "bar")
  case Cenoura(i: Int)
  case Nabo
```

But since its just a simple overloaded `derive` there is no harm in addind it.

Also many tests from circe-derivation could be migrated to test the new Configured* classes, however it seems the infrastructure for Scala 3 tests it not yet available in the tests module.

Lasering:
@felixbr went a slightly different path from the overloads. Now the `derived` methods have two using parameter lists (allowed in scala 3):

```scala
inline final def derived[A](using conf: Configuration = Configuration.default)(using inline A: Mirror.Of[A]): Encoder.AsObject[A] =
```

So you can do:
```scala
given Encoder.AsObject[Bar] = ConfiguredEncoder.AsObject.derived(using myExplicitConfiguration)
```

@travisbrown can I use significant indentation? Scalastyle suggests no, but it also complains about `traits or objects may not have parameters` which leads me to believe its not being used for Scala 3 sources.

felixbr:
This looks pretty good! 😄

Judging from the example the config doesn't need to be "inline", right? Even though it is called from an "inline" function. I still have to get more intuition for this new stuff.

Lasering:
@felixbr The usage of `inline` is transparent to the caller. Even if the `conf` argument was declared `inline` you would invoke the method `derived` in the same way as the example above. The fact `derived` is declared `inline` also makes no impact for use sites.

Lasering:
Implemented strict decoding for sum and product types. `DecodingFailure`s now have better messages.

The renaming object could be improved to something akin to [PureConfig NamingConventions](https://github.com/pureconfig/pureconfig/blob/master/core/src/main/scala/pureconfig/NamingConvention.scala).

fchaillou:
Hello all,
We're big users of circe-generic-extras as well and we're using the `transformConstructorNames` function to map the discriminator name.

This is a really great addition, thanks a lot for your work !

Are you planning on supporting that configuration as well ?
I could try  to take a look at implementing it by the way if not.
Thanks,
Fabien

Lasering:
@fchaillou That scenario confuses me. If you need to transform the `discriminator` why don't you set it directly with the transformed value. Can you give an example with code so I can understand it better?

fchaillou:
I'm using the `transformConstructorNames ` to change the **value** of the discriminator in my json.
Basically I could have the following hierarchy : 
```
enum Bar:
  case Kung(s: String = "bar")
  case Foo(i: Int)
```
and in the json payload i want to control the value of the discriminator :

Basically for the case Kung i would expect the following json { "type": "kung_server",  "s":  "bar"}
and for Foo I would expect  { "type": "foo_server",  "s":  "bar"}

FYI, here is what I do right now in scala 2 :

```scala
def adtJsonConfiguration(mapping: Map[Class[_], String]): Configuration = {
    def buildNormalizedClassName(clazz: Class[_]): String =
      clazz.getSimpleName.replaceAll("\\$", "")

    val normalizedMapping = mapping.map {
      case (clazz, value) => buildNormalizedClassName(clazz) -> value
    }

    Configuration.default
      .withDiscriminator("type")
      .copy(
        transformConstructorNames = name => normalizedMapping.getOrElse(name, sys.error(s"Unexpected ADT type $name"))
      )
  }
```` 

I hope it is clear

**Comment by Lasering** [comment]
I see, that makes a lot of sense. Maybe the best approach is to implement the configuration as in the generic-extras, that way the tests can be migrated directly.

**Comment by Lasering** [comment]
@fchaillou Added full support for `transformMemberNames` and `transformConstructorNames`.

**Comment by fchaillou** [comment]
Thanks a lot @Lasering, really appreciated and great work again :)

**Comment by Lasering** [comment]
Is there anything else I can do to help this move along?

File: modules/core/shared/src/main/scala-3/io/circe/derivation/Default.scala [hunk 1]
```diff
@@ -0,0 +1,46 @@
...
+import scala.quoted.*
+import scala.deriving.*
+import scala.compiletime.constValue
+
+/**
+  * Original code by Dmytro Mitin, with slight modifications by Simão Martins.
+  * See: https://stackoverflow.com/questions/68421043/type-class-derivation-accessing-default-values
+  */
+
+trait Default[T]:
```
lavrov, line 12:
```suggestion
trait Default[T] extends Serializable:
```

modules/core/shared/src/main/scala-3/io/circe/derivation/package.scala
```diff
@@ -0,0 +1,40 @@
...
+
+inline final def summonEncoders[T <: Tuple]: List[Encoder[_]] =
+  inline erasedValue[T] match
+    case _: EmptyTuple => Nil
+    case _: (t *: ts) => summonEncoder[t] :: summonEncoders[ts]
+
+inline final def summonEncoder[A]: Encoder[A] =
+  summonFrom {
+    case encodeA: Encoder[A] => encodeA
+    case _: Mirror.Of[A] => Encoder.AsObject.derived[A]
```
lavrov, line 20:
```suggestion
    case _: Mirror.Of[A] => ConfiguredEncoder.derived[A](using summonInline[Configuration])
```

modules/core/shared/src/main/scala-3/io/circe/derivation/package.scala
```diff
@@ -0,0 +1,40 @@
...
+
+inline final def summonDecoders[T <: Tuple]: List[Decoder[_]] =
+  inline erasedValue[T] match
+    case _: EmptyTuple => Nil
+    case _: (t *: ts) => summonDecoder[t] :: summonDecoders[ts]
+
+inline final def summonDecoder[A]: Decoder[A] =
+  summonFrom {
+    case decodeA: Decoder[A] => decodeA
+    case _: Mirror.Of[A] => Decoder.derived[A]
```
lavrov, line 31:
```suggestion
    case _: Mirror.Of[A] => ConfiguredDecoder.derived[A](using summonInline[Configuration])
```

lavrov (changes requested):
> Is there anything else I can do to help this move along?

I have found a few issues while trying to port `ConfiguredSemiautoDerivedSuite` from 
circe-generic-extras. I believe the chances are pretty low this gets merged without tests.

modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredEncoder.scala
```diff
@@ -0,0 +1,44 @@
+package io.circe.derivation
+
+import scala.deriving.Mirror
+import scala.compiletime.constValue
+import io.circe.{Encoder, Json, JsonObject}
+
+trait ConfiguredEncoder[A](using conf: Configuration) extends Encoder.AsObject[A], DerivedInstance[A]:
```
lavrov, line 7: Why does the trait have public methods `elemEncoders`, `encodeProduct` and `encodeSum`?

modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredDecoder.scala
```diff
@@ -0,0 +1,161 @@
+package io.circe.derivation
+
+import scala.deriving.Mirror
+import scala.compiletime.constValue
+import Predef.genericArrayOps
+import cats.data.{NonEmptyList, Validated}
+import io.circe.{Decoder, DecodingFailure, ACursor, HCursor}
+
+trait ConfiguredDecoder[A](using conf: Configuration) extends Decoder[A], DerivedInstance[A]:
```
lavrov, line 9: Why does the trait have public methods `elemDecoders` and `elemDefaults`?

modules/core/shared/src/main/scala-3/io/circe/derivation/Configuration.scala
```diff
@@ -0,0 +1,44 @@
...
+ * @param useDefaults Whether to allow default values as specified for any case-class members.
+ * @param discriminator Optional key name that, when given, will be used to store the name of the constructor of an ADT
+ *                      in a nested field with this name. If not given, the name is instead stored as a key under which
+ *                      the contents of the ADT are stored as an object.
+ * @param transformConstructorNames Transforms the value of any constructor names in the JSON allowing, for example,
+ *                                  formatting or case changes.
+ * @param strictDecoding Whether to fail when superfluous fields are found.
+ */
+case class Configuration(
+  transformMemberNames: String => String = Predef.identity,
```
lavrov, line 21: Would not it make more sense to define transformations as two functions, i.e. separate functions for encoding and decoding?

Lasering:
> have found a few issues while trying to port `ConfiguredSemiautoDerivedSuite` from circe-generic-extras. I believe the chances are pretty low this gets merged without tests.

@lavrov I would love to add the tests but the problem I referred above still remains:

> Also many tests from circe-derivation could be migrated to test the new Configured* classes, however it seems the infrastructure for Scala 3 tests it not yet available in the tests module.

I can try and change the build to support Scala 3 tests, but that seems like another PR altogether and outside of the scope of this one.

modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredEncoder.scala
Lasering: I saw no reason to hide these methods.

modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredDecoder.scala
Lasering: I saw no reason to hide these methods.

modules/core/shared/src/main/scala-3/io/circe/derivation/Configuration.scala
Lasering: Not sure If I understand what you are getting at. You want a transformMemberNames function that will be used when encoding and another one (also transforming the member names) used when decoding? If those two functions don't agree then doing `decode(encode(obj))` will never work.

modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredEncoder.scala
lavrov: Public interface is something you cannot change after release as that would break binary compatibility. Did you foresee any usage of those methods outside of this file?

modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredEncoder.scala
```diff
@@ -0,0 +1,44 @@
...
+    JsonObject.fromIterable(iterable)
+  
+  final def encodeSum(index: Int, a: A): JsonObject =
+    val (constructorName, json) = encodeElemAt(index, a, conf.transformConstructorNames)
+    conf.discriminator match
+      case None => JsonObject.singleton(constructorName, json)
+      case Some(discriminator) => json.asObject.getOrElse(JsonObject.empty).add(discriminator, Json.fromString(constructorName))
+
+object ConfiguredEncoder:
+  inline final def derived[A](using conf: Configuration = Configuration.default)(using mirror: Mirror.Of[A]): ConfiguredEncoder[A] =
```
lavrov, line 27: Using default configuration here is error prone in my opinion: if you forget to import your own configuration or remove it by accident your code will compile anyway.

modules/core/shared/src/main/scala-3/io/circe/derivation/Configuration.scala
lavrov: Hm, I think I misunderstood the way it works.

modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredEncoder.scala
lavrov: And having both `final def encodeProduct(a: A): JsonObject` and `final def encodeSum(index: Int, a: A): JsonObject` does not make sense.

modules/core/shared/src/main/scala-3/io/circe/derivation/Default.scala
Lasering: I don't understand why `Default` needs to be serializable.

modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredEncoder.scala
Lasering: Now Configured* derived methods no longer have a default configuration. And Encoder.AsObject, Decoder and Codec.AsObject now use the default configuration.

modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredEncoder.scala
Lasering: There are more alternatives to making the methods either public and final or private. We could add an annotation like mayChange, internal, experimental, etc. These annotations would imply these methods can be used, but the library is still free to break the binary compatibility, aka, use them at your own risk.

I see no problem in having both `encodeProduct` and `encodeSum` be final. Can you give more details?

I foresee a slight possibility of someone wanting to customize the derivation in a way that is not supported by the current configuration, and not wanting to re-implement most methods from `ConfiguredEncoder`/`ConfiguredDecoder`, just because they are private.

modules/core/shared/src/main/scala-3/io/circe/derivation/Default.scala
lavrov: `Encoder` and `Decoder` traits are both `Serializable`. Circe test set ensures that is true. As `Default` is declared as a field of `ConfiguredDecoder` it makes it non-serializable. https://github.com/circe/circe/pull/1800/files#diff-378e58dd77c3ec12a0e0b230ec5d55a15679e08f37f01ef83c5ee06967ca42d4R143

lavrov: There might be another way of making `ConfiguredDecoder` serializable of course.

Lasering:
Dropped `EnumConfiguration`. Migrated tests from circe-generic-extras and added a few more.

All test are passing locally. @travisbrown can you run/approve to run the CI?

modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredEncoder.scala
Lasering: Also the `derived` methods on `Configured`* classes return the ConfiguredEncoder/Decoder/Codec explicitly to more easily enable the use of the methods from those classes (like `encodeProduct` or `decodeSum`)

lavrov: > I see no problem in having both encodeProduct and encodeSum be final. Can you give more details?

Sorry for not being clear, my note was not about them being final but rather about their presence in general. Suppose I have a type `A` for which I get a derived instance of `ConfiguredEncoder`. `A` is either sum or product, not both. However it seems I am allowed to call both `encodeProduct` and `encodeSum`. I would rather keep the `ConfiguredEncoder` as minimal as possible. It could serve one distinct purpose: ability to express that one wants a configured codec e.i codec that requires `Configuration` in scope. Methods that you think could be reused may be put in another trait (which you can mix-in in the implementation).

Lasering: I see what you are getting at, but the trait approach would not work. How can you have the methods public in the trait, but make them private when you mix them in?

Lack of experience/knowledge from the programmer invoking the methods is a weak reason not to expose them. We could make the derived method transparent and return a more specialized type (think `ConfiguredSumDecoder` and `ConfiguredProductDecoder`). Those types could then have the appropriate method. Personally I don't think is worth the extra complexity (extra classes, the more confusing derived method, etc), however the implementation is easy.

lavrov: > How can you have the methods public in the trait, but make them private when you mix them in?
You can mix them in `derived` method where you actually provide implementation.

> Lack of experience/knowledge from the programmer invoking the methods is a weak reason not to expose them. 
I believe we should aim for typesafety here. Invoking these methods will throw exceptions at runtime in certain cases which is not ideal for an interface designed for everyday use.

Lasering:
> @felixbr I also think this is the correct way to go:
> 
> ```scala
> case class Foo(...) derives Codec.AsObject(explicitlyPassedConfiguration)
> ```
> 
> But unfortunately Scala 3 does not support it. Maybe we could open a discussion on the contributors forum.

@felixbr https://contributors.scala-lang.org/t/ergonomics-of-configurable-derivation/5441

Lasering:
@zarthross @zmccoy what are the next steps?

zmccoy:
Hi @Lasering I admittedly have a bit of catching up to do on a lot of these.  I just wanted you to know that @zarthross and I are going through and taking stock still as transitioning has finished.

Lasering:
Rebased to latest commits. Tests (including in Scala3) passing except this one:
> io.circe.DecoderSuite.Decoder[Float] -  should match the rounding of Float.parseFloat for known problematic inputs (#1063)

Can someone run the CI?

zarthross:
@Lasering I'm actually working on disabling that test for JS and adding JS to the CI.  (See: https://github.com/circe/circe/pull/1905).  I've approved for now, but it'll be worth merging main once more after #1905 gets in.

Lasering:
CI is failing due to scalafmtCheck on:

- modules/core/shared/src/main/scala-3/io/circe/Derivation.scala
- modules/core/shared/src/main/scala-3/io/circe/derivation/*

Probably because I'm using significant indentation. What to do?

1. Make a new scalafmt.conf file for Scala 3 sources allowing significant indentation
2. Disable scalafmt for Scala 3 sources
3. Do not use significant indentation

ybasket:
> CI is failing due to scalafmtCheck on:
> 
> * modules/core/shared/src/main/scala-3/io/circe/Derivation.scala
> * modules/core/shared/src/main/scala-3/io/circe/derivation/*
> 
> Probably because I'm using significant indentation. What to do?
> 
> 1. Make a new scalafmt.conf file for Scala 3 sources allowing significant indentation
> 2. Disable scalafmt for Scala 3 sources
> 3. Do not use significant indentation

I think you could just configure scalafmt those files differently using a `fileOverride`, see last part of https://scalameta.org/scalafmt/docs/configuration.html#scala-dialects.

Awesome that you tackle this topic btw, thank you, I'll keep fingers crossed for this PR ❤️

Lasering:
@zarthross can you run the CI again?

Lasering:
Now the CI is fully passing (the MapDecoder and BiggerDecimal losing coverage must be coming from elsewhere) the only thing missing is the docs. I'm sure this will take a while to consider acceptable breakage, after that I can update the docs to explain how configurable derivation works on Scala 3.

Swoorup:
> Now the CI is fully passing (the MapDecoder and BiggerDecimal losing coverage must be coming from elsewhere) the only thing missing is the docs. I'm sure this will take a while to consider acceptable breakage, after that I can update the docs to explain how configurable derivation works on Scala 3.
> 
> On the subject of breakage I would also like to have some discussion over [#1800 (comment)](https://github.com/circe/circe/pull/1800#issuecomment-940493993)
> 
> And as future work derivation for union types should also be implemented see [#1800 (comment)](https://github.com/circe/circe/pull/1800#issuecomment-949791796)

Can't wait for this change to land. 😄

meldmy:
@zarthross  Could you please merge this PR? 🙏🏻

Lasering:
@zarthross 

As it stands the the only breakage would be if someone was using the `DerivedEncoder` or `DerivedDecoder` classes directly, which have been replaced with `ConfiguredDecoder`/`ConfiguredEncoder`/`ConfiguredCodec`. Although a more consistent check should be performed with Mima.

As @lavrov pointed out the new code might introduce binary problems in the future since the derived methods now return the `Configured*` classes directly and not the `Encoder.AsObject[A]`/`Decoder[A]`/`Codec.AsObject[A]` as before (see the comments on the [review](https://github.com/circe/circe/pull/1800/files#diff-55e23c2205296c1faf76d9cbfea4eb4d3932594f78468fad47e94aea20bfc311)).

The biggest breakage would be to proceed with:

> I don't have a strong reasoning for this but it seems more ergonomic to put the `derived` methods in:
> 
> * the object `Encoder` instead of object `Encoder.AsObject`
> * the object `Codec` instead of object `Codec.AsObject`
> 
> I wager most people will try `derives Encoder` or `derives Codec` and when that fails will be forced to google the answer.

In this case I think the gained improved user experience still greatly outweighs the problems that will arise from the breakage. This is based on:

- Most users don't know derivation was already supported in 0.14
- The derivation was "hidden" behind the `AsObject` objects
- The fix to the breakage is simply changing:

    - `derives Encoder.AsObject` to `derives Encoder`
    - `derives Codec.AsObject` to `derives Codec`

Lasering:
The compiler flag `-Yretain-trees` is no longer needed. Thanks [ftucky](https://stackoverflow.com/users/17922001/ftucky)!

Lasering:
@zmccoy @zarthross @travisbrown I would really like to see this merged. What are your reservations or what is making the process so slow? Is there anything I can do to help? Maybe work on some other PRs. A more detailed explanation of the changes. Trim down the PR so its easier to review then open another with the remaining changes. Is there anything that would make you more comfortable to merge the PR?

TL;DR; what is keeping this PR from being merged? Maybe we can talk/discuss some points to make some progress.

travisbrown:
@Lasering I'm no longer working on Scala open source projects for the reasons described [here](https://meta.plasm.us/posts/2021/11/17/scala-open-source/), and I'm turning off further notifications for this PR. Thanks for understanding.

Swoorup:
Merge it 😛

Lasering:
Looking into improving the error messages using the new classes from https://github.com/circe/circe/pull/1617. Namely `WrongTypeExpectation` and `MissingField`. Would this be welcomed?

zarthross:
@Lasering Do you mind merging latest main into this branch so we can re-run ci?

Lasering:
@zarthross done.

Lasering:
Running `coreJVM / mimaReportBinaryIssues` I get:

```
[error] Circe core: Failed binary compatibility check against io.circe:circe-core_3:0.14.0! Found 5 potential problems (filtered 37)
[error]  * class io.circe.Derivation does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("io.circe.Derivation")
[error]  * object io.circe.Derivation does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("io.circe.Derivation$")
[error]  * interface io.circe.DerivedDecoder does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("io.circe.DerivedDecoder")
[error]  * interface io.circe.DerivedEncoder does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("io.circe.DerivedEncoder")
[error]  * interface io.circe.DerivedInstance does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("io.circe.DerivedInstance")
```
How should I proceed?

Also have a look at https://github.com/circe/circe/pull/1800/files#r739704540.

Lasering:
The build.sbt currently in main broke the PR. Sources under `core/shared/src/main/scala-3` give weird errors:

```
[error] -- Error: /home/simon/Documents/circe/modules/core/shared/src/main/scala-3/io/circe/Derivation.scala:10:2 
[error] 10 |  inline final def derived[A](using inline A: Mirror.Of[A]): Encoder.AsObject[A] =
[error]    |  ^^^^^^
[error]    |  indented definitions expected, identifier inline found
[error] -- Error: /home/simon/Documents/circe/modules/core/shared/src/main/scala-3/io/circe/Derivation.scala:10:2 
[error] 10 |  inline final def derived[A](using inline A: Mirror.Of[A]): Encoder.AsObject[A] =
[error]    |  ^^^^^^
[error]    |  indented definitions expected, identifier inline found
[error] -- Error: /home/simon/Documents/circe/modules/core/shared/src/main/scala-3/io/circe/derivation/Configuration.scala:4:2 
[error] 4 |  val default: Configuration = Configuration()
[error]   |  ^^^
[error]   |  indented definitions expected, val found
```
This sources are using significant based indentation (validated by scalafmt), and are correctly compiling with the build.sbt in Lasering/circe. Unfortunately I cannot find what is causing the problem. Adding scalacOption `-indent` does not work since the code not under `scala-3` errors out.

armanbilge:
@Lasering I pulled up your branch to take a look but it seems you haven't merged in the latest main yet?

Lasering:
Thanks that fixed it. Although I needed to add it to both the core and tests projects.

felixbr:
I looked at the CI failure and it found a binary incompatibility:

```
[error] Circe core: Failed binary compatibility check against io.circe:circe-core_3:0.14.2 (e:info.versionScheme=early-semver)! Found 5 potential problems
[error]  * class io.circe.Derivation does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("io.circe.Derivation")
[error]  * object io.circe.Derivation does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("io.circe.Derivation$")
[error]  * interface io.circe.DerivedDecoder does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("io.circe.DerivedDecoder")
[error]  * interface io.circe.DerivedEncoder does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("io.circe.DerivedEncoder")
[error]  * interface io.circe.DerivedInstance does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("io.circe.DerivedInstance")
```

If I understand this correctly, it's trying to find classes like `io.circe.Derivation` in the Scala 2 version of `circe-core`. This doesn't make much sense because derivation only exists in Scala 3.

Either we silence the errors as false-positives or add stub-implementations to the Scala 2 code path. Not sure if stubs are necessary tbh.

Apart from that error, is there anything else blocking this PR? It's been over a year and it's still blocking some Scala 3 migrations for me at work, so if there is anything I can help with, please let me know.

armanbilge:
This PR has been retargeted to 0.15.x, but CI has not been re-run so it still shows the MiMa issues when this branch was targeted at 0.14.x.

Lasering:
While using this PR in a personal project I came across this issue on Scala 3.1.2 https://github.com/lampepfl/dotty/issues/15042

Lasering:
[Mirrors are not generated for union types](https://github.com/lampepfl/dotty/pull/15279) hence derivation does not work for union types. This should be explicit in the documentation.

Lasering:
Is there anything I can change or improve to make this move along. It has been more than 1 year since this PR has been working and tested. Breaking changes can be mitigated or entirely removed, but without any feedback in either direction this is simply gaining dust and, I assume, delaying the Scala 3 migration to a lot of people.

armanbilge:
> Breaking changes can be mitigated or entirely removed

Now that this PR is targeted to 0.15.x there should no longer be any breaking changes. See https://github.com/circe/circe/pull/1800#issuecomment-1212074971. If you push another commit or close/reopen the PR that will trigger CI on a fresh merge, which will no longer check compatibility against 0.14.x.

I do not know when 0.15.0 will be released. If there is a way to make some of these changes in a non-breaking way, you may want to target that at series/0.14.x

zarthross:
@Lasering Sorry for the delay, I'll try and give this a good look over this weekend and give any feedback.  I would love to get this into 0.14 if possible, but it looks like there are bin compat issues which is why i switched the target to 0.15.

Lasering:
@zarthross 0.15 is perfect. Let me try to rebase to 0.15 on my side, to see if there isn't any unexpected issues.

Lasering:
Branch was updated to be on top of series/0.15.x, code still compiling and passing the tests.

Any estimates on the 0.15.x release?

Lasering:
@zarthross can you run the CI?

Lasering:
@zmccoy maybe you could run the CI?

zmccoy:
@Lasering Done!

zmccoy:
@Lasering  @zarthross and I will chat in the coming week for releases, he's out at the moment and I'm about to be missing for the holiday.

zarthross:
@Lasering Sorry for the delay.  I took a closer look at those bin-compat issues, and I'm sure they can be worked around.  I would LOVE for this to be based on 0.14 instead of 0.15 so we can get a release of circe-generic-extras release with scala-3 support, and this is just the code changes we need for that.

Here is a commit i was working on a rebased branch of this PR: https://github.com/circe/circe/commit/7f6d95846bfeb6ab1e4680def04afec513907950  This should give some idea of the changes I was looking at possibly making in addition to the bin-compat fixes.  I would be happy to force-push that to this PR if you like.  Or open another one.  Or you can just cherry pick my WIP commit.  Let me know and we can continue pushing this forward.

Lasering:
I've cherry-picked your commit.

I can rebase to be on top of 0.14.x, that will cause some MIMA issues I'll need some help to resolve.

plokhotnyuk:
@Lasering I appreciate how much work you done to port these missing features to Scala 3!

But, I'm curious, have you saw that [jsoniter-scala](https://github.com/plokhotnyuk/jsoniter-scala) provides auto-derivation with all these code generation options emitting codecs that are [much more efficient in runtime](https://plokhotnyuk.github.io/jsoniter-scala/)?

Lasering:
@plokhotnyuk jsoniter and the code in this PR have very different approaches, jsoniter relies on macros for the (auto) derivation, where as this PR uses Scala 3 derivation that is built into the compiler. There are users and use-cases for both scenarios.

zarthross:
@Lasering Among my fixes are those fixes for the mima issues.

Lasering:
@zarthross So I just need to target series/0.14.x or is it main?

zarthross:
series/0.14.x

Lasering:
Done. Also the initial review by lavrov is still relevant: https://github.com/circe/circe/pull/1800#pullrequestreview-793614446

I can also squash the commits if needed.

Lasering:
@zarthross the test that didn't pass in the CI has nothing to do with this PR:
```
io.circe.pointer.literal.PointerInterpolatorSuite.The pointer string interpolater should work with arbitrary interpolated strings
```

Lasering:
That test is passing locally, so I guess its a problem in the CI. Maybe running it again will solve it.

armanbilge:
> That test is passing locally, so I guess its a problem in the CI.

The data for that test is randomly generated. So it's very possible it can pass locally even if the failure is legitimate, because it is using different data.

To replicate it you should set the seed for the test suite as indicated in the error.

<details>

```
==> X io.circe.pointer.literal.PointerInterpolatorSuite.The pointer string interpolater should work with arbitrary interpolated strings 0.02s munit.FailException: /home/runner/work/circe/circe/modules/pointer-literal/src/test/scala/io/circe/pointer/literal/PointerInterpolatorSuite.scala:47
46:
47:  property("The pointer string interpolater should work with arbitrary interpolated strings") {
48:    Prop.forAll(ScalaCheckInstances.genPointerReferenceString) { (v: String) =>

Failing seed: dCZ7SqVQkx0Poo9i91WvyTPwWHzCR5fjs6xz9NRE9VM=

  override val scalaCheckInitialSeed = "dCZ7SqVQkx0Poo9i91WvyTPwWHzCR5fjs6xz9NRE9VM="

Falsified after 6 passed tests.
> Labels of failing property: 
Expected Right(/foo/~0/bar) but got Left(io.circe.pointer.PointerSyntaxError: Syntax error at 5; expected: token character)
> ARG_0: "~"
> ARG_0_ORIGINAL: "ቋ~0ꙮ帱ᘈ讅"
```

</details>

However as you point out, it is unlikely to be related to your PR, but an existing error that has surfaced due to the random test generation.

Lasering:
Using that seed the test fails locally. So there is really a problem in pointerLiteral in Native. But it is unrelated to this PR. Re-running that job will most likely make it pass, thus allowing this PR to progress.

armanbilge:
> So there is really a problem in pointerLiteral in Native.

Do you not get this failure on the JVM or JS?

Lasering:
Fails on all platforms. I had only tested it on Native because the CI was only failing on Native, but it makes sense to also fail in the other platforms since the seed is not the same for the other platforms in CI.

kailuowang:
> Using that seed the test fails locally. So there is really a problem in pointerLiteral in Native. But it is unrelated to this PR. Re-running that job will most likely make it pass, thus allowing this PR to progress.

+1 on re-run CI and merge and release. This will help many projects upgrade to Scala 3.

Lasering:
@zarthross can you re-run the CI?

zmccoy:
@Lasering That should be re-run now.

Lasering:
Its ready to be merged :)

zmccoy:
@zarthross and I are getting together this week to chat about this merge and releases 👍

Lasering:
I can improve slightly the error messages using `Reason` from the https://github.com/circe/circe/pull/1617 plus stating that transformConstructorNames is being used. For example on ConfiguredEnumDecoder:
```scala
val transform = if conf.transformConstructorNames == Predef.identity then "" else " transformConstructorNames is being applied"
val error = s"enum '${constValue[mirror.MirroredLabel]}' does not contain case: '$caseName'. Expected one of: ${labels.mkString(",")}$transform."
Left(DecodingFailure(error, c.history))
```
Or in ConfiguredDecoder instead of `DecodingFailure(s"$name: expected a json object.", c.history)` it could be `DecodingFailure(Reason.WrongTypeExpectation("object", c.value), c.history)`

zarthross:
> Or in ConfiguredDecoder instead of `DecodingFailure(s"$name: expected a json object.", c.history)` it could be `DecodingFailure(Reason.WrongTypeExpectation("object", c.value), c.history)`

I would say we should use the `DecodingFailure.Reason` type as much as possible.

Lasering:
@zarthross Added Reason.WrongTypeExpectation where it makes sense.

zmccoy:
@Lasering Once the .gitignore merge conflict is solved I'll make sure I watch to make sure I rerun if any weird failures happen on pulls etc :)

Lasering:
Resolved .gitignore merge conflict, and rebased to the latest changes of series/0.14.x.

modules/core/shared/src/main/scala-3/io/circe/Derivation.scala
```diff
@@ -1,300 +1,52 @@
...
-    case encodeA: Encoder[A] => encodeA
-    case _: Mirror.Of[A]     => Encoder.AsObject.derived[A]
-  }
-
-  inline final def summonDecoder[A]: Decoder[A] = summonFrom {
-    case decodeA: Decoder[A] => decodeA
-    case _: Mirror.Of[A]     => Decoder.derived[A]
-  }
+  inline final def summonDecoders[T <: Tuple]: Array[Decoder[_]] = summonDecoders[T].toArray
+  inline final def summonEncoders[T <: Tuple]: Array[Encoder[_]] = summonEncoders[T].toArray
```
zarthross: These are accidentally infinitely recursive at runtime 
```suggestion
  inline final def summonDecoders[T <: Tuple]: Array[Decoder[_]] = derivation.summonDecoders[T](using Configuration.default).toArray
  inline final def summonEncoders[T <: Tuple]: Array[Encoder[_]] = derivation.summonEncoders[T](using Configuration.default).toArray
```

modules/core/shared/src/main/scala-3/io/circe/Derivation.scala
```diff
@@ -1,300 +1,52 @@
...
-          encodeWith(m.ordinal(a))(a) match {
-            case (k, v) => JsonObject.singleton(k, v)
-          }
-      }
-    }
-}
+@deprecated(since = "0.14.4")
+private[circe] trait DerivedInstance
+private[circe] trait DerivedEncoder
+private[circe] trait DerivedDecoder
```
zarthross: Do these also need to be marked `deprecated` I don't see them used anywhere?

modules/core/shared/src/main/scala-3/io/circe/derivation/Configuration.scala
```diff
@@ -0,0 +1,43 @@
...
+  transformConstructorNames: String => String = Predef.identity,
+  useDefaults: Boolean = false,
+  discriminator: Option[String] = None,
+  strictDecoding: Boolean = false
+):
+  def withTransformMemberNames(f: String => String): Configuration = copy(transformMemberNames = f)
+  def withSnakeCaseMemberNames: Configuration = withTransformMemberNames(renaming.snakeCase)
+  def withScreamingSnakeCaseMemberNames: Configuration = withTransformMemberNames(renaming.screamingSnakeCase)
+  def withKebabCaseMemberNames: Configuration = withTransformMemberNames(renaming.kebabCase)
+
```
zarthross, line 31: We'll need to add `withPascal...` for circe-generic-extras parity, but that shouldn't hold this up.

modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredEnumDecoder.scala
```diff
@@ -0,0 +1,27 @@
+package io.circe.derivation
+
+import scala.deriving.Mirror
+import scala.compiletime.constValue
+import Predef.genericArrayOps
+import io.circe.{ Decoder, DecodingFailure, HCursor }
+
+trait ConfiguredEnumDecoder[A] extends Decoder[A]
```
zarthross, line 8: Should this extend `DerivedInstance` like the other `Configured*` types?

modules/core/shared/src/main/scala-3/io/circe/Derivation.scala
Lasering: Those were being used in:
https://github.com/circe/circe/blob/series/0.14.x/modules/core/shared/src/main/scala-3/io/circe/Derivation.scala#L44
https://github.com/circe/circe/blob/series/0.14.x/modules/core/shared/src/main/scala-3/io/circe/Derivation.scala#L65
https://github.com/circe/circe/blob/series/0.14.x/modules/core/shared/src/main/scala-3/io/circe/Derivation.scala#L137
But they were already `private[circe]` so there is no need to mark them deprecated.

modules/core/shared/src/main/scala-3/io/circe/derivation/Configuration.scala
Lasering: Added support for PascalCase.

modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredEnumDecoder.scala
Lasering: Looking at the code with fresh eyes I would say `DerivedInstance` could be deleted. It is mainly used in `ConfiguredDecoder` and `ConfiguredEncoder`, but  ConfiguredEncoder only uses the `elemLabels`. In both cases the core logic is in the `Configured*` creation, aka:

```scala
new ConfiguredEncoder[A]
      with DerivedInstance(
        constValue[mirror.MirroredLabel],
        summonLabels[mirror.MirroredElemLabels]
      ):

new ConfiguredDecoder[A]
      with DerivedInstance(
        constValue[mirror.MirroredLabel],
        summonLabels[mirror.MirroredElemLabels]
      ):
```

So `DerivedInstance` isn't adding any value.

Alternatively it could be used as a marker trait for encoders/decoders/codecs that were automatically derived. In that case it could be added to all the `Configured*` types but without any members.

modules/core/shared/src/main/scala-3/io/circe/Derivation.scala
zarthross: its my understanding that another artifact within the circe org could still reference those traits since they will be within the `io.circe` package.  I would like them marked deprecated to prevent them from accidentally being referenced elsewhere.

modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredEnumDecoder.scala
zarthross: I see you deleted it, that works for me.

modules/core/shared/src/main/scala-3/io/circe/Derivation.scala
zarthross: Go ahead and mark these deprecated for me and assuming the build passes I think we can get this merged in and a release cut this week. 

Thank you for your work on this, its greatly appreciated!

zarthross (approved):

zmccoy (approved):

javierarrieta:
That is great, thanks a million for the hard work on this. Sadly I don't work in Scala anymore, but this was one of the main blockers of our migration to Scala3 15 months ago when I left.

I personally am really grateful and hopefully I will be able to use it in the future, but I am pretty sure this will help quite a few people moving to Scala 3.

fchaillou:
> That is great, thanks a million for the hard work on this. Sadly I don't work in Scala anymore, but this was one of the main blockers of our migration to Scala3 15 months ago when I left.
> 
> I personally am really grateful and hopefully I will be able to use it in the future, but I am pretty sure this will help quite a few people moving to Scala 3.

My thoughts exactly, except I'm still working on this :)
Any plans on cutting a release in the next few days ?
Thanks again for the great work !

zmccoy:
@fchaillou We're actually getting together to prep this release tonight.  Hold tight :)

zarthross:
@Lasering Huge thanks for working on this with us!  It's now released as part of https://github.com/circe/circe/releases/tag/v0.14.4

PR #1818, opened 2021-08-31
Drop Scala 2.12

travisbrown:
I don't know the exact timing on this, but it's likely to happen before 0.15.0, which is likely to happen this year.

This is just a start—there are a lot of deprecations (especially in the tests) that we can finally clean up now.


djspiewak:
Okay I'll wade into the hornet's nest… (with the caveat that absolutely nothing I'm saying here should be construed as me attempting to control what you do with your free time)

The 2.12 -> 2.13 jump is a much more significant one than most Scala upgrades. A lot of production codebases are still stuck on 2.12 for a myriad of reasons and are likely to remain so for a long time. Circe abandoning support for that version is not going to push them faster, it's simply going to leave them without meaningful maintenance for their JavaScript ingestion layer.

One obvious example: neither Databricks nor Flink currently support Scala 2.13. Anyone who touches those ecosystems at all (which is conservatively around 2/3 of Scala's current adoption footprint!) is going to be stuck on 2.12 for a while longer. Based on past history, "a while" is likely to be a meaningful period of time, since these two only just got off of 2.11!

Additionally, sbt is stuck on Scala 2.12 for now. @eed3si9n has mentioned that 2.13 support might not be that hard, but there's a whole ecosystem of plugins and such which would need to follow. I wouldn't underestimate the entropy in that space.

In general, the Scala community has converged around cross-building for the two most recent stable Scala versions, plus a long tail of "best effort, whatevs" depending on the project and the ecosystem. With the arrival of Scala 3 and the significant adoption hurdle it represents, most projects have expanded the "two most recent" to "two latest Scala 2 versions and singular latest Scala 3", at least for the time being. We're in a delicate place right now as the ecosystem slowly lurches forward, and it's important that we (as foundational library maintainers) try not to introduce undue burden.

Speaking of undue burden, a lot of this calculus comes down to a question of how hard it is to maintain Circe's 2.12 support. This is something that only you can answer, really. Every project is different. At least with Cats Effect, supporting 2.12 in the first place was quite challenging (most of CE3 was first written against Scala 2.13 and then ported forward and backward to Scala 3 and 2.12, respectively), but those challenges are effectively in the past now. Scala 2.12 support even yields some surprising edge-case benefits, since the set of compiler warnings is a bit different and sometimes catches things that 2.13 and 3.0 miss. I can't remember the last time I had any particular heartburn over maintaining support, beyond the longer build times when creating a release. But every project is different, and if Circe's support for 2.12 is an acute and ongoing pain, I can understand the desire to put it to bed.

Which brings me to a final point, which is that making this choice is effectively forcing the hand of a large swathe of the ecosystem. Anything downstream of Circe must *also* drop Scala 2.12 support (or pin themselves to an old version), and that magnifies the impact of this decision quite dramatically. In other words, you're not just looking at the Circe footprint, but also the transitive footprint of anything which has submodules which bring in Circe, which is *vast* (partially because of http4s). I sincerely would not underestimate the import of this change. Circe dropping Scala 2.12 very likely means that the majority of the Scala ecosystem will follow suit in short order, but then you end up in a scenario where the venn diagrams of various subsets of Scala's adoption space no longer have any overlap, which is undesirable to say the least.

Again, not telling you how you should spend your time, just attempting to document the implications of what this would mean and suggesting that it is worth a meaningful discussion weighing the pros and cons.

travisbrown:
@djspiewak Thanks for the feedback! The key part of the decision for me comes down to the fact that I just don't see adopters "pinning themselves to an old version" as a bad option. Circe is pretty stable: while we don't make guarantees about bincompat between 0.x series, we haven't broken bincompat in any real way (specifically excluding the Jawn modules because of Jawn 1.0) since 0.12.0 in September 2019. Most adopters are still on 0.13.x, which is a year and a half old.

We'll continue to publish fixes for bugs in 0.14.x (although a lot of people are using it and nothing's turned up since May). 0.14.x supports both Scala 2.12 and Scala 3, so on the off-chance that someone wanted to migrate directly from 2.12 to 3, they can do that with 0.14. The 0.15 release will have some more substantial improvements than 0.13 or 0.14 did (e.g. better error types), but it still won't really be a big release.

Keeping 2.12 around isn't excessively burdensome, but it does involve some inelegant stuff that it'll be nice to get rid of. Given that 0.14 isn't going anywhere and won't become a bad choice just because 0.15 is out, the cost of keeping Scala 2.12 around could be even smaller than it actually is and I'd still be inclined to drop it.

I'll acknowledge that the place where this could cause problems is if e.g. Circe 0.15 comes out in October and the http4s maintainers want to publish a new release that updates to Circe 0.15 but keeps Scala 2.12 support. In that case they have three options:

* stick to Circe 0.14 like everyone else who needs Scala 2.12;
* publish the Circe modules only for Scala 2.13 and 3 and include 2.12 for everything else; or
* drop 2.12 themselves.

My guess is that many of the maintainers of libraries like http4s will be happy to have an excuse to drop 2.12 by the time Circe 0.15 is out, and I'm happy to provide that. They can just tell their 2.12 users to use their previous stable version, the same as I am here.

djspiewak:
> So the goal is also to make life a little more inconvenient for e.g. Databricks customers, and to make it as clear as possible who's responsible for that inconvenience.

For what it's worth, this will backfire.

I can't speak too much about this in public contexts, but this argument basically boils down to "I'm going to make my adopters suffer so that they can finally apply pressure to Databricks and make them realize this is important." The flawed assumption here is that your adopters are not *already* applying pressure to Databricks. At the end of the day, Spark is many orders of magnitude larger and more entropic than Circe, even including the transitive ecosystem effect. Making your users feel pain is more likely to make them *stop* working for the success of the ecosystem, rather than vice versa.

Large vendors move when large companies exert pressure, and large companies exert pressure when internal elements are motivated to push the boulder up hill. If the boulder suddenly grows spikes, it *always* becomes much easier for those internal elements to just let it go and find a different thing to care about. Large companies do *not* exert pressure when dependencies tell them something they already know (in this case, that it's important to upgrade to Scala 2.13).

travisbrown:
@djspiewak Again I think "if the boulder suddenly grows spikes" is overestimating the inconvenient for users here. The difference between 0.14 and 0.15 will basically be a very slightly nicer error API in the latter. A Scala-2.12-free 0.15 release is not suddenly going to cause 0.14 to crash their services or whatever.

I don't have any delusions about how effective my little protest will be, and I also don't think anything about it is dramatic enough to do something like "backfire".

djspiewak:
> I don't have any delusions about how effective my little protest will be, and I also don't think anything about it is dramatic enough to do something like "backfire".

I hope you're right. The version treadmill is a brittle cycle for library adoption. Once fallen off, it often becomes just another internalized element of tech debt, particularly when it's something that happens at ecosystem scale (see also: how long did it take all of us to upgrade to ScalaCheck 0.14?).

johnynek:
Just a small note with a few points:

1. I don't like to see personifications of big companies. Big companies don't use circe. Engineers at big companies who have been given the ability to choose open source libraries but *not the ability to pay external contractors* use circe. You don't hurt, e.g. Stripe with this call, you hurt the few engineers at Stripe trying to argue that Scala is still a good choice for development. If they bring this concern to their manager very likely the manager will say: "can we use another library, or can we just use the old version, or maybe it is time we switch to Java", because, surprise, their manager also usually doesn't have the ability to just pay an external contractor. I have actually gone through this process on a team twice, and in both cases it was a huge time investment and took months.
2. That said, I think adding a switch to your build so that a cross build that depends on circe switching to an old version for 2.12 isn't so burdensome. I've used this approach for cats-parse support for 2.11 and it has been fine. The main cost of this is that this is now a specialized build tax on all the downstream libraries, and possibly interacts badly with scala-steward.
3. This post was framed on Twitter like you are burned out with maintaining OSS software, possibly especially with scala. I think folks could work with you to find a new maintainer.

rossabaker:
> My guess is that many of the maintainers of libraries like http4s will be happy to have an excuse to drop 2.12 by the time Circe 0.15 is out, and I'm happy to provide that. They can just tell their 2.12 users to use their previous stable version, the same as I am here.

These problems skew toward enterprise engineer and away from disillusioned maintainer.  We share the latter title, so I sympathize, but I intend to keep 2.12 support in http4s and hope the ecosystem does too.

Cats-Effect 3 is a higher priority than 2.13.  It is far more effort to maintain branches across breaking stacks than crossbuilding for multiple Scalas.  If we can get to Cats-Effect 3, we can stop maintaining and stop training on multiple versions.  I can sell this.  I can't sell 2.13.

Could we just use the old version?  Yes, but the more points we lag, the harder the upgrade.  One key project at work is on Scala 2.11 and http4s-0.18 and circe-0.9, where it will die.  It is my job to not allow this to happen to a hundred more.

Are we unusual?  2.12 is still 45% of http4s downloads, finally overtaken by 2.13 in May 2021.  It's still the median version: we have many more users on 2.11 than 3.  2.12 is near steady relatively and growing absolutely.

Do I care?  Some small percentage of our users give a lot back, and about half of them are stuck too.  Cutting them off when they exert so little drag would be bad for http4s and bad for business.

Should you care?  :shrug:  But Databricks won't.  If you want the strongest Circe while pouring less of yourself into it, I echo Oscar's final point.

PR #1834, opened 2021-10-01
Documentation fixes

mattkohl:
The main change here is to standardize how module names are styled in the documentation. Instead of being sometimes marked out & others not, now every time a module is referred to, it's enclosed in back-ticks, as in `generic-extras`.

The PR includes a couple of grammatical & case adjustments as well.

Thanks for all the work you 🙏


zmccoy (approved):

docs/src/main/tut/codecs/custom-codecs.md
```diff
@@ -87,7 +87,7 @@ json.as[Map[Foo, Int]]
 
 It's often necessary to work with keys in your JSON objects that aren't idiomatic case class member
 names in Scala. While the standard generic derivation doesn't support this use case, the
-experimental circe-generic-extras module does provide two ways to transform your case class member
+`generic-extras` module does provide two ways to transform your case class member
```
zarthross, line 90: We need to keep the 'experimental' verbiage here for now.  The future of `generic-extras` is still unknown and that repo is labeled as 'experimental'.

docs/src/main/tut/codecs/custom-codecs.md
```diff
@@ -132,7 +132,7 @@ implicit val config: Configuration = Configuration.default
 Bar(13, "Qux").asJson
 ```
 
-It's worth noting that if you don't want to use the experimental generic-extras module, the
+It's worth noting that if you don't want to use the `generic-extras` module, the
```
zarthross, line 135: Same as the other note.  Want people to consider it experimental, know that it may not be supported in exactly the same way in the future.

zarthross (changes requested):

zmccoy (approved):
Thanks for fixing those up!

zarthross (approved):

PR #1855, opened 2021-11-18
Add `printToAppendable`

armanbilge:
Not sure if the precedent here is to avoid side-effecting APIs at all costs (if so, apologies), but personally I would find this useful. I started this PR to add an API for writing to an `OutputStream` but realized it could be more general.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Printer.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Printer.scala
@@ -74,8 +74,8 @@ final case class Printer(
     final def onNumber(value: JsonNumber): Unit = value.appendToStringBuilder(writer)
   }
 
-  private[this] final class AppendableByteBufferFolder(
-    writer: Printer.AppendableByteBuffer
+  private[this] final class AppendableFolder(
+    writer: Appendable
   ) extends Printer.PrintingFolder(writer, pieces, dropNullValues, escapeNonAscii, sortKeys) {
     final def onBoolean(value: Boolean): Unit = writer.append(java.lang.Boolean.toString(value))
     final def onNumber(value: JsonNumber): Unit = writer.append(value.toString)
@@ -189,9 +189,7 @@ final case class Printer(
       w
     } else new StringBuilder()
 
-    val folder = new StringBuilderFolder(writer)
-
-    json.foldWith(folder)
+    printToAppendable(json, writer)
 
     writer.toString
   }
@@ -209,15 +207,18 @@ final case class Printer(
       else Printer.NoSizePredictor
 
     val writer = new Printer.AppendableByteBuffer(cs, predictor)
-    val folder = new AppendableByteBufferFolder(writer)
-
-    json.foldWith(folder)
+    printToAppendable(json, writer)
 
     writer.toByteBuffer
   }
 
   final def printToByteBuffer(json: Json): ByteBuffer =
     printToByteBuffer(json, StandardCharsets.UTF_8)
+
+  final def printToAppendable(json: Json, out: Appendable): Unit = {
+    val folder = new AppendableFolder(out)
+    json.foldWith(folder)
+  }
 
   /**
    * The same pretty-printer configuration that outputs fields in sorted order.
--- a/modules/core/shared/src/main/scala/io/circe/Printer.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Printer.scala
@@ -189,7 +189,7 @@ final case class Printer(
       w
     } else new StringBuilder()
 
-    printToAppendable(json, writer)
+    unsafePrintToAppendable(json, writer)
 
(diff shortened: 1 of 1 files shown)


modules/core/shared/src/main/scala/io/circe/Printer.scala
```diff
@@ -209,16 +207,19 @@ final case class Printer(
...
   }
 
   final def printToByteBuffer(json: Json): ByteBuffer =
     printToByteBuffer(json, StandardCharsets.UTF_8)
 
+  final def printToAppendable(json: Json, out: Appendable): Unit = {
+    val folder = new AppendableFolder(out)
+    json.foldWith(folder)
+  }
+
```
zarthross, line 222: Thanks for the PR!  I think circe has really tried to avoid public methods with side effecting signatures, so I'm fairly hesitant to introduce one. That being said,  I think if we mark the boundaries of safeness and let others know that thar be dragons.. then we can probably let it in.  So something like this:
```suggestion
  object Unsafe {
    final def printToAppendable(json: Json, out: Appendable): Unit = {
      val folder = new AppendableFolder(out)
      json.foldWith(folder)
    }
  }
```

And we should add some scaladoc to the method as well.  For instance, it should be noted that this method returns only after all JSON is written to the appendable and that any use of the `Appendable` outside this method voids the warranty 😆 .

zarthross (changes requested):

armanbilge:
>  I think circe has really tried to avoid public methods with side effecting signatures, so I'm fairly hesitant to introduce one.

Yes, that's completely fair! Thanks for being open to it. The goal here is mostly to expose a higher-performance API, so here's an idea to do it safely:
1. Keep the changes I have here, but make it `private[circe] def printToAppendable`
2. Then in `circe-fs2`, expose a public API that uses `printToAppendable` internally. This way we could use effects to provide a safe/pure API.

The catch with this is that even though `printToAppendable` is technically private to circe it would need the usual bincompat guarantees because `circe-fs2` is released/versioned independently. This is easily mitigated with a comment drawing attention to this. What do you think?

Otherwise...

👍 to `unsafe` and docs. Just a suggestion: instead of an `Unsafe` object, what about naming the method `unsafePrintToAppendable`? This is how `unsafe` methods are named in Cats Effect, and has the advantage of putting `unsafe` directly at the call site (rather than a tricky `import circe.Printer.Unsafe._` and then just an innocent-looking `printToAppendable`. No strong feelings either way, just a suggestion.

Thanks again for the review, and congrats on maintainership :)

zarthross:
I completely understand the goal of the higher-performance API. 

 Lets for now go with `unsafePrintToAppendable` and start work toward something in `circe-fs2` to replace it.

I'm thinking that we might need a different implementation completely instead of what we have here... maybe something based on `Json.Folder[fs2.Stream]` instead of the current side effectful `Json.Folder[Unit]`.  

Happy to help and thanks!

armanbilge:
Sounds good, updated!

> I'm thinking that we might need a different implementation completely instead of what we have here...

I'm not sure. I was imagining a circe version of this fs2 method:
```scala
def writeOutputStream[F[_]](
    fos: F[OutputStream],
    closeAfterUse: Boolean = true
)(implicit F: Sync[F]): Pipe[F, Byte, INothing]
```
https://github.com/typelevel/fs2/blob/f0272ce00d930cf4eceaae67023517365c5fca6f/io/shared/src/main/scala/fs2/io/io.scala#L99-L108

So something like this (bikeshed the method names):
```scala
// this one prints a stream of JSON as an array:
// [{},{},{}]
def printArrayToAppendable[F[_]](
    fapp: F[Appendable],
)(implicit F: Sync[F]): Pipe[F, Json, INothing]

// this one uses a custom delimiter instead, e.g.:
// {}\n{}\n{}
def printToAppendable[F[_]](
    fapp: F[Appendable],
    delim: CharSequence,
)(implicit F: Sync[F]): Pipe[F, Json, INothing]
```

zarthross:
To bikeshed a bit, i was just imaging something of the signature `def printToStream[F[_]](json: Json, cs: Charset): Stream[F, Byte]` but we could go and do something like  `def printPipe[F[_]](cs: Charset): Pipe[F, Json, Byte]` as well.  

For your use case in Feral you would be able to then do `printToStream(json).through(fs2.io.writeOutputStream(channel))` or something.

Reason I suggest a different printer implementation is I think having an FS2 native implementation would be able to suspend the encoding of the json better, and just overall work better with FS2 than trying to hack around `Appendable`

zarthross (approved):

zarthross:
@armanbilge Thanks again for the PR.  I'll have @zmccoy give it a review, and we'll get this into the next release.

zmccoy (approved):
Thanks for working through this with us!  This looks good to me.

PR #1874, opened 2021-12-05
add scala 3 macro support for literal

i10416:
Hello, I added scala 3 macro support for literal module.

- `+literalJVM/test` passed for scala 2.12, 2.13 and 3.1


build.sbt
```diff
@@ -451,11 +451,30 @@ lazy val literalBase = circeCrossModule("literal", mima = previousCirceVersion,
...
+    Test / unmanagedSourceDirectories ++= {
+      def extraDirs(suffix: String) =
+        CrossType.Pure.sharedSrcDir(baseDirectory.value, "test").toList.map(f => file(f.getPath + suffix))
+
+      CrossVersion.partialVersion(scalaVersion.value) match {
+        case Some((2, y)) => extraDirs("-2") ++ (if (y >= 13) extraDirs("-2.13+") else Nil)
+        case Some((3, _)) => extraDirs("-3") ++ extraDirs("-2.13+")
+        case _            => Nil
+      }
+    }
```
zarthross, line 477: This doesn't appear to be necessary to me?  I removed it and re-ran the tests for Literal and they appeared just fine.   I'm pretty sure these are already being set correctly.

modules/literal/src/main/scala-3/io/circe/literal/JsonLiteralMacros.scala
```diff
@@ -0,0 +1,98 @@
+package io.circe.literal
+
+
```
zarthross, line 3: Looks like we need to run scalafmt on these changes.  I see a few places missing spaces or having the space in the wrong place.  Can you correct that please?

zarthross (commented):
Thanks for the PR, everything looks great, just a few nit picks.

Nit: Can you remove `scala-3/.../LiteralInstances.scala` since its not needed and its empty anyways?

i10416:
Thank you for reviewing. I will fix what you pointed out soon!

modules/literal/src/main/scala-3/io/circe/literal/JsonLiteralMacros.scala
i10416: Sure!

I updated scalafmt to latest for this.

zmccoy:
@i10416 Looks like the build.sbt file isn't formatted with scalaFmt correctly.

i10416:
Oh, I forgot running scalafmtSbt

zarthross (approved):

zmccoy (approved):

build.sbt
```diff
@@ -451,11 +451,11 @@ lazy val literalBase = circeCrossModule("literal", mima = previousCirceVersion,
   .settings(macroSettings)
   .settings(
     libraryDependencies ++= Seq(
-      ("com.chuusai" %%% "shapeless" % shapelessVersion % Test).cross(CrossVersion.for3Use2_13),
       "org.scalacheck" %%% "scalacheck" % scalaCheckVersion % Test,
       "org.scalameta" %%% "munit" % munitVersion % Test,
       "org.scalameta" %%% "munit-scalacheck" % munitVersion % Test
-    )
+    ) ++ (if (isScala3.value) Seq("org.typelevel" %% "jawn-parser" % jawnVersion)
```
armanbilge, line 457: This won't work as-is on Scala.js.

```suggestion
    ) ++ (if (isScala3.value) Seq("org.typelevel" %%% "jawn-parser" % jawnVersion)
```

PR #1875, opened 2021-12-06
Add Scala3 Macro support for Pointer Literals

zarthross:
Thanks @i10416 for #1874 and inspiring this PR.


modules/pointer-literal/src/main/scala-3/io/circe/pointer/literal/PointerLiteralMacro.scala
```diff
@@ -0,0 +1,36 @@
...
+private object PointerLiteralMacros {
+
+  final def pointerImpl(sc: Expr[StringContext], argsExpr: Expr[Seq[Any]])(using q: Quotes): Expr[Pointer] = {
+    import q.reflect.*
+    val stringParts: Seq[String] = sc match {
+      case '{StringContext($parts:_*)} => parts.valueOrAbort
+    }
+
+    if (Pointer.parse(stringParts.mkString("X")).isRight) {
+      val args = argsExpr match {
```
zmccoy, line 16: Looks like formatting is off here.

modules/pointer-literal/src/main/scala-3/io/circe/pointer/literal/PointerLiteralMacro.scala
```diff
@@ -0,0 +1,36 @@
...
+      case Varargs(argExprs) => argExprs
+      case other => report.error("Invalid arguments for pointer literal.");Nil
+    }
+      val input = args.zip(stringParts.tail).foldLeft(Expr(stringParts.head)) {
+        case (acc, (a, p)) => 
+          val pExpr = Expr(p)
+          '{$acc + $a.toString.replaceAll("~", "~0").replaceAll("/", "~1") + $pExpr }
+      }
+
+      val t = '{_root_.io.circe.pointer.Pointer.parse($input)
```
zmccoy, line 26: For my own edification, is there a reason to do this as a value and then return that value compared to just using the expression itself as the last part of the expression?

modules/pointer-literal/src/main/scala-3/io/circe/pointer/literal/PointerLiteralMacro.scala
zarthross: Fixed.

modules/pointer-literal/src/main/scala-3/io/circe/pointer/literal/PointerLiteralMacro.scala
zmccoy: I at least learned it isn't necessary 💯

zmccoy (approved):

PR #1887, opened 2021-12-14
remove unused jekyll and ruby

xuwei-k:


zarthross (approved):
Thanks!  I hadn't noticed that the `ci` step had `jekyll` that only the doc step needed.

zmccoy (approved):
Thank you!

PR #1890, opened 2021-12-14
avoid `scala.App`

xuwei-k:
https://docs.scala-lang.org/scala3/book/methods-main-methods.html

> The previous functionality of App, which relied on the “magic” DelayedInit trait, is no longer available. App still exists in limited form for now, but it doesn’t support command line arguments and will be deprecated in the future.
> 
> If programs need to cross-build between Scala 2 and Scala 3, it’s recommended to use an explicit main method with an Array[String] argument instead:


zarthross (approved):

zarthross:
Thanks for the PR!

The Examples appear to be woefully out of date, and detached from CI.  Something we'll want to fix in the future, I've made a note.

PR #1863, opened 2021-11-21
Backport #1791 Cross-build jawn for Scala.js

armanbilge:
This backports https://github.com/circe/circe/pull/1791 to circe 0.14.x. That PR introduced a cross-build for the jawn module to Scala.js. It makes no changes to the JVM API.

Backporting this would help http4s 0.23 out of a strange situation, where http4s-circe for JVM depends on circe 0.14.x but http4s-circe for JS depends on circe 0.15.0-M1.

Thanks in advance for your consideration! Much appreciated.

PR #1906, opened 2022-01-16
Main is actually 0.14

zarthross:
It looks to me that, with a few minor changes, all the changes in `0.15.0-M1` are compatible with 0.14 series.  Someone feel free to correct me if I'm wrong.

This PR fixes 
- Fixes minor changes in core by adding the needed methods.
- Overrides Mima Warnings for new methods, because those are fine.
- Resets the version to `0.14.2-SNAPSHOT`

This PR base is set to #1905  so it runs the JS tests as well. (Will change when merging that PR).

This supersedes #1863 

## Mima Details:

Overrode these because its really just changes to private memebers.  This might be a problem if used from Java code, but its changes in Macros, so should be safe... and if someone uses it from java... thats on them I think.
```
[error] Circe pointer literal: Failed binary compatibility check against io.circe:circe-pointer-literal_2.13:0.14.0! Found 4 potential problems
[error]  * class io.circe.pointer.literal.PointerLiteralMacros is declared final in current version
[error]    filter with: ProblemFilters.exclude[FinalClassProblem]("io.circe.pointer.literal.PointerLiteralMacros")
[error]  * method c()scala.reflect.macros.blackbox.Context in class io.circe.pointer.literal.PointerLiteralMacros does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("io.circe.pointer.literal.PointerLiteralMacros.c")
[error]  * method pointerStringContext(scala.collection.immutable.Seq)scala.reflect.api.Trees#TreeApi in class io.circe.pointer.literal.PointerLiteralMacros does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("io.circe.pointer.literal.PointerLiteralMacros.pointerStringContext")
[error]  * method this(scala.reflect.macros.blackbox.Context)Unit in class io.circe.pointer.literal.PointerLiteralMacros does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("io.circe.pointer.literal.PointerLiteralMacros.this")
```

Ignored these since they are just adding methods, and we are in early semvar, so thats ok.
```
[error] Circe testing: Failed binary compatibility check against io.circe:circe-testing_2.13:0.14.0! Found 8 potential problems
[error]  * abstract synthetic method io$circe$testing$ArbitraryInstances$_setter_$io$circe$testing$ArbitraryInstances$$arbitraryMissingField_=(org.scalacheck.Gen)Unit in interface io.circe.testing.ArbitraryInstances is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("io.circe.testing.ArbitraryInstances.io$circe$testing$ArbitraryInstances$_setter_$io$circe$testing$ArbitraryInstances$$arbitraryMissingField_=")
[error]  * abstract synthetic method io$circe$testing$ArbitraryInstances$_setter_$io$circe$testing$ArbitraryInstances$$arbitraryWrongTypeExpectation_=(org.scalacheck.Gen)Unit in interface io.circe.testing.ArbitraryInstances is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("io.circe.testing.ArbitraryInstances.io$circe$testing$ArbitraryInstances$_setter_$io$circe$testing$ArbitraryInstances$$arbitraryWrongTypeExpectation_=")
[error]  * abstract synthetic method io$circe$testing$ArbitraryInstances$_setter_$io$circe$testing$ArbitraryInstances$$arbitraryCustomReason_=(org.scalacheck.Gen)Unit in interface io.circe.testing.ArbitraryInstances is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("io.circe.testing.ArbitraryInstances.io$circe$testing$ArbitraryInstances$_setter_$io$circe$testing$ArbitraryInstances$$arbitraryCustomReason_=")
[error]  * abstract synthetic method io$circe$testing$ArbitraryInstances$_setter_$arbitraryReason_=(org.scalacheck.Arbitrary)Unit in interface io.circe.testing.ArbitraryInstances is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("io.circe.testing.ArbitraryInstances.io$circe$testing$ArbitraryInstances$_setter_$arbitraryReason_=")
[error]  * abstract synthetic method io$circe$testing$ArbitraryInstances$$arbitraryMissingField()org.scalacheck.Gen in interface io.circe.testing.ArbitraryInstances is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("io.circe.testing.ArbitraryInstances.io$circe$testing$ArbitraryInstances$$arbitraryMissingField")
[error]  * abstract synthetic method io$circe$testing$ArbitraryInstances$$arbitraryWrongTypeExpectation()org.scalacheck.Gen in interface io.circe.testing.ArbitraryInstances is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("io.circe.testing.ArbitraryInstances.io$circe$testing$ArbitraryInstances$$arbitraryWrongTypeExpectation")
[error]  * abstract synthetic method io$circe$testing$ArbitraryInstances$$arbitraryCustomReason()org.scalacheck.Gen in interface io.circe.testing.ArbitraryInstances is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("io.circe.testing.ArbitraryInstances.io$circe$testing$ArbitraryInstances$$arbitraryCustomReason")
[error]  * abstract method arbitraryReason()org.scalacheck.Arbitrary in interface io.circe.testing.ArbitraryInstances is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("io.circe.testing.ArbitraryInstances.arbitraryReason")
```

Fixed these by adding the missing types and methods:
```
[error] Circe core: Failed binary compatibility check against io.circe:circe-core_2.13:0.14.0! Found 2 potential problems
[error]  * method this(java.lang.String)Unit in class io.circe.DecodingFailure's type is different in current version, where it is (io.circe.DecodingFailure#Reason)Unit instead of (java.lang.String)Unit
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("io.circe.DecodingFailure.this")
[error]  * class io.circe.Printer#AppendableByteBufferFolder does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("io.circe.Printer$AppendableByteBufferFolder")
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -197,28 +197,7 @@ def circeProject(path: String)(project: Project) = {
     description := s"circe $docName",
     moduleName := s"circe-$path",
     name := s"Circe $docName",
-    allSettings,
-    mimaBinaryIssueFilters ++= {
-      import com.typesafe.tools.mima.core._
-      Seq( // Methods were added... this is fine.
-        "io.circe.testing.ArbitraryInstances.io$circe$testing$ArbitraryInstances$_setter_$io$circe$testing$ArbitraryInstances$$arbitraryMissingField_=",
-        "io.circe.testing.ArbitraryInstances.io$circe$testing$ArbitraryInstances$_setter_$io$circe$testing$ArbitraryInstances$$arbitraryWrongTypeExpectation_=",
-        "io.circe.testing.ArbitraryInstances.io$circe$testing$ArbitraryInstances$_setter_$io$circe$testing$ArbitraryInstances$$arbitraryCustomReason_=",
-        "io.circe.testing.ArbitraryInstances.io$circe$testing$ArbitraryInstances$_setter_$arbitraryReason_=",
-        "io.circe.testing.ArbitraryInstances.io$circe$testing$ArbitraryInstances$$arbitraryMissingField",
-        "io.circe.testing.ArbitraryInstances.io$circe$testing$ArbitraryInstances$$arbitraryWrongTypeExpectation",
-        "io.circe.testing.ArbitraryInstances.io$circe$testing$ArbitraryInstances$$arbitraryCustomReason",
-        "io.circe.testing.ArbitraryInstances.arbitraryReason"
-      ).map(ProblemFilters.exclude[ReversedMissingMethodProblem](_)) ++ Seq(
-        // Pointer literals were refactored, but in a binary compatable way. These warnings are just because private stuff changed in ways that
-        // That would not be bin-compat if used from java... but this is Macro Code so it should be fine.
-        ProblemFilters.exclude[FinalClassProblem]("io.circe.pointer.literal.PointerLiteralMacros"),
-        ProblemFilters.exclude[DirectMissingMethodProblem]("io.circe.pointer.literal.PointerLiteralMacros.c"),
-        ProblemFilters
-          .exclude[DirectMissingMethodProblem]("io.circe.pointer.literal.PointerLiteralMacros.pointerStringContext"),
-        ProblemFilters.exclude[DirectMissingMethodProblem]("io.circe.pointer.literal.PointerLiteralMacros.this")
-      )
-    }
+    allSettings
   )
 }
 
@@ -226,7 +205,7 @@ def circeModule(path: String, mima: Option[String]): Project = {
   val id = path.split("-").reduce(_ + _.capitalize)
   Project(id, file(s"modules/$path"))
     .configure(circeProject(path))
-    .settings(mimaPreviousArtifacts := mima.map("io.circe" %% moduleName.value % _).toSet)
+    .settings(mimaPreviousArtifacts := mima.map("io.circe" %%% moduleName.value % _).toSet)
 }
 
 def circeCrossModule(path: String, mima: Option[String], crossType: CrossType = CrossType.Full) = {
@@ -235,8 +214,8 @@ def circeCrossModule(path: String, mima: Option[String], crossType: CrossType = 
     .crossType(crossType)
     .settings(allSettings)
     .configure(circeProject(path))
-    .jvmSettings(
-      mimaPreviousArtifacts := mima.map("io.circe" %% moduleName.value % _).toSet
+    .settings(
+      mimaPreviousArtifacts := mima.map("io.circe" %%% moduleName.value % _).toSet
     )
     .jsSettings(
       coverageEnabled := false,
@@ -338,6 +317,7 @@ lazy val docs = project
   )
   .settings(docSettings)
   .settings(noPublishSettings)
+  .disablePlugins(MimaPlugin)
   .settings(macroSettings)
   .enablePlugins(GhpagesPlugin)
   .enablePlugins(MicrositesPlugin)
@@ -402,6 +382,7 @@ lazy val circe = project
   .in(file("."))
   .settings(allSettings)
   .settings(noPublishSettings)
+  .disablePlugins(MimaPlugin)
   .settings(
     console / initialCommands :=
       """
@@ -577,10 +558,14 @@ lazy val parser = parserBase.jvm
 lazy val parserJS = parserBase.js
 
 lazy val scalajs =
-  circeModule("scalajs", mima = None).enablePlugins(ScalaJSPlugin).settings(jsProjectSettings).dependsOn(coreJS)
+  circeModule("scalajs", mima = previousCirceVersion)
+    .enablePlugins(ScalaJSPlugin)
+    .settings(jsProjectSettings)
+    .dependsOn(coreJS)
 lazy val scalajsJavaTimeTest = circeModule("scalajs-java-time-test", mima = None)
   .enablePlugins(ScalaJSPlugin)
-  .settings(noPublishSettings: _*)
+  .settings(noPublishSettings)
+  .disablePlugins(MimaPlugin)
   .settings(
     libraryDependencies ++= Seq(
       "org.scalameta" %%% "munit" % munitVersion % Test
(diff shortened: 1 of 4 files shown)


armanbilge:
Overall, 👍 if you can maintain bincompat and the 0.14.x series. Without knowing the details of the changes here, a couple questions.

> Overrode these because its really just changes to private memebers. This might be a problem if used from Java code, but its changes in Macros, so should be safe

I have no experience with bincompat for macros, but I always thought this was especially tricky. Since I think you would also want to maintain bincompat for the code _generated_ by the macros, which I don't think MiMa is capable of checking. So even if the changes are purely to private members etc., it still seems like you could break bincompat here. But someone with more expertise on macros should weigh in on this.

> Ignored these since they are just adding methods, and we are in early semvar, so thats ok.

Adding methods is general is not okay. In many instances it is, but an example of when it is not okay is when you are adding an abstract method to an abstract class / trait. Because then, anyone who has extended it will now be missing an implementation of this newly added method. This reads a lot like some of those warnings, e.g. the last one.

```
abstract method arbitraryReason()org.scalacheck.Arbitrary in interface io.circe.testing.ArbitraryInstances is present only in current version
^^^^^^^^                                                     ^^^^^^^^^
```

So, I'd double-check on those warnings before silencing them.

rossabaker:
#1617 is the one that concerned me on a prior analysis.  There is a slight non-breaking change in the error messages of the decoding failures that's probably okay.

You can get around the arbitraries failures by adding the new ones to an Arbitrary object that extends the instances.  In http4s, we phased out the ArbitraryInstances trait and just expect them to be imported.

zarthross:
@armanbilge 
> > Overrode these because its really just changes to private memebers. This might be a problem if used from Java code, but its changes in Macros, so should be safe
> 
> I have no experience with bincompat for macros, but I always thought this was especially tricky. Since I think you would also want to maintain bincompat for the code _generated_ by the macros, which I don't think MiMa is capable of checking. So even if the changes are purely to private members etc., it still seems like you could break bincompat here. But someone with more expertise on macros should weigh in on this.

These are the macro changes in question: https://github.com/circe/circe/pull/1875.  I'm about 99% sure they are fine since they are all private changes. 
 
> > Ignored these since they are just adding methods, and we are in early semvar, so thats ok.
> 
> Adding methods is general is not okay. In many instances it is, but an example of when it is not okay is when you are adding an abstract method to an abstract class / trait. Because then, anyone who has extended it will now be missing an implementation of this newly added method. This reads a lot like some of those warnings, e.g. the last one.
> 
> ```
> abstract method arbitraryReason()org.scalacheck.Arbitrary in interface io.circe.testing.ArbitraryInstances is present only in current version
> ^^^^^^^^                                                     ^^^^^^^^^
> ```
> 
> So, I'd double-check on those warnings before silencing them.

For `arbitraryReason` it does have a default implementation on the trait, you can see it here with the other methods added: https://github.com/circe/circe/pull/1617/files#diff-a7abaecc2b8daa507cce5bb8076e5c62fa683b0f77cfe7183127d5f7c6582edaR95 

@rossabaker , do you think https://github.com/circe/circe/pull/1617 is a big enough change to warrant a bump to 0.15 on its own?  

If any of this is a problem,  I'd be happy to make a 0.15 release that's  'probably' bin-compat with 0.14  :smile:

armanbilge:
> For `arbitraryReason` it does have a default implementation on the trait, you can see it here with the other methods added

Unfortunately, this doesn't matter :) binary-compatibility isn't checking how the source code is written, it's checking how the bytecode is generated. The scala compiler may do surprising things, such as omitting a default implementation from the bytecode even if you've defined it in the the actual source. This is why MiMa is so important. Ross's suggestion for this in https://github.com/circe/circe/pull/1906#issuecomment-1013947942 seemed good.

rossabaker:
I'd rather not ripple the ecosystem on the off chance people were parsing an unstructured string error message.  I like your 0.14 plan.  I'd just be sure to highlight that one in the changelog.

zarthross:
@armanbilge Did some digging with @isomarcte  and you are correct, but I think we can make some changes to fix the bincompat issues.  Thank you for the feedback!

@rossabaker I'll make sure we mention that in the changelog.

build.sbt
```diff
@@ -113,7 +113,7 @@ def priorTo2_13(scalaVersion: String): Boolean =
     case _                              => false
   }
 
-val previousCirceVersion = Some("0.14.0-M3")
+val previousCirceVersion = Some("0.14.0")
```
isomarcte, line 116: It _shouldn't_ matter, but shouldn't this be 0.14.1?

modules/core/shared/src/main/scala/io/circe/Error.scala
```diff
@@ -45,6 +45,11 @@ object ParsingFailure {
  * decoding history resulting in the failure.
  */
 sealed abstract class DecodingFailure(val reason: Reason) extends Error {
+  // Added to satisfy MiMA
+  def this(message: String) = {
```
isomarcte, line 49: Do you want to deprecate this constructor?

isomarcte (approved):

modules/core/shared/src/main/scala/io/circe/Error.scala
rossabaker: If it's just for MiMa, you can make it `private[DecodingFailure]`, but if it was there before, it would be a source-breaking change.  Either way, deprecation sounds smart.

zarthross:
Superseded.

PR #1908, opened 2022-01-21
Fixup CI

zarthross:
- Added Java 11 to build matrix. (Http4s builds on this, we should too)
- Moved Scalafmt Checks to their own actions.
- Set scalafmt `project.git = true` so scalafmt only checks tracked files (no more checking metals.sbt)
- Added Mima Check Actions
- Fixed Mima Checks:
  - They now work on JS
  - We aren't trying to check non-published projects.
- Added `circeJS` and `circeJVM` aggregation projects to help run commands on all JS or all JVM projects separately.

FYI: Mima Checks will fail, that is expected.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -11,43 +11,44 @@ ThisBuild / organization := "io.circe"
 ThisBuild / crossScalaVersions := List(Scala3, "2.12.15", "2.13.8")
 ThisBuild / scalaVersion := crossScalaVersions.value.last
 
-ThisBuild / githubWorkflowJavaVersions := Seq("8", "17").map(JavaSpec.temurin)
+ThisBuild / githubWorkflowJavaVersions := Seq("8", "11", "17").map(JavaSpec.temurin)
 ThisBuild / githubWorkflowPublishTargetBranches := Nil
+ThisBuild / githubWorkflowBuildPreamble +=
+  WorkflowStep.Use(
+    UseRef.Public("actions", "setup-node", "v2.4.0"),
+    name = Some("Setup NodeJS v16"),
+    params = Map("node-version" -> "16"),
+    cond = Some("matrix.ci == 'ciNodeJS'")
+  )
 ThisBuild / githubWorkflowBuild := Seq(
   WorkflowStep.Sbt(
     List(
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -205,7 +235,9 @@ def circeModule(path: String, mima: Option[String]): Project = {
   val id = path.split("-").reduce(_ + _.capitalize)
   Project(id, file(s"modules/$path"))
     .configure(circeProject(path))
-    .settings(mimaPreviousArtifacts := mima.map("io.circe" %% moduleName.value % _).toSet)
+    .settings(
+      mima.map(version => mimaPreviousArtifacts := Set("io.circe" %%% moduleName.value % version)).toSeq
```
armanbilge, line 239: I'm afraid this is not sufficient. You actually need to compare against _all_ previous releases in the current series. See https://github.com/circe/circe/pull/1906#discussion_r788919550.

zarthross: 👍, sorry, noticed your comment on the other PR after I had already pushed this up.  I'll change it up so it does a more though test of the versions when I get a chance.

armanbilge: I'm a broken record, but you should consider using [sbt-typelevel](https://github.com/typelevel/sbt-typelevel/) or at least sbt-typelevel-ci-release. It determines previous MiMa artifacts automatically. It'll also automatically configure your build to cross on JVM/JS like you've done manually here and also fixes the bug that forced you to set `githubWorkflowArtifactUpload := false`.

zarthross: I'll consider switching after we get a release out... right now I'm just trying to get some basic things into CI, not ready to rock the boat by switching the publishing methods and other stuff just yet.

build.sbt
```diff
@@ -392,20 +426,36 @@ lazy val circe = project
       """.stripMargin
   )
   .aggregate(aggregatedProjects: _*)
-  .dependsOn(core, generic, literal, parser)
```
zmccoy, line 395: Maybe add this back for console command.

zmccoy (approved):

zmccoy (approved):

Issue #1912, opened 2022-01-28
JsonCodecMacrosSuite failing


SethTisue:
as of 827794b2ce4c73eae8cdc57386a2e98f2bfe7ed6

```
==> X io.circe.generic.simple.JsonCodecMacrosSuite.@JsonCodec should allow only one, named argument set to true  0.08s munit.ComparisonFailException: /Users/tisue/circe/modules/generic-simple/src/test/scala/io/circe/generic/JsonCodecMacrosSuite.scala:176
175:    // Can't supply both
176:    assertNoDiff(compileErrors("@JsonCodec(encodeOnly = true, decodeOnly = true) case class X(a: Int)"), "")
177:    // Must specify the argument name
diff assertion failed
=> Obtained
    """|error: Unsupported arguments supplied to @JsonCodec
       |@JsonCodec(encodeOnly = true, decodeOnly = true) case class X(a: Int)
       | ^
       |""".stripMargin
=> Diff (- obtained, + expected)
-error: Unsupported arguments supplied to @JsonCodec
-@JsonCodec(encodeOnly = true, decodeOnly = true) case class X(a: Int)
- ^
+
    at munit.FunSuite.assertNoDiff(FunSuite.scala:11)
    at io.circe.generic.simple.JsonCodecMacrosSuite.$anonfun$new$2(JsonCodecMacrosSuite.scala:176)
  + Codec[Seq[Foo]]: codec.roundTrip 0.405s
...
[error] Failed tests:
[error] 	io.circe.generic.simple.JsonCodecMacrosSuite
[error] (genericSimpleJVM / Test / test) sbt.TestsFailedException: Tests unsuccessful
```

not sure when it regressed exactly

(found by the Scala 2 community build)

zmccoy:
Thank you for the heads up @SethTisue.

PR #1913, opened 2022-01-28
Fix jsoncodecmacrosuite tests

zmccoy:
Fixes a few tests that were missed in the munit migration.


zmccoy:
Should fix: https://github.com/circe/circe/issues/1912.  To note as well that this isn't currently published nor in the root module.

zarthross (approved):

Issue #1513, opened 2020-08-05
Add support for Java Locale


ryanb93:
It's not uncommon to pass Java Locale's between clients and web services, we usually end up writing our own encoders per project but this would be nice to have in the core circe `Encoder` and `Decoder` that are provided by default. 

Jackson uses the `.toString` method to serialise and then three different constructors for deserialisation. In Scala this would look something like:

```scala
  implicit final lazy val encodeLocale: Encoder[Locale] = new Encoder[Locale] {
    final def apply(a: Locale): Json = Json.fromString(a.toString)
  }

  implicit final lazy val decodeLocale: Decoder[Locale] = new Decoder[Locale] {
    private[this] def fail(c: HCursor): Result[Locale] = Left(DecodingFailure("Locale", c.history))

    final def apply(c: HCursor): Result[Locale] = c.value match {
      case Json.JString(value) =>
          value.split(Array('_', '-')) match {
            case Array(language)                   => Right(new Locale(language))
            case Array(language, country)          => Right(new Locale(language, country))
            case Array(language, country, variant) => Right(new Locale(language, country, variant))
            case _ => fail(c)
          }
      case _ => fail(c)
    }
  }
```

This round trips with no problems, however `Locale` can be a bit tricky in how it's created as it contains no validation or indication of 'invalid' `Locale` values. 

```
scala> new Locale("dlfkjakldfasd")
val res0: java.util.Locale = dlfkjakldfasd

scala> Locale.forLanguageTag("dlfkjakldfasd")
val res1: java.util.Locale =
```
As a result we usually use the `forLanguageTag` instead so we only accept valid tags, not sure if this is something a JSON library should be opinionated about however.

Has `Locale` ever been considered before for being added?

PR #1925, opened 2022-03-07
Add encoder / decoder for java.util.Locale

cristiborta:
As a follow up for [this](https://github.com/circe/circe/issues/1513#issue-673314224) open issue, since it's something I also need in one of our projects.
Basically, an implicit `encoder` and `decoder` for `java.util.Locale` was added to default implicit `collections`.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -1476,9 +1476,10 @@ object Decoder
         )
     )
 
+  /** Deserializing `java.util.Locale` using a IETF BCP 47 string representation. */
   implicit final lazy val localeDecoder: Decoder[Locale] =
     new Decoder[Locale] {
-      private[this] def fail(c: HCursor): Result[Locale] = Left(DecodingFailure("Locale", c.history))
+      private[this] def fail(c: HCursor): Result[Locale] = Left(DecodingFailure("java.util.Locale", c.history))
       final def apply(c: HCursor): Result[Locale] = c.value match {
         case Json.JString(value) =>
           value match {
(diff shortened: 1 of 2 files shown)


zarthross:
Thanks for the PR @cristiborta.  To be honest, I'm not convinced creating this as the standard encoder is a good idea.  There are multiple ways to specify a locale, and I think we shouldn't necessarily be picking one.  
For instance, the Java `Locale` type specifically uses the form (similar to [BCP 47](https://en.wikipedia.org/wiki/BCP_47)):
`en_US`
But on POSIX ([ISO/IEC 15897](https://en.wikipedia.org/wiki/ISO/IEC_15897)) the `locale` is uses the form:
`en_US.UTF-8` 

I'm not opposed to this on principle, but I think we would want to do 2 this:
1.) Make it complete/standard compliant.
2.) Scope the encoder/decoder to that specific standard. 

Here's a longer example to illustrate:
```scala
val longTag = "ja-JP-u-ca-japanese"
//longTag: String = "ja-JP-u-ca-japanese"

val r = (new java.util.Locale.Builder).setLanguageTag(longTag).build()
//r: java.util.Locale = ja_JP_#u-ca-japanese

r.getDisplayName
//: String = "Japanese (Japan, Japanese Calendar)"
```

cristiborta:
Thank you for your fast reply and for the comments, @zarthross , you are right, I think I should consider the case a bit. Refactored using `forLanguageTag` and `toLanguageTag`, and that should cover all possible concerns. Let me know what you think

cristiborta:
Scoped the encoder/decoder for `Locale`. Please have a look @zarthross

modules/core/shared/src/main/scala/io/circe/Decoder.scala
```diff
@@ -1477,6 +1476,29 @@ object Decoder
         )
     )
 
+  /** Deserializing `java.util.Locale` using a IETF BCP 47 string representation. */
+  object locale {
```
zarthross, line 1480: I meant something more along the lines of:

```
object locale {
  object IETF_BCP_47 {
     implicit final lazy val decoder: Decoder[Locale] ...
  }
}
```

so end users will do
`import io.circe.Decoder.locale.IETF_BCP_47._`

zarthross:
Also, please do the same for the `Encoder`, as well as adding some tests.

Issue #1077, opened 2019-02-16
Scala Native Support


sorenbug:
Currently seems to be blocked by https://github.com/typelevel/cats/issues/1549?

(Primarily made this issue so I can link to it from other projects that I'd like to see a Scala Native discussion on.)

travisbrown:
Thanks @sorenbug, but yes, this is blocked for the foreseeable future.

travisbrown:
Just for the record: I'm planning to drop 2.11 support in an upcoming release and will not include Scala Native support (even if Cats eventually does) until it works for more recent Scala versions.

sorenbug:
Sounds good.

travisbrown:
Closing for now, we can revisit if / when Cats supports Scala Native and Scala Native supports a modern Scala version.

joroKr21:
> Closing for now, we can revisit if / when Cats supports Scala Native and Scala Native supports a modern Scala version.

They do!

sideeffffect:
FYI, Scala Native support has been merged to jawn https://github.com/typelevel/jawn/pull/340

armanbilge:
Scala Native support is landing in https://github.com/circe/circe/pull/2017.

PR #1929, opened 2022-03-21
[WIP] Add support for Scala native

adampauls:
Starts to address https://github.com/circe/circe/issues/1077, which was closed years ago but I would re-open if I could. 

I thought I would put this up as proof that there isn't too much to do. The outstanding issues are:
- [x] Fix issue in `testingNative` where `io.circe.Spaces2PrinterSuite` cannot read `modules/tests/shared/src/test/resources/io/circe/spaces2-example.json` (but `testingJVM` can)
- [x] See if `"scodec-bits"` dependency can support native.
- [ ] See if `"eu.timepit" %%% "refined"` can support native.
- [ ] Implement CopyOnWriteArrayList properly or find a workaround.

I would love to see this done but don't know the project well enough to solve all these problems quickly. I thought I would put this draft PR up for discussion. Hope that's alright!


armanbilge:
Nice! I've been interested in this as well as @lolgab. Now that Native supports Scala 3 I'm hoping that more of the ecosystem will adopt, including scodec. All of that is basically blocked from here:
- https://github.com/typelevel/scalacheck/pull/879

adampauls:
Is it reasonable to publish a native version without support for `refined` and `scodec`?

armanbilge:
It is reasonable. It is also reasonable to publish without Scala 3. But, we are so close! Let's do it right, and put our energy into driving this from the bottom up. For example, I just added support for shared source folder e.g. `js-native` so that you don't have to copy files like you did here.
- https://github.com/portable-scala/sbt-crossproject/pull/145#pullrequestreview-916094636

adampauls:
Sure thing, whatever you think is best! 

BTW I just pushed a fix and `sbt validateNative` seems to work now, so I think there aren't many unknowns left.

adampauls:
It looks like the upstream efforts are stalled. Do you think it's worth pushing something out without scodec-bits, Scala 3, or refined?

armanbilge:
Depends what you mean by pushing something out :) I'd say upstreams and circe are moving at roughly the same pace e.g. see https://github.com/circe/circe/issues/1933.

adampauls:
Heh, good question. I think what I meant was to submit a PR like this one if I can figure out a way to simply exclude the non-native-ready parts of the code from the native build. I know it's not great practice to have the native build support less functionality than the JVM/JS builds, but it's also better than nothing.

fiatjaf:
Good morning, just adding here that `scodec-bits` Scala Native support has been merged: https://github.com/scodec/scodec-bits/pull/373

armanbilge:
Cats is now published for Native on Scala 2.12, 2.13, 3. So Circe should be able to do the same.

https://github.com/typelevel/cats/releases/tag/v2.8.0

Note that my build changes in https://github.com/circe/circe/pull/1968 will make configuring the build for Native very easy.

armanbilge:
@adampauls were you ever actually able to run the tests? I started looking into this myself and it seems that Native is missing `CopyOnWriteArrayList`. This should be ported to Scala Native from Scala.js before we can move ahead. Would be fantastic if you (or any motivated on-lookers) can take that on. Thank you!

https://github.com/scala-js/scala-js/blob/33ed7588ad80eaff5802aaf3b6f620b408a200cd/javalib/src/main/scala/java/util/concurrent/CopyOnWriteArrayList.scala

adampauls:
I am on vacation at the moment but can try to do it soon!

armanbilge:
I already opened https://github.com/scala-native/scala-native/pull/2725. I'm sorry, I missed the fact that it was on your checklist you made in March. Fixing this should have been #1 priority since it affects the Circe Core module and requires a change in Scala Native itself.

armanbilge:
For anyone who wants to play around I've published a bootleg snapshot of Circe Native.
https://s01.oss.sonatype.org/content/repositories/snapshots/com/armanbilge/circe-core_native0.4_3/0.14.2-87-a48d472-SNAPSHOT/

armanbilge:
I now have a PR up at https://github.com/circe/circe/pull/2017.

zarthross:
Superseded by 2017

PR #1931, opened 2022-03-27
use `lengthCompare`

xuwei-k:


zarthross (approved):

zmccoy:
Thank you!  

For reasoning:  lengthCompare is more efficient.   For some collection types computing the length is a O(N) operation, instead of O(1).  This makes it O(M) worst case where M is the size you are comparing instead of N (length of the list).

zmccoy (approved):

PR #1948, opened 2022-05-03
Make main Bincompat with v0.14

zarthross:
It looks to me that, with a few minor changes, all the changes in `0.15.0-M1` are compatible with 0.14 series.  Someone feel free to correct me if I'm wrong.

This PR fixes 
- Fixes minor changes in core by adding the needed methods.
- Correctly sets mima based on scala 3 support
- Resets the version to `0.14.2-SNAPSHOT`
- tried to fix 'extras' but disabled publish instead.  It was published as an empty artifact before, and the tests need fixing, we might want to give it an evaluation before we actually publish it.

Call outs:
- The decoder failure messages change which could cause some issues, I'm fine with releasing this as 0.15 if we want to do that, but say its binary compatible with 0.14.x

This supersedes https://github.com/circe/circe/pull/1906


modules/core/shared/src/main/scala/io/circe/Error.scala
```diff
@@ -45,6 +45,11 @@ object ParsingFailure {
  * decoding history resulting in the failure.
  */
 sealed abstract class DecodingFailure(val reason: Reason) extends Error {
+  // Added to satisfy MiMA
+  def this(message: String) = {
```
armanbilge, line 49: Does this deserve a deprecation message?

armanbilge: Ah, I guess a constructor on a `sealed abstract class` isn't publicly usable anyway. 👍

armanbilge: Oh lol, we've been through this already in https://github.com/circe/circe/pull/1906#discussion_r788915361.

zarthross: Thanks.  Been too long, i should have reviewed that previous PR.

zmccoy (approved):

isomarcte (approved):

PR #1963, opened 2022-05-24
Make scala-steward update all scalasteward PRs.

zarthross:
We have a bunch of scala-steward PRs that are out of date, 
the mima checks should be passing now.  This change should
get scala-steward to update those PRs for us.


armanbilge:
Also, you can just close/re-open those PRs to get CI to re-run on a fresh merge commit.

zarthross:
Thanks @armanbilge.   I'll give that a try.  It does seem a little weird, but if it works it works.  I'm going to go ahead with the scala-steward.conf change too, so its more automated.

PR #1967, opened 2022-05-26
Make Codec.apply take implicit Decoder and Encoder

ikeyan:
Currently, Codec.apply searches for `Codec` instance, but [this comment tells you should not do it](https://github.com/circe/circe/blob/main/modules/core/shared/src/main/scala/io/circe/Codec.scala#L10-L11).

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Codec.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Codec.scala
@@ -26,7 +26,13 @@ trait Codec[A] extends Decoder[A] with Encoder[A] {
 }
 
 object Codec extends ProductCodecs with EnumerationCodecs {
-  def apply[A](implicit instance: Codec[A]): Codec[A] = instance
+  def apply[A](implicit decoder: Decoder[A], encoder: Encoder[A]): Codec[A] =
+    decoder match {
+      case codec: Codec[_] if codec eq encoder =>
+        // if codec: Decoder[A] with Decoder[B], A and B must be identical
+        codec.asInstanceOf[Codec[A]]
+      case _ => from(decoder, encoder)
+    }
 
   implicit val codecInvariant: Invariant[Codec] = new Invariant[Codec] {
     override def imap[A, B](fa: Codec[A])(f: A => B)(g: B => A): Codec[B] = Codec.from(fa.map(f), fa.contramap(g))


Lasering:
See also #1495

modules/core/shared/src/main/scala/io/circe/Codec.scala
```diff
@@ -26,7 +26,13 @@ trait Codec[A] extends Decoder[A] with Encoder[A] {
 }
 
 object Codec extends ProductCodecs with EnumerationCodecs {
-  def apply[A](implicit instance: Codec[A]): Codec[A] = instance
+  def apply[A](implicit decoder: Decoder[A], encoder: Encoder[A]): Codec[A] =
+    decoder match {
+      case codec: Codec[_] if codec eq encoder =>
+        // if codec: Decoder[A] with Decoder[B], A and B must be identical
```
hamnis, line 32: is this comment correct? 
Should not this be something like:
// if codec: Decoder[A] with Encoder[A], encoder and decoder must be the same instance

ikeyan: I confirmed that you cannot create a class which extends Decoder[A] and Decoder[B] at the same time where A !== B.
```scala
  class Animal {
    def name: String = "animal"
  }
  class Dog extends Animal {
    override def name: String = "dog"
  }
  class AnimalDecoder extends Decoder[Animal] {
    override def apply(c: HCursor): Decoder.Result[Animal] = ???
  }
  // illegal inheritance;
  //  self-type LtiCodecs.this.DogDecoder does not conform to io.circe.Decoder[LtiCodecs.this.Dog]'s selftype io.circe.Decoder[LtiCodecs.this.Dog]
  class DogDecoder extends AnimalDecoder with Decoder[Dog] {
    override def apply(c: HCursor): Decoder.Result[Dog] = ???
  }

  trait AnimalDecoderTrait extends Decoder[Animal] {
    override def apply(c: HCursor): Decoder.Result[Animal] = ???
  }
  trait DogDecoderTrait extends Decoder[Dog] {
    override def apply(c: HCursor): Decoder.Result[Dog] = ???
  }
  // illegal inheritance;
  //  class AnimalAndDogDecoderTrait inherits different type instances of trait Decoder:
  // io.circe.Decoder[LtiCodecs.this.Dog] and io.circe.Decoder[LtiCodecs.this.Animal]
  trait AnimalAndDogDecoderTrait extends AnimalDecoderTrait with DogDecoderTrait {}
```

Therefore:
decoder: Decoder[A] && decoder eq codec && codec: Codec[_]
⇒ ∃B. decoder: Decoder[A] & Codec[B]
⇒ ∃B. decoder: Decoder[A] & Decoder[B] & Codec[B]
⇒ ∃B. decoder: Decoder[A] & Codec[A] && A == B
⇒ decoder: Codec[A]

ikeyan: Same situation: https://www.scala-lang.org/old/node/3423

ikeyan: @hamnis

hamnis: Well yes, this is expected.

Why one would extend the same type twice is basically what my comment is about.

hamnis:
We should not do this and rather have an unsafeSummon as suggested in #1495

Issue #1911, opened 2022-01-24
io.circe.NoSpacesSortKeysPrinterSuite flakey spec


zmccoy:
On the JS run this test was seen failing.  I haven't looked into it to se if it's related to JS or affects all and is just a good input to see a failure.  

`==> X io.circe.NoSpacesSortKeysPrinterSuite.Printer with sortKeys should sort the object keys 0.26s munit.FailException: /home/runner/work/circe/circe/modules/tests/shared/src/test/scala/io/circe/SortedKeysSuite.scala:30 assertion failed
29:      val keys = parsed.keys.toVector
30:      assert(keys.sorted === keys)
31:    }

Failing seed: Q77xP9WJjYRbRkQLYsOsBNr-Ex78IBI9PAVcu7f3fZP=
You can reproduce this failure by adding the following override to your suite:

  override val scalaCheckInitialSeed = "Q77xP9WJjYRbRkQLYsOsBNr-Ex78IBI9PAVcu7f3fZP="

Falsified after 88 passed tests.
> ARG_0: Map(4 -> List(),  -> List())
[...]

PR #1973, opened 2022-06-11
Permit Parsing Of Leading Zeros

isomarcte:
This commit changes `BiggerDecimal.parseBiggerDecimalUnsafe` to permit parsing of leading '0' characters, and by extension allows all of our numeric codecs to tolerate decoding `JsonString` values which represent numbers with leading zeros. This is more in line with the `fromString` methods on the various numeric classes.

Note, this change should be given some serious scrutiny. It appears that the original authors had at some time explicitly forbade this behavior, as noted by the existence of the `AFTER_ZERO` state as well as the explicit test case for invalid input including the `String` `"01"`.

I've searched around the issues and I couldn't find any reasoning for this decision, so I still think being lenient here is the right choice, but please double check me.


isomarcte:
@zarthross I can't seem to reproduce that failure locally, and I don't think it is related to this PR.

armanbilge:
Perhaps it was https://github.com/circe/circe/issues/1911 you saw?

isomarcte:
@zarthross thoughts here?

modules/numbers/shared/src/main/scala/io/circe/numbers/BiggerDecimal.scala
```diff
@@ -263,132 +262,115 @@ object BiggerDecimal {
...
-        if (i >= len) FAILED
-        else {
-          if (input.charAt(i) != '0') START
-          else {
-            i = i + 1
-            AFTER_ZERO
+      var parsedNonLeadingDigit: Boolean = false
+
+      if (i >= len) {
+        null
```
zarthross, line 268: I don't mind the short cut here, but we need to make it more clear this is a FAILED state.
```suggestion
        null // state = FAILED
```

modules/numbers/shared/src/main/scala/io/circe/numbers/BiggerDecimal.scala
```diff
@@ -4,6 +4,7 @@ import java.io.Serializable
 import java.lang.StringBuilder
 import java.math.{ BigDecimal, BigInteger }
 import scala.annotation.switch
+import scala.annotation.tailrec
```
zarthross, line 7: This isn't needed:

```suggestion
```

zarthross (approved):
Looks good, just 2 minor things I'd like fixed.

modules/numbers/shared/src/main/scala/io/circe/numbers/BiggerDecimal.scala
isomarcte: Comment added, FYI that is also listed in the ScalaDoc. 

I don't love using `null` here. I don't think the single boxing of one `Either` is a problematic cost, but this is most similar to the existing API.

modules/numbers/shared/src/main/scala/io/circe/numbers/BiggerDecimal.scala
isomarcte: :+1:

zmccoy (approved):
Other than the minor naming nitpick I think this looks good.

modules/numbers/shared/src/main/scala/io/circe/numbers/BiggerDecimal.scala
```diff
@@ -263,132 +261,115 @@ object BiggerDecimal {
...
       var i = if (input.charAt(0) == '-') 1 else 0
-
-      var state =
-        if (i >= len) FAILED
-        else {
-          if (input.charAt(i) != '0') START
-          else {
-            i = i + 1
-            AFTER_ZERO
+      var parsedNonLeadingDigit: Boolean = false
```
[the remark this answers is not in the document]
isomarcte: updated

travisbrown:
For the record I think accepting non-standard JSON is a mistake (except maybe in some rare cases via a custom parser, but definitely not the default parser).

zarthross:
Hey Travis, while I would generally agree with you, this is only impacting the string based json numbers decoding (which I don’t think is in the spec), and not json parsing.  I think this change is fine, especially given the stdlib parse methods like `BigDecimal.apply` accept the leading zeros just fine when handling strings.

PR #1974, opened 2022-06-14
Add `pathString` To ACursor

isomarcte:
This allows for computing the JavaScript-style path, as is currently done in `CursorOp.opsToPath`, but because it has direct access to the cursor history (e.g. `lastCursor`), the computation is much more efficient.

I'm also reasonably certain that `opsToPath` is broken for certain cursor history values, and I do not believe `pathString` is.

@zarthross @zmccoy This still needs to have `opsToPath` deprecated and some other bookkeeping done, but I wanted to get your thoughts before I went too far.


modules/testing/shared/src/main/scala/io/circe/testing/ArbitraryInstances.scala
```diff
@@ -127,6 +127,34 @@ trait ArbitraryInstances extends ArbitraryJsonNumberTransformer with CogenInstan
   implicit def arbitraryAsArrayEncoder[A: Cogen]: Arbitrary[Encoder.AsArray[A]] = Arbitrary(
     Arbitrary.arbitrary[A => Vector[Json]].map(Encoder.AsArray.instance)
   )
+
+  // implicit def arbitraryACursor(implicit A: Arbitrary[Json]): Arbitrary[ACursor] = {
```
isomarcte, line 131: Oops. I meant to remove this. 

I was playing around with generating arbitrary (successful) cursor traversals, but I didn't finish it.

isomarcte:
@zarthross or @zmccoy can you restart the CI? Test failure is unrelated and fixed in #1975

isomarcte:
@zarthross @zmccoy this is ready for review.

isomarcte:
@lzm0 I'm so glad to hear that!

isomarcte:
The most recent failure is unrelated to the changes in this PR. It is fixed in #1991. Once #1991 is merged, I'll merge this branch with main and we can continue here.

zarthross:
@isomarcte #1991 Is merged.

zarthross:
I went ahead and merged in main, but it appears this breaks bincompat.

modules/core/shared/src/main/scala/io/circe/Error.scala
```diff
@@ -52,32 +52,36 @@ sealed abstract class DecodingFailure(val reason: Reason) extends Error {
...
-    case MissingField                       => "Missing required field"
-    case CustomReason(message)              => message
-  }
+  /**
+   * The location in the JSON where the decoding failure occurred. For various
+   * reasons, in 0.14.x this may not always be present.
+   */
+  def pathToRootString: Option[String] = None
+
+  final def message: String =
```
armanbilge, line 61: I think MiMa is complaining about this. Since this is a `sealed` class I think this would be okay to add an exception for. But it's also easy enough to remove the `final` modifier.

```suggestion
  def message: String =
```

zarthross: I think I'd prefer removing `final` rather than adding an exception.

isomarcte:
This is an interesting failure on the latest run.

```
 ==> X io.circe.JavaTimeCodecSuite.Codec[ZonedDateTime]: codec.roundTrip0.09s munit.FailException: /home/runner/work/circe/circe/modules/tests/shared/src/test/scala/io/circe/JavaTimeCodecSuite.scala:117
116:  checkAll("Codec[LocalDateTime]", CodecTests[LocalDateTime].codec)
117:  checkAll("Codec[ZonedDateTime]", CodecTests[ZonedDateTime].codec)
118:  checkAll("Codec[OffsetDateTime]", CodecTests[OffsetDateTime].codec)
Failing seed: jzluEHo31Jf7maMiO43ajGrviCAw7HF6IIkW1ZOfWRJ=
  override val scalaCheckInitialSeed = "jzluEHo31Jf7maMiO43ajGrviCAw7HF6IIkW1ZOfWRJ="
Falsified after 66 passed tests.
> Labels of failing property: 
Expected: Right(2815-04-05T02:13:08+12:00[NZ])
Received: Right(2815-04-05T02:13:08+13:00[NZ])
> ARG_0: 2815-04-05T02:13:08+12:00[NZ]
    at munit.Assertions.fail(Assertions.scala:261)
    at munit.Assertions.fail$(Assertions.scala:255)
    at munit.FunSuite.fail(FunSuite.scala:11)
    at munit.ScalaCheckSuite.$anonfun$propToTry$4(ScalaCheckSuite.scala:104)
    at scala.util.Try$.apply(Try.scala:213)
    at munit.ScalaCheckSuite.propToTry(ScalaCheckSuite.scala:104)
    at munit.ScalaCheckSuite.$anonfun$scalaCheckPropTransform$3(ScalaCheckSuite.scala:46)
    at scala.concurrent.impl.Promise.liftedTree1$1(Promise.scala:33)
    at scala.concurrent.impl.Promise.$anonfun$transform$1(Promise.scala:33)
    at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:64)
    at munit.Suite$$anon$1.execute(Suite.scala:26)
    at scala.concurrent.impl.CallbackRunnable.executeWithValue(Promise.scala:72)
    at scala.concurrent.impl.Promise$KeptPromise$Kept.onComplete(Promise.scala:372)
    at scala.concurrent.impl.Promise$KeptPromise$Kept.onComplete$(Promise.scala:371)
    at scala.concurrent.impl.Promise$KeptPromise$Successful.onComplete(Promise.scala:379)
    at scala.concurrent.impl.Promise.transform(Promise.scala:33)
    at scala.concurrent.impl.Promise.transform$(Promise.scala:31)
    at scala.concurrent.impl.Promise$KeptPromise$Successful.transform(Promise.scala:379)
    at munit.internal.FutureCompat$ExtensionFuture.transformCompat(FutureCompat.scala:16)
    at munit.ScalaCheckSuite.$anonfun$scalaCheckPropTransform$2(ScalaCheckSuite.scala:48)
    at munit.GenericTest.$anonfun$withBodyMap$1(GenericTest.scala:33)
    at munit.MUnitRunner.$anonfun$runTestBody$1(MUnitRunner.scala:296)
    ```

File: modules/core/shared/src/main/scala/io/circe/Error.scala (hunk 2 above)
Comment (isomarcte) [reply modules/core/shared/src/main/scala/io/circe/Error.scala:61]: Updated

**Comment by isomarcte** [comment]
RE that Java time failure. Apparently the implementation for parsing of ISO 8601 time values with a zone id (which is not part of ISO 8601/RFC-3339, but a common extension) will change the value of the offset to match the zone id in the event that the offset disagrees with the time zone.

This is avoided by the default arbitrary instances which are provided in ScalaCheck. I'll make a PR to switch to those.

**Comment by isomarcte** [comment]
To elaborate, the reason we got the mismatch is that `ZonedDateTime.parse` repairs the offset time to match the zone id, but you can still _manually_ construct a `ZonedDateTime` with an offset which does not match the zone id, which is what the circe Arbitrary instances were doing.

**Comment by armanbilge** [comment]
I mean, we don't have to wait until 0.15.x, just until the next ScalaCheck release :)

**Comment by isomarcte** [comment]
@armanbilge @zarthross the issue isn't what I thought it was originally. I think it might be related to some discontinuity in local time at the exact moment this `ZoneDateTime` is pointing too. See https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/time/zone/ZoneOffsetTransition.html (or don't you might lose your mind).

This is one reason I chose to use `ZoneOffset` rather than `ZoneId` for these in ScalaCheck. I'm going to PR that change here.

**Comment by isomarcte** [comment]
Test failure is unrelated and fixed in #1997

File: modules/core/shared/src/main/scala/io/circe/Decoder.scala [hunk 3]
```diff
@@ -87,15 +87,17 @@ trait Decoder[A] extends Serializable { self =>
...
-    } else {
-      DecodingFailure(MissingField, history)
-    }
+    val reason: DecodingFailure.Reason =
+      cursor match {
+        case cursor: FailedCursor =>
+          if (cursor.missingField) {
+            DecodingFailure.Reason.MissingField
+          } else {
+            DecodingFailure.Reason.CustomReason("Unable to decode JSON")
```
zarthross, line 96: This change in reason for the non-missing field case could cause issues... just like when we changed them for 0.14.2.   Would the performance issue we are trying to solve be that bad if we keep this message the same?

modules/core/shared/src/main/scala/io/circe/PathToRoot.scala
```diff
@@ -0,0 +1,117 @@
...
+    } else {
+      path.value
+        .foldLeft((new StringBuilder(path.value.size * 5), true)) {
+          case ((sb, _), ObjectKey(keyName)) =>
+            (sb.append(".").append(keyName), false)
+          case ((sb, _), ArrayIndex(index)) =>
+            (sb.append("[").append(index.toString).append("]"), false)
+        }
+        ._1
+        .toString
```
zarthross, line 77: Maybe I'm just completely missing something, but whats the `Boolean` half of the fold doing here? It appears to be ignored on both pattern matches and the output?

modules/core/shared/src/main/scala/io/circe/PathToRoot.scala
```diff
@@ -0,0 +1,117 @@
...
+
+  def fromHistory(ops: List[CursorOp]): Either[String, PathToRoot] = {
+    type F[A] = Either[String, A] // Kind Projector?
+
+    ops.reverse
+      .foldM[F, Vector[PathElem]](Vector.empty[PathElem]) {
+        case (acc :+ PathElem.ArrayIndex(n), CursorOp.MoveLeft) if n > 0 =>
+          Right(acc :+ PathElem.ArrayIndex(n - 1))
+        case (acc :+ PathElem.ArrayIndex(n), CursorOp.MoveLeft) =>
+          Left("Attempt to move beyond beginning of array in cursor history.")
```
zarthross, line 89: I think if we make the fail condition more explicit, and the success the fail through case, this is a bit easier to read.

```suggestion
        case (acc :+ PathElem.ArrayIndex(n), CursorOp.MoveLeft) if n <= 0 =>  
          Left("Attempt to move beyond beginning of array in cursor history.")
        case (acc :+ PathElem.ArrayIndex(n), CursorOp.MoveLeft) =>
          Right(acc :+ PathElem.ArrayIndex(n - 1))
```

modules/core/shared/src/main/scala/io/circe/PathToRoot.scala
```diff
@@ -0,0 +1,117 @@
...
+        case (acc :+ PathElem.ArrayIndex(n), CursorOp.MoveRight) if n < Int.MaxValue =>
+          Right(acc :+ PathElem.ArrayIndex(n + 1))
+        case (acc :+ PathElem.ArrayIndex(n), CursorOp.MoveRight) =>
+          Left("Attempt to move to index > Int.MaxValue in array in cursor history.")
```
zarthross, line 93: Same note as above.
```suggestion
        case (acc :+ PathElem.ArrayIndex(n), CursorOp.MoveRight) if n >= Int.MaxValue =>
          Left("Attempt to move to index > Int.MaxValue in array in cursor history.")
        case (acc :+ PathElem.ArrayIndex(n), CursorOp.MoveRight)  =>
          Right(acc :+ PathElem.ArrayIndex(n + 1))
```

modules/core/shared/src/main/scala/io/circe/PathToRoot.scala
```diff
@@ -0,0 +1,117 @@
...
+        case (acc :+ _, CursorOp.MoveUp) =>
+          Right(acc)
+        case (acc, CursorOp.MoveUp) =>
+          Left("Attempt to move up, but never descended in cursor history.")
```
zarthross, line 97: I think the text is a bit missleading because `c.down.up.up` would yield this message too. Maybe something like: 

```suggestion
        case (acc, CursorOp.MoveUp) =>
          Left("Attempt to move up above the root of the json.")
```

modules/core/shared/src/main/scala/io/circe/PathToRoot.scala
```diff
@@ -0,0 +1,117 @@
...
+        case (acc, CursorOp.DownField(name)) =>
+          Right(acc :+ PathElem.ObjectKey(name))
+        case (acc, CursorOp.DownArray) =>
+          Right(acc :+ PathElem.ArrayIndex(0))
+        case (acc, CursorOp.DownN(n)) =>
+          Right(acc :+ PathElem.ArrayIndex(n))
+        case (acc :+ _, CursorOp.DeleteGoParent) =>
+          Right(acc)
+        case (acc :+ _, CursorOp.DeleteGoParent) =>
+          Left("Attempt to move up, but never descended in cursor history.")
```
zarthross, line 111: These are the same 2 patterns, so the 2nd one never matches?  Also i don't think the wording of the 2nd pattern matches the pattern?

zarthross (changes requested):

modules/core/shared/src/main/scala/io/circe/Decoder.scala
zarthross: I know we made a bit of a headache for some users, so wondering what can be done about that... hate to thrash too much 😅

modules/core/shared/src/main/scala/io/circe/PathToRoot.scala
isomarcte: @zarthross oh nothing. Sorry, I meant to remove it. 

So, there are two syntax styles that come up for this in `CursorOp` and in `DecodingFailure`. 

Given,

```json
{
    "foo": [{"bar": 1}]
}
```

Pointing to `1` would either look like `foo[1].bar` _or_ `.foo[1].bar` (note the leading `.`).

I was using the `Boolean` to trigger the special behavior of omitting the leading `.`. However I decided that was a really arbitrary distinction. 

For 0.15.x I'm going to propose that we scrap all that and use the JSON pointer syntax as it actually has defined rules, rather than a pseudo standard.

I'll get that fixed.

modules/core/shared/src/main/scala/io/circe/PathToRoot.scala
isomarcte: I'll fiddle with it and post back.

modules/core/shared/src/main/scala/io/circe/PathToRoot.scala
isomarcte: 👍

zarthross (approved):

modules/core/shared/src/main/scala/io/circe/Error.scala
```diff
@@ -44,8 +44,14 @@ object ParsingFailure {
  * An exception representing a decoding failure and (lazily) capturing the
  * decoding history resulting in the failure.
  */
-sealed abstract class DecodingFailure(val reason: Reason) extends Error {
-  @deprecated("use a DecodingFailure.Reason", since = "0.14.2")
+sealed abstract class DecodingFailure(private val lazyReason: Eval[Reason]) extends Error {
+
+  @deprecated("use a DecodingFailure.Reason", since = "0.14.3")
```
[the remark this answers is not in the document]
isomarcte: Oh, I should actually remove that. I think we can permit people to pass in a strict `Reason`.

zmccoy (approved):

andreisilviudragnea:
Hello. I want to make a few comments regarding the work in this PR:
- `This allows for computing the JavaScript-style path, as is currently done in CursorOp.opsToPath, but because it has direct access to the cursor history (e.g. lastCursor), the computation is much more efficient.` -> Is there a benchmark test justifying why `CursorOp.opsToPath` is much more inefficient?
- `I'm also reasonably certain that opsToPath is broken for certain cursor history values, and I do not believe pathString is.` -> Why not just write a test that shows those cursor history values that `CursorOp.opsToPath` cannot handle and the new implementation can.

Apart from my comments above, I believe this PR has complicated the error handling and reporting logic of circe without adding any real benefit. The new implementation is still broken and now the code is more complex. I tried to revert and fix some of the changes of this PR here: https://github.com/circe/circe/pull/2030.

@zmccoy @zarthross @jeffmay @travisbrown I want your opinion about the desired quality of the code of this repo, because I feel this PR did not do any good. There is also another one in progress here https://github.com/circe/circe/pull/2015 that again I do not believe is in the right direction for this project. I also believe big changes like these need to be discussed more beforehand.

zarthross:
@andreisilviudragnea
 1. I've posted benchmarks [here](https://github.com/circe/circe/pull/2030#issuecomment-1271774405).   If you look at how the cursor history works for sequence decoders, you'll see there there is a `MoveRight` operation performed for every array element parsed.  This results in a cursor history that can be far longer than the trail from the current cursor, by parents, up to the top cursor.  By definition looking at the cursors parents to find the path will be faster.  (benchmarks in the referenced issues proves this for the general case)
 2. Yes.. its broke for certain cursor histories involving `DeleteGoParent`.  We should have had a few more tests to point out how this made improvements, and possibly a few  ignored tests showing the deficient of both approaches.   I welcome any PR to improve on this.
 3. This PR did make improvements to the area of performance, even if it might have introduced its own bugs. (See response to your PR.)
 
Naturally, we want high quality code in the Circe project, but we are all humans, and we err... no one is perfect. Rather than coming back to an old PR and criticizing the works of others that have come before you, its better to learn why they did what they did and then to offer improvements.

Regarding #2015, if you feel like that's not the right direction for this project, you are completely free to comment on that issue with your exact concerns, provide feedback, that is the entire point of PRs.  ` I also believe big changes like these need to be discussed more beforehand.`  That's exactly what  PR is.. time to discuss these changes.  This PR was here for over a month, with comments from multiple people before it was merge.  #2015 has likewise been sitting for over a month, and is not even targeting a main series branch, but an experimental branch... now is the time to comment.

Regarding Travis and Jeff... Travis is no longer participating in open source Scala work, so I'll ask that you refrain from tagging him in the future, and Jeff only has admin access to the circe-yaml project.

PR #1991, opened 2022-07-10
Fix The PointerInterpolatorSuite For All Scala Versions

isomarcte:
This takes the fix from https://github.com/circe/circe/pull/1975 and ports it to all Scala versions. At the time I wrote the original fix, I didn't realize we had this class split for the Scala 2/3.

To fix this issue for all Scala versions, I added `genPointerReferenceString` which generates a valid JSON pointer reference value, properly handling the escape.

Note, there is probably still more to do here in the future. The tests are slightly different in their structure for Scala 2/3 and from glancing at the RFC I suspect there are issues with the current pointer implementation when a reference contains Unicode characters which are represented by surrogate pairs of bytes under UTF-8. However, fixing those would be a more intrusive change and thus reserved for a different PR.


armanbilge:
Is there a reason the _tests_ have to be split across Scala 2/3? This means we're not necessarily testing cross-compilability of downstream projects using these APIs.

isomarcte:
@armanbilge the way that the pointer literal macro builds the enclosing position `String` varies between Scala 2 and 3 macros. I believe that is the reason for the split. So the tests of the form, `assertNoDiff(compileErrors`, have different diffs on 2/3.

If we dropped the requirement of testing the format of the error `String`, then I think we could easily unify this test class.

armanbilge:
I see, thanks. While version-specific tests are make sense in this case I also think some common tests would be good as well, otherwise it's easy to fall into the trap of only changing one implementation (and causing a divergence in behavior). But perhaps this can be one of the follow-ups you alluded it.

zarthross (approved):

zarthross:
I agree with @armanbilge ... having a common set a of tests would be fantastic.  If only the errors differ maybe those can be in a split trait that is used by the common tests so we have better parity.

PR #1992, opened 2022-07-11
Unify The Majority Of Pointer Literal Tests

isomarcte:
Only the tests which check the compile error string are specialized to specific Scala versions now. This is because the way that Scala 2/3 macros build up the string describing the position is slightly different.


isomarcte:
As requested @armanbilge and @zarthross

armanbilge (approved):
LGTM. Thank you for following-up on that!

Issue #1994, opened 2022-07-11
Drop Custom Arbitrary Instances For java.time


isomarcte:
ScalaCheck provides these out of the box in a more complete form. We have to wait until the next release of ScalaCheck, due to needing the instances for Scalajs in testing.

See discussion in: https://github.com/circe/circe/pull/1974

PR #1995, opened 2022-07-11
Use ZoneOffset Rather Than ZoneId In Most Arbitraries

isomarcte:
This works around special case rules for ZoneId values which cause issues with testing. We still test ZoneId itself, but we use ZoneOffset to construct the various time values.

This commit also change the `Arbitrary` `ZoneOffset` to be based of the ScalaCheck implementation which fully covers the domain. We can remove this once https://github.com/circe/circe/issues/1994 is complete.


armanbilge (approved):

modules/tests/shared/src/test/scala/io/circe/JavaTimeCodecSuite.scala
```diff
@@ -25,6 +26,28 @@ class JavaTimeCodecSuite extends CirceMunitSuite {
...
+    new Choose[ZoneOffset] {
+      override def choose(min: ZoneOffset, max: ZoneOffset): Gen[ZoneOffset] =
+        min.compareTo(max) match {
+          case 0                    => Gen.const(min)
+          case result if result > 0 => Gen.fail
+          case _                    =>
+            // Looks flipped, but it is not.
+            Gen.choose(max.getTotalSeconds, min.getTotalSeconds).map(value => ZoneOffset.ofTotalSeconds(value))
+        }
+    }
```
zarthross, line 41: I'm fine with bumping scalacheck to 1.16.0 if we need to so we don't have to do this backport.

zarthross (changes requested):

modules/tests/shared/src/test/scala/io/circe/JavaTimeCodecSuite.scala
armanbilge: We need ScalaCheck 1.17.0 actually.

isomarcte:
@zarthross the failure in CI is fixed in #1992. I'll merge with main.

modules/tests/shared/src/test/scala/io/circe/JavaTimeCodecSuite.scala
zarthross: Ok, we can keep this backported until 1.17.x then.

PR #1996, opened 2022-07-11
Cleanup PointerInterpolatorSuite

isomarcte:


isomarcte:
The big fix here is not deconstructing the result of the parse as `val Right(expected)` because when that is wrong for whatever reason we mask the real failure with a match error.

isomarcte:
Test failure is unrelated and fixed in #1997

zmccoy (approved):

isomarcte:
@zarthross anything I need to do here?

zarthross (approved):

PR #1998, opened 2022-07-12
Introduce forTypedProduct to work better with Scala 3

hamnis:
When using Scala 3 we see that tuples does not always implement ProductN.
Meaning that we cannot use generic tuples, but only "named".

For instance while using the code

```Scala
case class Person(name: String, age: Int)

object Person:
 given Encoder.AsObject[Person] = Encoder.forProduct2("name",
"age")(Tuple.fromProductTyped[Person])
```

is not possible, since this type of tuple does not implement Product2.

The only way we can do this in Scala3 is by having an instance like
this:

```Scala
 given Encoder.AsObject[Person] = Encoder.forProduct2("name", "age")(x
=> (x.name, x.age))
```

which I think at least is less useful and more error prone than the
version above.

Related dotty bug. https://github.com/lampepfl/dotty/issues/15253

EDIT:

Added support for call-by-name implicits for Scala 2.13 and 3. to allow us to use recursive encoders.

### DIFF
--- a/project/Boilerplate.scala
+++ b/project/Boilerplate.scala
@@ -272,6 +272,10 @@ object Boilerplate {
         }
       val outputType = if (arity != 1) s"Product$arity[${`A..N`}]" else `A..N`
 
+      val deprecate = if (arity != 1)
+        s"""@deprecated(message="does not work correctly for Scala 3. Prefer forTupleProduct$arity", since="0.14.6")"""
+      else ""
+
       block"""
         |package io.circe
         |
@@ -279,6 +283,7 @@ object Boilerplate {
         -  /**
         -   * @group Product
         -   */
+        -  $deprecate
         -  final def forProduct$arity[Source, ${`A..N`}]($memberNames)(f: Source => $outputType)(implicit
         -    $instances
         -  ): Encoder.AsObject[Source] =
@@ -317,8 +322,8 @@ object Boilerplate {
         -  /**
         -   * @group Product
         -   */
-        -  final def forProduct$arity[Source, ${`A..N`}]($memberNames)(f: Source => $outputType)(implicit
-        -    $instances, dummy: DummyImplicit
+        -  final def forTupleProduct$arity[Source, ${`A..N`}]($memberNames)(f: Source => $outputType)(implicit
+        -    $instances
         -  ): Encoder.AsObject[Source] =
         -    new Encoder.AsObject[Source] {
         -      final def encodeObject(a: Source): JsonObject = {
@@ -365,6 +370,10 @@ object Boilerplate {
         }
       val outputType = if (arity != 1) s"Product$arity[${`A..N`}]" else `A..N`
 
+      val deprecate = if (arity != 1)
+        s"""@deprecated(message="does not work correctly for Scala 3. Prefer forTupleProduct$arity", since="0.14.6")"""
+      else ""
+
       block"""
         |package io.circe
         |
@@ -372,6 +381,7 @@ object Boilerplate {
         -  /**
         -   * @group Product
         -   */
+        -  $deprecate
         -  final def forProduct$arity[A, ${`A..N`}]($memberNames)(f: (${`A..N`}) => A)(g: A => $outputType)(implicit
         -    $decoderInstances,
         -    $encoderInstances
@@ -430,10 +440,9 @@ object Boilerplate {
         -  /**
         -   * @group Product
         -   */
-        -  final def forProduct$arity[A, ${`A..N`}]($memberNames)(f: (${`A..N`}) => A)(g: A => $outputType)(implicit
+        -  final def forTupleProduct$arity[A, ${`A..N`}]($memberNames)(f: (${`A..N`}) => A)(g: A => $outputType)(implicit
         -    $decoderInstances,
-        -    $encoderInstances,
-        -    dummy: DummyImplicit
+        -    $encoderInstances
         -  ): Codec.AsObject[A] =
         -    new Codec.AsObject[A] {
         -      final def apply(c: HCursor): Decoder.Result[A] = $result
@@ -466,6 +475,8 @@ object Boilerplate {
       val memberVariableNames = (0 until arity).map(i => s"s$i").mkString(", ")
       val memberArbitraryItems = (0 until arity).map(i => s"s$i <- Arbitrary.arbitrary[String]").mkString("; ")
 
+      val forProduct = if (arity != 1) "forTupleProduct" else "forProduct"
+
       block"""
         |package io.circe
         |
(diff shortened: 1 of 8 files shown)


zarthross:
> is not possible, since this type of tuple does not implement Product2.

That is unfortunate...

> which I think at least is less useful and more error prone than the version above.

Both methods are exactly as error prone, because you have to ensure the order of the strings matches the order of the tuple you are making.  
 
> Related dotty bug. [lampepfl/dotty#15253](https://github.com/lampepfl/dotty/issues/15253)

Not really a bug, but appears to be buy design, specifically phasing out `TupleN` in favor of the more generic tuples.


That all being said, I recommend using circe-generic to get around this, especially if you are concerned about user error since it ensures the names and tuple parts of are always aligned.  

I'll take a closer look at this when i get a chance, but I'm concerned, that despite MIMA passing, that this would be a binary breaking change.

armanbilge:
> I'm concerned, that despite MIMA passing, that this would be a binary breaking change.

I can look into this. Either it's binary-compatible, or MiMa has a bug, or we have a mis-configuration that should be fixed.

hamnis:
I dont want to have DTOs, meaning derving this automatically.

I care too much about names of json properties for not having those reflected in the scala case classes directly. This has been the best way of deciding these in a relative declarative fashion.

Generating this code and insourcing this part of Circe is a workable solution if this will not be merged.

If you for instance consume JSON-LD then you really dont want to use that as scala case class fields.

isomarcte:
I think we want an API to be compatible with operating on tuples, that is now most people use `forProductN` in my experience. 

The only way forward that I see here is to add a new non-overloaded tuple based variant, which will be redundant for Scala 2, but fix the issue in Scala 3.

@hamnis @armanbilge @zarthross thoughts?

isomarcte:
I've an alternative idea too. Let me POC it real quick.

hamnis:
We could also generate sources for scala2/scala3 accordingly, by retaining the old constraints in scala2 and having a pure tuple version in scala3.

hamnis:
@isomarcte did you find an alternative solution?

hamnis:
@zarthross any way forward here? are there any users that uses ProductN instead of TupleN?

isomarcte:
@hamnis I'm sorry, I missed your ping. I'll take a second look today.

isomarcte:
@zarthross @hamnis [I've a poc for a non-source breaking fix here.](https://github.com/isomarcte/circe/commit/a711a0eb788e4a6219a8152e0f0d4ca7e71654f4) However, after working through the problem a little bit, I noticed that we are only using `ProductN` for the generated `Encoder` path, not the `Decoder` or `Codec`. This means that while the current code permits other ProductN subtypes besides tuples and case classes for encoding, we don't have code to support that for decoding. 

I'm of the opinion that we should keep the `forProductN` methods in the `Encoder`, `Decoder`, and `Codec` paths consistent with each other, and to that end I think we should just merge @hamnis's PR here. 

For 0.15.x we might consider renaming these to `forTupleN` and if we _really want to_ we can add `forProductN` types which actually deal with `ProductN` in all code paths (not sure if that is really of any value or not).

Thoughts?

hamnis:
According to https://github.com/lampepfl/dotty/issues/15253#issuecomment-1135848470 this will not really improve the situation since the long term goal is to get rid of the arity based types. It will work for the short term though.
using `(A, B)` will also work in a long term solution.

isomarcte:
@hamnis I think it will actually, unless I'm misreading that thread. When you use the `("a", 1)` syntax, you are getting an instance of `String *: Int :* EmptyTuple`, which I believe is the tuple type they intend to keep for the long term.

```scala
scala> summon[(Int, Int) =:= Int *: Int *: EmptyTuple]
val res0: (Int, Int) =:= (Int, Int) = generalized constraint

scala> final case class Person(name: String, age: Int)
// defined case class Person

scala> Tuple.fromProductTyped[Person](Person("a", 1)).isInstanceOf[String *: Int *: EmptyTuple]
val res1: Boolean = true
```

hamnis:
Yes, my comment was clumsily worded. I believe that this branch is the most "future" proof of the current solutions.

Having said that we can generate for both scala2 and scala3 here, where the scala3 source has the tuple constraint and scala2 remains unmodified.

isomarcte:
@zarthross thoughts on merging this?

zarthross:
Before considering merging it, I'd like to see some unit tests demonstrating what the change is doing, mostly so in the future we don't accidentally remove the support.

hamnis:
@zarthross Added tests. Please advice if they should live somewhere else.

zAPFy:
@zarthross, I would love to see this merged. Any way I can assist to get some progress?

zmccoy:
Hi @hamnis @zarthross and I were going through this and had a few things we wanted to bring up.  We're a bit weary of changing the current behavior as it does in this PR as this will break cases like https://github.com/akka/alpakka-kafka/blob/1e25e79f33e85703cfc3edd39eceaa3a3b5f471b/core/src/main/scala/akka/kafka/ConsumerMessage.scala#L94 in scala 2 if it were to utilize forProductN because of the subtyping of Product2 not being a Tuple.  Because of this existence of such things in the wild, it may be safer to have a new boilerplate method that creates code like
```
  final def forProduct2[Source, A0, A1](nameA0: String, nameA1: String)(f: Source => (A0, A1))(implicit
    encodeA0: Encoder[A0], encodeA1: Encoder[A1], dummy: DummyImplicit
  ): Encoder.AsObject[Source] =
  ```
This uses a dummy implicit so that users won't see any change and it will still utilize the correct method according to types after type erasure.  With this change then, it would be good to include as many test cases, or ways to create instances of Case classes, products, tuples, etc to make sure this doesn't break any of them, as well as helps us not break this in the future.

**Comment by hamnis** [comment]
@zarthross @zmccoy sorry for the delay. I have updated with overloads as requested.

**Comment by hamnis** [comment]
I think we need a new name for this to avoid compile errors of existing code. `forProductTupleN` ?

File: project/Boilerplate.scala [hunk 1]
```diff
@@ -317,8 +322,8 @@ object Boilerplate {
         -  /**
         -   * @group Product
         -   */
-        -  final def forProduct$arity[Source, ${`A..N`}]($memberNames)(f: Source => $outputType)(implicit
-        -    $instances, dummy: DummyImplicit
+        -  final def forTupleProduct$arity[Source, ${`A..N`}]($memberNames)(f: Source => $outputType)(implicit
```
hamnis: Bikeshedding of new name is needed.
forNamedProduct, ofProduct are some suggestions.

zarthross: I agree, a new name is needed.   Could we just use `fromProductTyped$arity`?   Also, I feel like we should only generate these for Scala3 to avoid any confusion or issues in Scala2.

hamnis: if we are only generating this for scala 3, then we could keep the original name and just change the type to tuple.

zarthross: I would rather keep the new name, I think my previous concerns with the change of the signature still apply even if this change is only for scala 3.

project/Boilerplate.scala
```diff
@@ -270,13 +272,18 @@ object Boilerplate {
         }
       val outputType = if (arity != 1) s"Product$arity[${`A..N`}]" else `A..N`
 
+      val deprecate = if (arity != 1)
+        s"""@deprecated(message="does not work correctly for Scala 3. Prefer forTupleProduct$arity", since="0.14.6")"""
+      else ""
+
```
zarthross: I don't want to deprecate this.  It works just fine for many existing workflows, just not the `Tuple.fromProductTyped`

hamnis:
@zarthross @zmccoy please re-review. sorry for the delays.

isomarcte:
@zarthross @zmccoy The pointer tests failed. I have a vague memory of them just being generally flaky. Is that right?

hamnis:
@zmccoy @zarthross is there anything that is missing from this PR now?

zarthross (approved):

zmccoy (approved):

PR #2004, opened 2022-07-19
Cursor Redesign

isomarcte:
This commit provides a redesign of `Cursor`. The reasons for this redesign

* `ACursor` is not sealed. This created some modeling and correctness issues around certain operations. For example, when attempting to create an efficient representation of a particular `ACursor` to the root of a JSON object (for errors and the like), we had to handle the case that the given `ACursor` we were working on was not a known type.
* Certain useful fields such as `lastCursor` and `lastOp` could not be exposed safely as they might be `null`.
* Certain cursor operations were not modeled in the `CursorOp`, e.g. `withFocus`, creating a difficult situation where `null` in the cursor history both indicated the beginning of the history, as well as an unknown operation.
* `HCursor` exposed and relied upon operations like `addOp` and `replace` which permitted corruption of the cursor history.
* Certain other methods had odd semantics (see `root`).

`Cursor` is primarily just moving all of the `ACursor`, `HCursor`, `ObjectCursor`, `ArrayCursor`, `TopCursor`, and `FailedCursor`, into a single file so we can seal it.

It provides better error information in `FailedCursor` (which will be addiative to decoding errors, not replace them). Safely exposes certain desirable methods, e.g. `lastCursor` and `lastOp`.

We also add a new `CursorOp.Replace` so we can fully record history on the cursor.

`Cursor` values may be converted to `ACursor` or `HCursor` values and the majority of the implementation of circe Cursors in this commit now use `Cursor` under the hood.


isomarcte:
This commit still needs some changes, hence the draft state. 

Things need to be deprecated, tests need to be added. There are also some `TODO` values added that need to be addressed in this or followup PRs.

@zarthross @zmccoy, assuming you guys are on board with this change (targeting 0.15.x), how much do you want in this PR vs follow up PRs. It's the tension of a complete implementation vs. smaller PRs.

isomarcte:
Closed in favor of #2015

PR #2005, opened 2022-07-19
Bump base version to 0.15 on 0.15.x branch

armanbilge:
https://typelevel.org/sbt-typelevel/faq.html#what-is-a-base-version-anyway

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -188,6 +188,7 @@ lazy val docs = project
     moduleName := "circe-docs",
     name := "Circe docs",
     mdocIn := file("docs/src/main/tut"),
+    evictionErrorLevel := Level.Warn,
     libraryDependencies ++= Seq(
       "io.circe" %% "circe-generic-extras" % "0.14.1",
       "io.circe" %% "circe-optics" % "0.14.1"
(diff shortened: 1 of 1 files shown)


zarthross (dismissed):
Thanks @armanbilge.

armanbilge:
Seems to me the circe-optics docs should live in the circe-optics repository.

zarthross:
Yeah.. the benchmarks also depend on circe-optics.    I'm not opposed to moving the circe-optics docs into circe-optics project, but I would like to somehow keep them on main site as well.  Maybe we can have circe-optics push its docs into the circe/circe docs branch as it gets updated?  Or maybe just link to a circe-optics doc site is sufficient...

armanbilge:
Circe org is big enough you could benefit from a dedicated sbt plugin. Especially if you decide to schism this repo as well (probably a good idea in the long term?)

https://github.com/http4s/sbt-http4s-org/

build.sbt
```diff
@@ -188,6 +188,7 @@ lazy val docs = project
     moduleName := "circe-docs",
     name := "Circe docs",
     mdocIn := file("docs/src/main/tut"),
+    evictionErrorLevel := Level.Warn,
```
armanbilge, line 191: Dealing with circe-optics is ok but untangling circe-generic-extras from the docs is harder. This will get us by for now.

zarthross (approved):

Issue #1980, opened 2022-06-28
Automatic derivation for recursive types is broken in Scala 3


luksow:
Hi,

```scala
#!/usr/bin/env scala-cli
using scala "3.1.2"

using lib "io.circe:circe-core_3:0.14.1"
using lib "io.circe:circe-parser_3:0.14.1"
using lib "io.circe:circe-generic_3:0.14.1"

import io.circe._
import io.circe.generic.auto._

case class R1(int: Int, r: Seq[R1])
case class R2(int: Int, r: Option[R2])

summon[Encoder[R1]]
summon[Encoder[R2]]
```

Both `summon`s fail but with slightly different messages:
```
[error] ./circe.sc:14:20: cannot reduce summonFrom with
[error]  patterns :  case given encodeA @ _:io.circe.Encoder[Seq[circe.R1]]
[error]              case given evidence$1 @ _:deriving.Mirror.Of[Seq[circe.R1]]
[error] summon[Encoder[R1]]
[error]
```
and
```
[error] ./circe.sc:15:20: method derived is declared as `inline`, but was not inlined
[error] 
[error] Try increasing `-Xmax-inlines` above 32
[error] summon[Encoder[R2]]
[error]                    ^
```

Probably related to https://github.com/lampepfl/dotty/issues/8183

armanbilge:
Thanks for the report. Note that the following works.
```scala
//> using scala "3.1.3"

//> using lib "io.circe::circe-core::0.14.2"
//> using lib "io.circe::circe-parser::0.14.2"

import io.circe._

case class R1(int: Int, r: Seq[R1]) derives Encoder.AsObject
case class R2(int: Int, r: Option[R2]) derives Encoder.AsObject

@main def main =
  summon[Encoder[R1]]
  summon[Encoder[R2]]
```

luksow:
@armanbilge Yes, thanks for mentoning this workaround and sorry I forgot to :smiley: 

I'm also looking for a different workaround as I'm not a big fan of mixing my precious data classes with 3rd party libraries - for the sake of domain purity.

armanbilge:
What is the "3rd party library" in this situation?

armanbilge:
Is this what you want?
```scala
//> using scala "3.1.3"

//> using lib "io.circe::circe-core::0.14.2"
//> using lib "io.circe::circe-parser::0.14.2"

import io.circe._

case class R1(int: Int, r: Seq[R1])
case class R2(int: Int, r: Option[R2])

[...]

PR #2009, opened 2022-07-31
Fix nested wrapped element in scala 3 macro derivation

### LINKED ISSUE #1695: Error message when deriving for List[A] in Scala 3 needs improvement
Currently, when you try to either derive `Encoder[List[A]]` directly (by `deriveEncoder[List[A]]`) or indirectly (by deriving for a type that contains a `List[A]`) with semiauto and there doesn't exist an `Encoder[A]`, you get an error message that really isn't all that helpful:
```
Maximal number of successive inlines (32) exceeded,
Maybe this is caused by a recursive inline method?
You can use -Xmax-inlines to change the limit.
```

The problem here is that the Scala 3 derivation code tries to auto-derive `Encoder[List[A]]` (which is insane) when it cannot use `Encoder.encodeList[A]` because of missing `Encoder[A]`.

Here is a scastie example showing it off:
https://scastie.scala-lang.org/heksesang/z1VPL9zST2S9tmRwMz5oJg/1

### LINKED ISSUE #2001: Decoding fails when only providing a decoder for the top level class in Scala 3
In the following snippet I have only derived a decoder for the parent class B. When trying to use the decoder, it gives a DecodingFailure on runtime, unless I also provide a decoder for the child class B. This only seems to be a problem when the child class is optional. I would expect either a compilation error like you would get in Scala 2, or a working decoder for the whole structure.

Any ideas why this happens?

```scala
import io.circe.Decoder
import io.circe.syntax._
import io.circe.parser.decode

case class A(i: Int)
case class B(a: Option[A])

// given Decoder[A] = deriveDecoder       // With this line, decoding succeeds
given Decoder[B] = deriveDecoder

println(decode[B]("""{"a": {"i": 1}}""")) // Left(DecodingFailure(Option, List(DownField(a))))
println(decode[B]("""{"a": null}"""))     // Left(DecodingFailure(Option, List(DownField(a))))
println(decode[B]("""{}"""))              // Left(DecodingFailure(Missing required field, List(DownField(a))))
```

Try it out here: https://scastie.scala-lang.org/hZ2gjH6IQnOV4ZMAQN7bUg

Also, thank you for this great library!

amumurst:
This works stared as an attempt to fix #2001 but also should now fix #1695 

Derivation of systems like the one described below for standard F's like `List`, `Option` and `Vector` is currently not working unless the inner value (Inner here) ALSO is derived. This PR attempts to remedy this.

# System
## Not working
If we derive ONLY the wrapper type, it will derive wrongly (see below for issues for concrete wrappers). Here implicit vals are used, but using `given` or `derives` also has these issues because they all just forward to the same `derived` method. This is also what `generic.semiAuto.{deriveEncoder, deriveDecoder}` forwards too. This PR also does not include Codec instances as they also just forwards to the Decoder/Encoder implementations addressed here. 
```scala
type Wrapper[A] = Option[A]
case class Inner(a: Long)
case class Outer(inner: Wrapper[Inner])
implicit val encoder: Encoder[Outer] = Encoder.AsObject.derived
implicit val decoder: Decoder[Outer] = Decoder.derived
```

## Working
Given the same system as above with the added encoder/decoder of inner. They will be correct.
```scala
type Wrapper[A] = Option[A]
case class Inner(a: Long)
case class Outer(inner: Wrapper[Inner])
implicit val innerEncoder: Encoder[Inner] = Encoder.AsObject.derived
implicit val innerDecoder: Decoder[Inner] = Decoder.derived
implicit val encoder: Encoder[Outer] = Encoder.AsObject.derived
implicit val decoder: Decoder[Outer] = Decoder.derived
```

# Wrappers
There are 2 main ways wrappers currently break

## Option
Option breaks by encoding Some and None as values. *NOTE* that this means that `Codec` laws actually currently work since they all test round-tripping (which is upheld by encoding `Some` and `None` as objects) 👍 .
Console reproducer:
```scala
import io.circe.{Encoder, Decoder, Json}
case class Inner(a: Long)
case class Outer(inner: Option[Inner])
implicit val encoder: Encoder[Outer] = Encoder.AsObject.derived
implicit val decoder: Decoder[Outer] = Decoder.derived

encoder(Outer(Some(Inner(5))))
decoder(Json.obj("inner" -> Json.obj("a" -> Json.fromInt(5))).hcursor)
encoder(Outer(None))
decoder(Json.obj("inner" -> Json.Null).hcursor)
```
Gives console output
```scala
// defined case class Inner
// defined case class Outer
val encoder: io.circe.Encoder[Outer] = anon$1@27853997
val decoder: io.circe.Decoder[Outer] = anon$2@73e6b44c
val res0: io.circe.Json = {
  "inner" : {
    "Some" : {
      "value" : {
        "a" : 5
      }
    }
  }
}
val res1: io.circe.Decoder.Result[Outer] = Left(DecodingFailure(Option, List(DownField(inner))))
val res2: io.circe.Json = {
  "inner" : {
    "None$" : null
  }
}
val res3: io.circe.Decoder.Result[Outer] = Left(DecodingFailure(Option, List(DownField(inner))))
```

## List/Seq-like types
Changing Wrapper type from Option to something sequence like (List, Vector, Map, Chain, etcetc) fails to compile.
```scala
import io.circe.{Encoder, Decoder, Json}
case class Inner(a: Long)
case class Outer(inner: List[Inner])
implicit val encoder: Encoder[Outer] = Encoder.AsObject.derived
implicit val decoder: Decoder[Outer] = Decoder.derived

encoder(Outer(List(Inner(5))))
decoder(Json.obj("inner" -> Json.arr(Json.obj("a" -> Json.fromInt(5)))).hcursor)
encoder(Outer(List.empty))
decoder(Json.obj("inner" -> Json.arr).hcursor)
```

Console output
```scala
-- Error:
4 |implicit val encoder: Encoder[Outer] = Encoder.AsObject.derived
  |                                       ^^^^^^^^^^^^^^^^^^^^^^^^
  |                       Maximal number of successive inlines (32) exceeded,
  |                       Maybe this is caused by a recursive inline method?
  |                       You can use -Xmax-inlines to change the limit.
  | This location contains code that was inlined from rs$line$1:4
  | This location contains code that was inlined from Derivation.scala:85
  | This location contains code that was inlined from Derivation.scala:19
  | This location contains code that was inlined from Derivation.scala:76
  | This location contains code that was inlined from Derivation.scala:15
  | This location contains code that was inlined from Derivation.scala:88
  | This location contains code that was inlined from Derivation.scala:19
  | This location contains code that was inlined from Derivation.scala:76
  | This location contains code that was inlined from Derivation.scala:15
  | This location contains code that was inlined from Derivation.scala:88
  | This location contains code that was inlined from Derivation.scala:19
  | This location contains code that was inlined from Derivation.scala:76
  | This location contains code that was inlined from Derivation.scala:76
  | This location contains code that was inlined from Derivation.scala:15
  | This location contains code that was inlined from Derivation.scala:88
  | This location contains code that was inlined from Derivation.scala:19
  | This location contains code that was inlined from Derivation.scala:76
  | This location contains code that was inlined from Derivation.scala:15
  | This location contains code that was inlined from Derivation.scala:88
  | This location contains code that was inlined from Derivation.scala:19
  | This location contains code that was inlined from Derivation.scala:76
  | This location contains code that was inlined from Derivation.scala:76
  | This location contains code that was inlined from Derivation.scala:15
  | This location contains code that was inlined from Derivation.scala:88
  | This location contains code that was inlined from Derivation.scala:19
  | This location contains code that was inlined from Derivation.scala:76
  | This location contains code that was inlined from Derivation.scala:15
  | This location contains code that was inlined from Derivation.scala:88
  | This location contains code that was inlined from Derivation.scala:19
  | This location contains code that was inlined from Derivation.scala:76
  | This location contains code that was inlined from Derivation.scala:15
  | This location contains code that was inlined from Derivation.scala:88
-- Error:
5 |implicit val decoder: Decoder[Outer] = Decoder.derived
  |                                       ^^^^^^^^^^^^^^^
  |                       Maximal number of successive inlines (32) exceeded,
  |                       Maybe this is caused by a recursive inline method?
  |                       You can use -Xmax-inlines to change the limit.
  | This location contains code that was inlined from rs$line$1:5
  | This location contains code that was inlined from Derivation.scala:106
  | This location contains code that was inlined from Derivation.scala:24
  | This location contains code that was inlined from Derivation.scala:55
  | This location contains code that was inlined from Derivation.scala:14
  | This location contains code that was inlined from Derivation.scala:109
  | This location contains code that was inlined from Derivation.scala:24
  | This location contains code that was inlined from Derivation.scala:55
  | This location contains code that was inlined from Derivation.scala:14
  | This location contains code that was inlined from Derivation.scala:109
  | This location contains code that was inlined from Derivation.scala:24
  | This location contains code that was inlined from Derivation.scala:55
  | This location contains code that was inlined from Derivation.scala:55
  | This location contains code that was inlined from Derivation.scala:14
  | This location contains code that was inlined from Derivation.scala:109
  | This location contains code that was inlined from Derivation.scala:24
  | This location contains code that was inlined from Derivation.scala:55
  | This location contains code that was inlined from Derivation.scala:14
  | This location contains code that was inlined from Derivation.scala:109
  | This location contains code that was inlined from Derivation.scala:24
  | This location contains code that was inlined from Derivation.scala:55
  | This location contains code that was inlined from Derivation.scala:55
  | This location contains code that was inlined from Derivation.scala:14
  | This location contains code that was inlined from Derivation.scala:109
  | This location contains code that was inlined from Derivation.scala:24
  | This location contains code that was inlined from Derivation.scala:55
  | This location contains code that was inlined from Derivation.scala:14
  | This location contains code that was inlined from Derivation.scala:109
  | This location contains code that was inlined from Derivation.scala:24
  | This location contains code that was inlined from Derivation.scala:55
  | This location contains code that was inlined from Derivation.scala:14
  | This location contains code that was inlined from Derivation.scala:109
```

# Reason
Both of these issues are the same, just manifest themselves differently.  
Call stack of issue in object `Derivation` is
`derived -> summon{Decoders, Encoders} -> summon{Decoders, Encoders}Rec -> summon{Decoder, Encoder}`

SummonEncoder is implemented as:
```scala
  inline final def summonEncoder[A]: Encoder[A] = summonFrom {
    case _: Mirror.Of[A]     => Encoder.AsObject.derived[A]
  }
  ```
For Option this means that A=Option[Inner]. It then checks if there exists an `implicit Encoder[Option[Inner]]` in scope. In the companion object of `Encoder` there is an implicit def that can produce that for all types Option[A] where A has a Encoder. 
```scala 
implicit final def encodeOption[A](implicit e: Encoder[A]): Encoder[Option[A]] = ...
```

But here we do not have that! Since we chose/forgot to derive an `Encoder` for `Inner`, there is no `Encoder[A]` available. And such there is also no `Encoder[Option[A]]` available. Well, then the code goes to the next thing and attempts to make that. It can then introspect `Option` and see that.. Well this is a Sum type with 2 known members, `Some` and `None`. Okay then it can just derive those based on the classnames!

For `List` the same thing happens. But `List` is not as easy for the compiler to produce encoder/decoder for because.. Well at top level it is fine as a Sum type with `Nil` and `::`, but `::` contains another `List`, hence the recursive inline method issue.

# Solution
The solution used here is to match on the A in `summon{Decoders, Encoders}Rec` with the wrapper type and extract the wrapped type out. Then we forward that to `summon{Decoder, Encoder}` to get the inner value, then pass that on to the respective wrapper `encoder/decoder` implementation in the companion objects of `Encoder/Decoder`

The implemented tests are added in a scala-3 folder in tests module. They are able to compile for scala-2 also, but they would then need to depend on the generic modules for semiAuto derivation. Since that module is only a forwarder to core for scala 3, I felt they belonged more in core (even though that means they can only run on 3). If someone feels otherwise I can move them to generic and replace all derived with `semiAuto.derive{Encoder, Decoder}` to get them running on both 2 and 3.

# Fixed Wrappers
I have based this list on the wrapper that has instances custom made in the Encoder/Decoder companion objects.

- [x] Option
- [x] Map
- [x] Seq
- [x] Set
- [x] List
- [x] Vector
- [x] Chain
- [x] NonEmptyList
- [x] NonEmptyVector
- [ ] NonEmptySet
- [ ] NonEmptyMap
- [ ] NonEmptyChain

### DIFF
--- a/modules/core/shared/src/main/scala-3/io/circe/Derivation.scala
+++ b/modules/core/shared/src/main/scala-3/io/circe/Derivation.scala
@@ -1,6 +1,6 @@
 package io.circe
 
-import cats.data.{ Chain, NonEmptyList, NonEmptyMap, NonEmptySet, NonEmptyVector, Validated }
+import cats.data.{ Chain, NonEmptyChain, NonEmptyList, NonEmptyMap, NonEmptySet, NonEmptyVector, Validated }
 import cats.kernel.Order
 
 import scala.collection.immutable.SortedSet
@@ -52,6 +52,8 @@ object Derivation {
       case _: (Map[k, t] *: ts) => Decoder.decodeMap(summonKeyDecoder[k], summonDecoder[t]) :: summonDecodersRec[ts]
       case _: (NonEmptyMap[k, t] *: ts) => // NonEmptyMap is a NewType, matching does not seem to work
         Decoder.decodeNonEmptyMap(summonKeyDecoder[k], summonOrder[k], summonDecoder[t]) :: summonDecodersRec[ts]
+      case _: (NonEmptyChain[t] *: ts) => // NonEmptyChain is a NewType, matching does not seem to work
+        Decoder.decodeNonEmptyChain(summonDecoder[t]) :: summonDecodersRec[ts]
       case _: (t *: ts) => summonDecoder[t] :: summonDecodersRec[ts]
     }
 
@@ -73,6 +75,8 @@ object Derivation {
       case _: (Map[k, t] *: ts) => Encoder.encodeMap(summonKeyEncoder[k], summonEncoder[t]) :: summonEncodersRec[ts]
       case _: (NonEmptyMap[k, t] *: ts) => // NonEmptyMap is a NewType, matching does not seem to work
         Encoder.encodeNonEmptyMap(summonKeyEncoder[k], summonEncoder[t]) :: summonEncodersRec[ts]
+      case _: (NonEmptyChain[t] *: ts) => // NonEmptyChain is a NewType, matching does not seem to work
+        Encoder.encodeNonEmptyChain(summonEncoder[t]) :: summonEncodersRec[ts]
       case _: (t *: ts) => summonEncoder[t] :: summonEncodersRec[ts]
     }
 }
--- a/modules/core/shared/src/main/scala-3/io/circe/Derivation.scala
+++ b/modules/core/shared/src/main/scala-3/io/circe/Derivation.scala
@@ -1,6 +1,9 @@
 package io.circe
 
-import cats.data.{ NonEmptyList, Validated }
+import cats.data.{ Chain, NonEmptyList, NonEmptyMap, NonEmptySet, NonEmptyVector, Validated }
+import cats.kernel.Order
+
+import scala.collection.immutable.SortedSet
 import scala.deriving.Mirror
 import scala.collection.mutable.WrappedArray
 import scala.compiletime.{ constValue, erasedValue, error, summonFrom }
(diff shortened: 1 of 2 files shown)


amumurst:
`NonEmpty{Set, Map, Chain}` don't work yet. They just don't match in `summon{Decoders, Encoders}Rec` with the current implementation, and fail. They are all NewTypes in cats so it feels like something weird is going on here. All help valued!

amumurst:
It also changes the error message in #1980 to 
```scala
scala> import io.circe._
     | import io.circe.generic.auto._
     | 
     | case class R1(int: Int, r: Seq[R1])
     | case class R2(int: Int, r: Option[R2])
     | 
     | summon[Encoder[R1]]
     | summon[Encoder[R2]]
-- Error:
7 |summon[Encoder[R1]]
  |                   ^
  |                 method derived is declared as erased, but is in fact used
  | This location contains code that was inlined from Derivation.scala:19
  | This location contains code that was inlined from Derivation.scala:66
  | This location contains code that was inlined from Derivation.scala:80
  | This location contains code that was inlined from Derivation.scala:15
  | This location contains code that was inlined from Derivation.scala:92
  | This location contains code that was inlined from auto.scala:18
-- Error:
7 |summon[Encoder[R1]]
  |                   ^
  |       method summonEncodersRec is declared as erased, but is in fact used
  | This location contains code that was inlined from Derivation.scala:66
  | This location contains code that was inlined from Derivation.scala:80
  | This location contains code that was inlined from Derivation.scala:15
  | This location contains code that was inlined from Derivation.scala:92
  | This location contains code that was inlined from auto.scala:18
-- Error:
8 |summon[Encoder[R2]]
  |                   ^
  |       method summonEncodersRec is declared as erased, but is in fact used
  | This location contains code that was inlined from Derivation.scala:63
  | This location contains code that was inlined from Derivation.scala:80
  | This location contains code that was inlined from Derivation.scala:15
  | This location contains code that was inlined from Derivation.scala:92
  | This location contains code that was inlined from auto.scala:18
```

I dunno if that is better or worse :-)

hamnis:
I think this has been fixed in 0.14.7, closing for now.

PR #2010, opened 2022-08-05
Make test failures better

isomarcte:
This PR changes tests which were using `assert(actual == expected)` or `assert(actual === expected)` into either

* `assertResult(actual, expected)`, for non-property tests
* `actual ?= expected`, for property tests

Tests of the form `test("testName"){forAll(...)}` were changed to `property("testName"){forAll(...)}`.

Tests of the form `property("prop"){forAll(a => assertEquals(actual, expected))}` were changed to `property("prop"){forAll(a => actual ?= expected)}`, e.g. they now use the property based check rather than the MUnit assertion based one.

The vast majority of the changes were done with an internal Scalafix rule I added. This rule was imprecise in some cases and thus there was some minor manual fixup to do as well.

I left the internal rule in place (though not active) since we can re-use the scaffolding for other internal rules in the future.


zarthross (approved):
Looks good too me.  Can you merge in latest `series/0.14.x` so we can merge this?

armanbilge:
@zarthross is this the only thing blocking an 0.14.3 release? I'm happy to fix the merge conflicts so we can get this going :)

zmccoy (approved):
Sorry this is so late, thank you for this!

PR #2015, opened 2022-08-24
Cursor refactor

isomarcte:
This commit provides a redesign of `Cursor`. The reasons for this redesign

* `ACursor` is not sealed. This created some modeling and correctness issues around certain operations. For example, when attempting to create an efficient representation of a particular `ACursor` to the root of a JSON object (for errors and the like), we had to handle the case that the given `ACursor` we were working on was not a known type.
* Certain useful fields such as `lastCursor` and `lastOp` could not be exposed safely as they might be `null`.
* Certain cursor operations were not modeled in the `CursorOp`, e.g. `withFocus`, creating a difficult situation where `null` in the cursor history both indicated the beginning of the history, as well as an unknown operation.
* `HCursor` exposed and relied upon operations like `addOp` and `replace` which permitted corruption of the cursor history.
* Certain other methods had odd semantics (see `root`).

`Cursor` is primarily just moving all of the `ACursor`, `HCursor`, `ObjectCursor`, `ArrayCursor`, `TopCursor`, and `FailedCursor`, into a single file so we can seal it.

It provides better error information in `FailedCursor` (which will be addiative to decoding errors, not replace them). Safely exposes certain desirable methods, e.g. `lastCursor` and `lastOp`.

We also add a new `CursorOp.Replace` so we can fully record history on the cursor.

`Cursor` values may be converted to `ACursor` or `HCursor` values and the majority of the implementation of circe Cursors in this commit now use `Cursor` under the hood.


isomarcte:
The `Decoder`, `DecodingFailure`, and related code would still need to be updated, but I'm leaving those for other PRs.

isomarcte:
@zarthross any thoughts here?

modules/core/shared/src/main/scala/io/circe/CursorOp.scala
```diff
@@ -1,47 +1,49 @@
...
    * object?
    */
-  def requiresObject: Boolean
+  final def requiresObject: Boolean =
+    this match {
+      case _: DownField =>
+        true
+      case _ =>
+        false
+    }
```
zarthross, line 20: I have to wonder if this is better than the original?  Is the method dispatch slower than the type check in the pattern match?  I know its easier to understand whats going on, but is the trade off worth it?

modules/core/shared/src/main/scala/io/circe/Cursor.scala
```diff
@@ -0,0 +1,1351 @@
...
+   *
+   * A failed cursor will always have an empty focus.
+   */
+  final def failed: Boolean = succeeded === false
+
+  /**
+   * Move from the current position to the root of the JSON structure.
+   *
+   * If the cursor is a [[Cursor#SuccessCursor]], this is the same as calling
+   * [[#up]] until [[#isRoot]] returns true. If the cursor is a has failed,
```
zarthross:
```suggestion
   * [[#up]] until [[#isRoot]] returns true. If the cursor has failed,
```

modules/core/shared/src/main/scala/io/circe/Cursor.scala
```diff
@@ -0,0 +1,1351 @@
...
+   */
+  sealed abstract class History extends Product with Serializable {
+
+    /** A representation of the [[History]] as a `List`. */
+    def asList: List[CursorOp]
+
+    override final def toString: String =
+      s"""History(${asList.mkString(", ")})"""
+  }
+
```
zarthross, line 881: Maybe add a comment about the typical direction a `History` stores the `CursorOps`, for instance, the `head` of the list is the first operation applied to the originating cursor and not the last (as it may have been previously)

modules/core/shared/src/main/scala/io/circe/Cursor.scala
```diff
@@ -0,0 +1,1351 @@
...
+   *       method is private.
+   */
+  private[circe] final def replayCursorOps[F[_]: Foldable](history: F[CursorOp]): Cursor =
+    if (failed) {
+      // No need to waste anyone's time.
+      this
+    } else {
+      val asList: List[CursorOp] =
+        history.toList
+      asList.foldLeft(this) {
```
zarthross, line 316:
```suggestion
      history
        .toList
        .foldLeft(this) {
```

modules/core/shared/src/main/scala/io/circe/Cursor.scala
```diff
@@ -0,0 +1,1351 @@
...
+   * Replay history against the cursor for any `Foldable` of [[CursorOp]].
+   *
+   * @note This method is error prone as it is not clear if the first element
+   *       is the first operation or the last operation. Historically the
+   *       first element was the last operation, because of how Circe
+   *       constructed the history, but it can make intuitive sense to have
+   *       the first element be the first operation. For this reason, this
+   *       method is private.
+   */
+  private[circe] final def replayCursorOps[F[_]: Foldable](history: F[CursorOp]): Cursor =
```
zarthross, line 309: I think its probably fine to make this public IF we add a comment explaining that the first element is assumed to be the start of the history and not the end.  I can imagine people reimplementing this themselves because they stored (for whatever reason) a History or list of cursor ops and wanted to apply it to another cursor.  Either open this, or maybe make it possible to construct a history from a list of cursor ops?  IDK, i don't have any strong feelings here... Thoughts?

modules/core/shared/src/main/scala/io/circe/Cursor.scala
```diff
@@ -0,0 +1,1351 @@
...
+    private[Cursor] def apply(value: Json, unsafeLastCursor: Cursor.SuccessCursor, unsafeLastOp: CursorOp): TopCursor =
+      TopCursorImpl(value, unsafeLastCursor, unsafeLastOp)
+  }
+
+  /**
+   * An internal trait for "normal" cursors. "normal" here means not the top
+   * and not a failure state. This is useful because "normal" cursors have
+   * some common properties which it is occassionaly nice to match on.
+   */
+  private sealed abstract class NormalCursor extends SuccessCursor {
```
zarthross, line 749: `Normal` i think is a little confusing?  What other options for names could we use?  `Intermediate`? I'm not sure, but `Normal` didn't initially feel right.

modules/tests/shared/src/test/scala/io/circe/CursorSuite.scala
```diff
@@ -0,0 +1,301 @@
...
+  val cursor: Cursor = Cursor.fromJson(j1)
+
+  property("focus should return the JSON value in a newly created cursor") {
+    forAll { (j: Json) =>
+      Cursor.fromJson(j).focus ?= Some(j)
+    }
+  }
+
+  property("Navigation into an object")(topProp)
+  private lazy val topProp = forAll { (j: Json) =>
```
zarthross, line 45: Why the `private lazy val` here?

modules/tests/shared/src/test/scala/io/circe/CursorSuite.scala
```diff
@@ -0,0 +1,301 @@
...
+    forAllP(intoObject)(value =>
+      ((value.top.fold(Prop.passed)(_ ?= j)) :| "Should return to top") &&
+        ((value.root.focus.fold(Prop.passed)(_ ?= j)) :| "Should return to root") &&
+        ((value.up.focus.fold(Prop.passed)(_ ?= j)) :| "Up should undo this operation")
+    )
+  }
+
+  property("Navigation into an array") {
+    forAll { (j: Json) =>
+      val c: Cursor = Cursor.fromJson(j).downArray
```
zarthross, line 62: Would these tests (navigate into object/array) be better if the `forAll` generated objects and arrays rather than generic json?

modules/core/shared/src/main/scala/io/circe/CursorOp.scala
isomarcte: TL;DR I'm totally fine moving this to the object.

That said, I'd be even more fine just removing it, except I'd like to minimize bincompat changes. It is only used by the old FailedCursor, which the work in this PR replaces. https://github.com/circe/circe/blob//modules/core/shared/src/main/scala/io/circe/FailedCursor.scala#L11

It is used to help describe the failure condition, which is now handled much more clearly by the CursorFailureReason error.

isomarcte: Since the use case in circe would be deprecated by this method and is not in performance critical code, do you still want me to move it?

isomarcte: FYI, my reason for moving it was just to try to make this file less verbose.

modules/core/shared/src/main/scala/io/circe/Cursor.scala
isomarcte: Yeah, I can do that. I struggled with what was the best way to both encode this and communicate this.

modules/core/shared/src/main/scala/io/circe/Cursor.scala
isomarcte: I can make it public and update the comment. I think having `History` as an always correct type, but permitting reply by an arbitrary `Foldable` makes sense.

modules/core/shared/src/main/scala/io/circe/Cursor.scala
isomarcte: Yeah, I totally agree, but I don't know what else to call it. `Intermediate` feels equally confusing to me.

modules/tests/shared/src/test/scala/io/circe/CursorSuite.scala
isomarcte: These are copied from the `ACursor` tests. The only changes I made were to make it compile with the new interface (which was always a simplification because `Cursor` is nicer to work with than `ACursor`).

I agree this is strange. I can update in this PR or a future one. I think the tests need a cleanup pass, I'm just not sure if you want that now or later.

modules/core/shared/src/main/scala/io/circe/Cursor.scala
isomarcte: I'd like to make the construction of `History` public, but I don't see any easy way to communicate at the types the opposite behavior of the list backing the type.

modules/core/shared/src/main/scala/io/circe/Cursor.scala
isomarcte: `HasParentCursor`?

isomarcte: `NotTopOrFailedCursor`? :laughing:

Issue #1972, opened 2022-06-10
v0.14.2 introduced significant performance degradation


lzm0:
We use Circe in several Scala cron jobs. Last week or so, we upgraded Circe from `v0.14.1` to `v0.14.2`, and some cron jobs took several times longer to complete.

We ran some profiling and found that https://github.com/circe/circe/pull/1617/ caused this performance reduction by introducing a new `cursorToDecodingFailure` function in `modules/core/shared/src/main/scala/io/circe/Decoder.scala` that takes linear time to run.

### `Demo.sc`
```Scala
import io.circe.Codec
import io.circe.generic.extras.Configuration
import io.circe.generic.extras.semiauto.deriveConfiguredCodec
import io.circe.parser.decode

import java.util.UUID
import scala.io.Source

case class Cat(
    id: UUID,
    friends: Seq[Cat] = Seq.empty
)

object Cat {
  implicit final val configuration: Configuration = Configuration.default.withDefaults
  implicit val personDecoder: Codec[Cat] = deriveConfiguredCodec
}

case class Dog(
    id: UUID,
    friends: Option[Seq[Dog]] = None
)

object Dog {
  implicit final val configuration: Configuration = Configuration.default.withDefaults
  implicit val personDecoder: Codec[Dog] = deriveConfiguredCodec
}

val fileSource = Source.fromFile("uuid-10k.json")
val fileContent = fileSource.getLines.mkString
fileSource.close()

val catStartTime = java.lang.System.currentTimeMillis()
decode[Seq[Cat]](fileContent)
println(s"Time taken for cats: ${java.lang.System.currentTimeMillis() - catStartTime} ms")

val dogStartTime = java.lang.System.currentTimeMillis()
decode[Seq[Dog]](fileContent)
println(s"Time taken for dogs: ${java.lang.System.currentTimeMillis() - dogStartTime} ms")
```
### `uuid-10k.json`
```
[{"id":"19b10dec-91fe-42ee-8da7-2cafe9529dad"},{"id":"2da8de7d-02fe-479c-8293-4bebbd24c4ef"}, ... ]
```

## Result
```
Time taken for cats: 3758 ms
Time taken for dogs: 154 ms
```
`Cat`s and `Dog`s are pretty similar. Surprisingly, one is significantly slower than the other. I believe many people use the style in the case class `Cat` to represent potentially empty list fields. This will cause problems in some production environments once they switch to `v0.14.2`.

zarthross:
[...]

PR #2030, opened 2022-09-27
Fix opsToPath and remove PathToRoot

### LINKED ISSUE #1449: Json path formatting of errors produced by non-empty collection deserializers
When a non-empty deserialiser fails and we try to convert the corresponding `List[CursorOp]` to a json path, we get this:

```scala
import io.circe.syntax._
import io.circe._
import cats.data._

CursorOp.opsToPath(
  Json.arr(0.asJson, 1.asJson, Json.Null).as[NonEmptyList[Int]].left.get.history
)

// [0]{!_/}[1]
```

This, for example, when returned in an HTTP error message, can be confusing to the client.

I've written a version of `CursorOp.opsToPath` that handles this special case (`DownArray` -> `DeleteGoParent` -> `DownArray`) and outputs `[2]` instead of `[0]{!_/}[1]`. We've been using this modified version in our app.

I was wondering if you'd be interested in having this merged into circe?
If so, I can submit a PR.

This is the modified version:

```scala
  def fromOps(history: List[CursorOp]): String = {
    val selections = foldOps(history.reverse, Nil)

    selections.foldLeft("") {
      case (str, SelectField(f)) => s".$f$str"
      case (str, SelectIndex(i)) => s"[$i]$str"
      case (str, Op(op))         => s"{${Show[CursorOp].show(op)}}$str"
    }
  }

  @tailrec
  private def foldOps(ops: List[CursorOp], acc: List[Selection]): List[Selection] =
    (ops, acc) match {
      case (Nil, _)                                              => acc
      case (DownField(k) :: tail, _)                             => foldOps(tail, SelectField(k) :: acc)
      case (DownArray :: DeleteGoParent :: DownArray :: tail, _) => foldOps(tail, SelectIndex(1) :: acc)
      case (DownArray :: tail, _)                                => foldOps(tail, SelectIndex(0) :: acc)
      case (MoveUp :: tail, _ :: accTail)                        => foldOps(tail, accTail)
      case (MoveRight :: tail, SelectIndex(i) :: accTail)        => foldOps(tail, SelectIndex(i + 1) :: accTail)
      case (MoveLeft :: tail, SelectIndex(i) :: accTail)         => foldOps(tail, SelectIndex(i - 1) :: accTail)
      case (RightN(n) :: tail, SelectIndex(i) :: accTail)        => foldOps(tail, SelectIndex(i + n) :: accTail)
      case (LeftN(n) :: tail, SelectIndex(i) :: accTail)         => foldOps(tail, SelectIndex(i - n) :: accTail)
      case (op :: tail, _)                                       => foldOps(tail, Op(op) :: acc)
    }
```

andreisilviudragnea:
Fixed the existing `CursorOp.opsToPath` to handle all types of cursors, including the special case for `DeleteGoParent`, used for decoding `NonEmpty` types.

The new logic introduced by https://github.com/circe/circe/pull/1974 for JavaScript-style paths does not fix the bugs of the existing `CursorOp.opsToPath`.

I added more tests for `ACursor.pathString` method, covering the https://github.com/circe/circe/issues/1449 bug and others specific to the `PathToRoot` implementation. `ACursor.pathString` has been reimplemented in terms of the now-fixed `CursorOp.opsToPath` and `PathToRoot` class has been deleted.

Fixes https://github.com/circe/circe/issues/1449


andreisilviudragnea:
@zmccoy @zarthross @jeffmay @travisbrown I would like to hear your opinion about #1974 and #2015 and my proposal here. I left a comment on that #1974 too.

Also, how many active maintainers does this repo have at the moment?

zarthross:
@andreisilviudragnea Thank you for your contribution to circe.  We are always grateful to have contributions.  That being said, there are some issues here in this PR that need to be addressed before we would be able to move forward.

1. This PR breaks binary compatibility with the existing Series 0.14 branch, as indicated by the Mima failures in the CI build.
2. This PR brings back 2 performance issues that were addressed by  #1974
  - Issue 1: Evaluation of the cursor path is non-trivial and causes performance issues for `or` and `either` type decoder combinators (see https://github.com/circe/circe/issues/1972) since the introduction of improved Decoder failure reason messages in https://github.com/circe/circe/pull/1617
  - Issue 2: Evaluation of the cursor ops history is less performant than using the cursor hierarchy to attempt the same goal of computing the cursors path.  This is evident when you take into account all the `MoveRight` operations that will be in the cursor op history when navigating long nested arrays.  (See below the benchmark)
3.  `showSelections` needs to use a `StringBuilder` instead of `mkString` as its less performant.  (See below benchmarks)
4. Needs to handle (somehow) the situation where `DeleteGoParent` is used in situations other than non-empty decoders patterns. (To be fair, this needs more addressing in the current `PathToRoot` implementation as well.

I will concede that the current implementation is not perfect, but there is a balancing act being performed here between  correctness and  performance.  The original `opsToPath` wasn't completely correct, and neither is `PathToRoot`.  They are both missing things (especially around `DeleteGoParent`).  This PR did point out a few cursor operations that do generate incorrect values in the current implementation and we should be able to get those fixed either way, so thank you for pointing those out. 

As for your question regarding how many active maintainers are there:
  - See https://github.com/circe/circe#contributors-and-participation
  - @zmccoy and I are the active maintainers, with significant contributions by @isomarcte and @armanbilge among others.
  - Travis has stepped down from open source Scala work, so please do not ping him on issues in Circe, he won't reply anyways.
  - jeffmay is currently the maintainer of circe-yaml, and not the main circe repo.

** Benchmarks
```
[info] Benchmark                         Mode  Cnt      Score      Error  Units
[info] v0.14.2.deepArrays                thrpt   20    6679.939 ±  134.656  ops/s
[info] v0.14.2.linear                    thrpt   20  102230.683 ± 2288.512  ops/s
[info] v0.14.3.deepArrays                thrpt   20  170606.527 ± 5768.646  ops/s
[info] v0.14.3.linear                    thrpt   20  203453.323 ± 2391.592  ops/s
[info] PR2030.deepArrays                 thrpt   20    6797.393 ±  113.063  ops/s
[info] PR2030.linear                     thrpt   20  168980.925 ± 7221.380  ops/s
[info] PR2030-string-builder.deepArrays  thrpt   20    6678.443 ±   84.574  ops/s
[info] PR2030-string-builder.linear      thrpt   20  212533.730 ± 6408.258  ops/s
```

Code to produce the benchmarks `modules/benchmark/src/main/scala-2/io/circe/benchmark/ACursorBenchmarks.scala`:
I'll try and add this file to circe at some point soon.

```scala
package io.circe.benchmark

import io.circe._
import io.circe.syntax._
import java.util.concurrent.TimeUnit
import org.openjdk.jmh.annotations._

/**
 * The following command will run the benchmarks with reasonable settings:
 *
 * > sbt "benchmark/jmh:run -i 10 -wi 10 -f 2 -t 1 io.circe.benchmark.ACursorBenchmarks"
 */
@State(Scope.Thread)
@BenchmarkMode(Array(Mode.Throughput))
@OutputTimeUnit(TimeUnit.SECONDS)
class ACursorBenchmarks {
  val dataLinear = (0 to 100).foldLeft(Json.True) { (acc, _) =>
    Json.obj("field" := acc)
  }
  val cursorLinear = (0 to 100).foldLeft[ACursor](dataLinear.hcursor) { (acc, _) =>
    acc.downField("field")
  }

  val dataLotsOfArrays = (0 to 100).foldLeft(Json.True) { (acc, _) =>
    Json.fromValues {
      (0 to 99).map(_.asJson) :+ acc
    }
  }
  val cursorLotsOfArrays = (0 to 100).foldLeft[ACursor](dataLotsOfArrays.hcursor) { (acc, _) =>
    (0 to 99).foldLeft(acc.downArray) { (acc, _) =>
      acc.right
    }
  }

  @Benchmark
  def linear: String =
    //CursorOp.opsToPath(cursorLinear.history)
    cursorLinear.pathString

  @Benchmark
  def deepArrays: String =
    //CursorOp.opsToPath(cursorLotsOfArrays.history)
    cursorLotsOfArrays.pathString
}

```

andreisilviudragnea:
Local results before this PR:
```
[info] Benchmark                      Mode  Cnt       Score      Error  Units
[info] ACursorBenchmarks.deepArrays  thrpt   20  390730.750 ± 5419.767  ops/s
[info] ACursorBenchmarks.linear      thrpt   20  470064.409 ± 1872.145  ops/s
```

Local results after fixing binary compatibility, before any optimization:
```
[info] Benchmark                      Mode  Cnt       Score      Error  Units
[info] ACursorBenchmarks.deepArrays  thrpt   20    8178.515 ±   25.232  ops/s
[info] ACursorBenchmarks.linear      thrpt   20  323991.668 ± 1284.678  ops/s
```

modules/core/shared/src/main/scala/io/circe/Error.scala
```diff
@@ -103,13 +103,7 @@ object DecodingFailure {
...
     ops: Eval[List[CursorOp]]
   ) extends DecodingFailure(lazyReason) {
-    override final def pathToRootString: Option[String] =
-      if (pathToRoot == Some(PathToRoot.empty)) {
-        // For backwards compatibility. We'll make this more consistent in 0.15.x
-        Some("")
-      } else {
-        pathToRoot.orElse(PathToRoot.fromHistory(ops.value).toOption).map(_.asPathString)
-      }
+    override final def pathToRootString: Option[String] = Some(CursorOp.opsToPath(history))
```
zarthross, line 106: by using `CursorOp.opsToPath` here aren't we using the slower version of this process now?  Seems to me that the benchmark improvement by switching some areas back to `PathToRoot` wont' be reflected here.

andreisilviudragnea: It's still work in progress.

I eventually understood the differences in performance with the tests you provided. I still don't know how to report errors clearly in some cases.

I would avoid trying to return meaningful errors for all cursor combinations, since some of them are not even used in the codebase.

PR #2041, opened 2022-10-15
(0.15.x) Add human-readable toString in errors

kubukoz:


.jvmopts
```diff
@@ -1,6 +1,6 @@
 -Dfile.encoding=UTF8
 -Xms1G
--Xmx3G
+-Xmx6G
```
kubukoz, line 3: `+test` wasn't having it with just 3 gigs of heap.

modules/core/shared/src/main/scala/io/circe/Error.scala
```diff
@@ -85,10 +84,14 @@ sealed abstract class DecodingFailure(private val lazyReason: Eval[Reason]) exte
   def reason: Reason =
     lazyReason.value
 
+  /**
+   * Creates compact, human readable string representations for DecodingFailure
+   * Cursor history is represented as JS style selections, i.e. ".foo.bar[3]"
+   */
   override final def toString: String =
-    // This should be the same as the Show instance, but we shouldn't change
-    // that until at least 0.15.x.
```
kubukoz, line 90: I would've changed it in 0.14.x if not for this comment.

kubukoz:
@zmccoy @zarthross fyi I've resolved the conflicts here

zmccoy:
@kubukoz Hey, sorry that this is so late, could I get you to rebase this onto the 0.14 branch.  @zarthross and I were talking about how we've already changed our error strings previously in a patch and we'll just follow that to not force folk to upgrade to 0.15 for their improvements.

kubukoz:
I didn't like that this one had `015` in the branch name, so I created a new one ;) #2075

Issue #1941, opened 2022-04-08
Consider switching to the Jawn parser on JS


armanbilge:
**tl;dr** The circe-jawn parser on JS has better semantics and in many cases out-performs the `JSON.parse` method currently used by circe-parser.

Benchmarks: https://armanbilge.github.io/jsoniter-scala/index-scalajs.html

Note that the above benchmark results were extracted from https://plokhotnyuk.github.io/jsoniter-scala/index-scalajs.html; all credit goes to @plokhotnyuk.

---

In https://github.com/circe/circe/pull/1791, the circe-jawn module was cross-built for JS. However, the circe-parser module continued to use the standard JS API [`JSON.parse`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse) for parsing and this was "unlikely to change" according to https://github.com/typelevel/jawn/pull/351#issuecomment-873123158. Presumably this was for performance-related reasons:
1. `JSON.parse` is provided by the runtime, so no parsing code has to be bundled in applications' size-sensitive JS artifacts.
2.  Runtimes are free to implement `JSON.parse` with highly-optimized native code, so it should be fast.

However, as noted in the circe docs, the use of `JSON.parse` also has some serious caveats:

https://github.com/circe/circe/blob/a015255b73d23505d5155deb5b09121d8d913cf6/docs/src/main/tut/parsing.md?plain=1#L61-L69

This (often surprising) difference in semantics can cause problems such as https://github.com/circe/circe/issues/393 and https://github.com/circe/circe/issues/1911#issuecomment-1069346826 and has "raised eyebrows".

> I'm surprised circe considers it better to be fast than correct, but it's their call. Longs themselves are correct in Scala.js. Apparently it's not correct to serialize Longs using the default mechanism of circe if you actually use their full range. Instead, using strings or a pair of Ints may work.

@sjrd in https://discord.com/channels/632150470000902164/635668814956068864/905900410517204992

Note that the circe-jawn.js parser _does not_ have these issues and instead has semantics that match the JVM, since it is parsing from `String` to the `Json` AST with exactly the same code.

[...]

PR #2042, opened 2022-10-16
Use the Jawn parser on JS

armanbilge:
Closes https://github.com/circe/circe/issues/1941.


modules/parser/js/src/main/scala/io/circe/parser/package.scala
```diff
@@ -1,17 +0,0 @@
...
-  final def parse(input: String): Either[ParsingFailure, Json] = (
-    try convertJsToJson(JSON.parse(input))
-    catch {
-      case NonFatal(exception) => Left(ParsingFailure(exception.getMessage, exception))
-    }
-  ) match {
-    case r @ Right(_)    => r.asInstanceOf[Either[ParsingFailure, Json]]
-    case Left(exception) => Left(ParsingFailure(exception.getMessage, exception))
-  }
-}
```
zarthross: This may be more complicated, but rather than removing this, would it make more sense to keep it around for some users if they want to use the old `JSON.parse` version?  Certainly switch to jawn by default.

armanbilge: Good point. I can move this to the `circe-scalajs` module, wdyt?

zarthross: That make sense to me.

zarthross (approved):

zmccoy (approved):

PR #2045, opened 2022-10-22
Use sbt circe org for building

lorandszakacs:
- [x] use sbt-circe-org sbt plugin for our build needs
- [ ] update docs and publish to new format
- [x] fix license header updating. Currently fails :sweat_smile: with `Not able to locate License file` or some sort. We do have a problem here in that I see some modules use a different license. Maybe we should disable license header management here?

Notable changes:
- the plugin drops scalastyle in favor of more compiler flags and scalafix
- we now have a bunch of warnings that will need fixing


armanbilge:
> * fix license header updating. Currently fails 😅 with `Not able to locate License file` or some sort. We do have a problem here in that I see some modules use a different license. Maybe we should disable license header management here?

I'm not sure what you mean about difference licenses. The only problem is that the license is specified with the wrong name. Actually you can just delete this line since ASL v2 is already the default.

https://github.com/circe/circe/blob/cee926f65eb63f0dd82ce2bd1beeb5dbb8f9c4f3/build.sbt#L455

build.sbt
```diff
@@ -1,59 +1,23 @@
...
-import microsites.ConfigYml
 import sbtcrossproject.{ CrossProject, CrossType }
 
 val Scala212V: String = "2.12.15"
 val Scala213V: String = "2.13.8"
 val Scala3V: String = "3.1.3"
 
 ThisBuild / tlBaseVersion := "0.14"
-ThisBuild / tlCiReleaseTags := false
+ThisBuild / tlCiReleaseTags := false // flip to true once we get the automated publishing process up and running
```
armanbilge:
```suggestion
ThisBuild / tlCiReleaseBranches := Seq() // set to `series/0.14.x` once we get the automated publishing process up and running
ThisBuild / tlCiReleaseTags := false // flip to true once we get the automated publishing process up and running
```

build.sbt
```diff
@@ -125,100 +86,20 @@ def circeCrossModule(path: String, crossType: CrossType = CrossType.Full) = {
...
-  gitHubPagesOrgName := "circe",
-  gitHubPagesRepoName := "circe",
-  gitHubPagesSiteDir := baseDirectory.value / "target" / "site",
-  /* } Publish GitHub Pages */
-  scalacOptions ~= {
-    _.filterNot(Set("-Yno-predef"))
-  },
-  git.remoteRepo := "git@github.com:circe/circe.git",
-  ScalaUnidoc / unidoc / unidocProjectFilter := inProjects(
-    numbers.jvm,
```
armanbilge, line 201: Instead of deleting this unidoc stuff, let's move it into a published unidoc artifact.
https://typelevel.org/sbt-typelevel/site.html#how-can-i-publish-unidoc-api-docs

build.sbt
```diff
@@ -125,100 +86,20 @@ def circeCrossModule(path: String, crossType: CrossType = CrossType.Full) = {
...
     moduleName := "circe-docs",
     name := "Circe docs",
-    mdocIn := file("docs/src/main/tut"),
     libraryDependencies ++= Seq(
       "io.circe" %% "circe-generic-extras" % "0.14.1",
       "io.circe" %% "circe-optics" % "0.14.1"
     )
   )
-  .settings(docSettings)
   .enablePlugins(NoPublishPlugin)
```
armanbilge: The site plugin should enable no-publish automatically.
```suggestion
```

zmccoy:
@zarthross I think you need a `githubWorkflowGenerate` as the ci.yml isn't up to date.

mzuehlke:
I guess it would make review of this PR easier, if the header changes are done in a separate PR ?
What is still missing here ? Am I able to help ?

zmccoy (approved):

PR #2052, opened 2022-11-05
Remove redundant time arbitraries

### LINKED ISSUE #1994: Drop Custom Arbitrary Instances For java.time

armanbilge:
We get these from Scalacheck now.

Closes https://github.com/circe/circe/issues/1994.


armanbilge:
Turns out that some of the encoders were generating output that was not parsable by the decoders. So I fixed that.

armanbilge:
Hm, another round-trip bug uncovered by this ...
```
scala> java.time.Duration.parse("PT-1399862021242856H-57M-0.278965764S")
val res0: java.time.Duration = PT-1399862021242856H-57M0.278965764S
```

armanbilge:
Ok, I think this was a JDK bug that was fixed in 9+ but never backported into JDK 8.

~~https://github.com/openjdk/jdk/commit/2cfdb952517459412ce488929e9b44910930b9f5~~ Edit: no, I think what I am observing is a parser bug, not a `toString` bug.

I'd argue this is kind of not our problem ... fixing this is highly non-trivial, since we'd have to cleanroom implement a method that does exactly what the JDK does without using the JDK's GPL sources :)

modules/core/shared/src/main/scala/io/circe/Decoder.scala
```diff
@@ -1222,12 +1223,136 @@ object Decoder
...
 
+  private[this] lazy val DurationPattern = Pattern.compile(
+    "([-+]?)P(?:([-+]?[0-9]+)D)?" + "(T(?:([-+]?[0-9]+)H)?(?:([-+]?[0-9]+)M)?(?:([-+]?[0-9]+)(?:[.,]([0-9]{0,9}))?S)?)?",
+    Pattern.CASE_INSENSITIVE
+  )
+
+  private[this] def parseDuration(text: CharSequence): Duration = {
+
+    /*
+     * Copyright (c) 2007-present, Stephen Colebourne & Michael Nascimento Santos
```
armanbilge: So this is annoying ... either this, or we just don't test this on JDK 8, cause it's broken for reasons out of our control /shrug.

zarthross: Its nice that we kinda fix a bug, for our JDK8 users with this, but ... for organizational purposes, can we move this code out of this file into its own class in a way we can remove it later without breaking bincompat?

armanbilge: Eh, I just removed it and disabled the relevant tests on JDK8. I should have follow my gut instinct and left it alone.

armanbilge:
Reported flake in https://github.com/circe/circe/issues/2074.

zarthross (approved):

zmccoy (approved):

PR #2063, opened 2022-11-23
Better error accumulation in Decoder#or and Decoder#either

morgen-peschke:
The documented way to create a `Decoder` for an ADT is to merge them using `Decoder#or`, as seen in this example from the microsite:

```scala
  List[Decoder[Event]](
    Decoder[Foo].widen,
    Decoder[Bar].widen,
    Decoder[Baz].widen,
    Decoder[Qux].widen
  ).reduceLeft(_ or _)
```

The errors produced from this `Decoder` are surprising, as only the errors for `Qux` will be returned:
```scala
println(decodeAccumulating[Event]("""{ "i": null }""").show)
// Invalid(NonEmptyList(DecodingFailure at .values: Missing required field))
```

The fix for this is to adjust `Decoder#or` to merge the failures, which produces a more useful error:
```scala
println(decodeAccumulating[Event]("""{ "i": null }""").show)
// Invalid(NonEmptyList(DecodingFailure at .values: Missing required field, DecodingFailure at .c: Missing required field, DecodingFailure at .s: Missing required field, DecodingFailure at .i: Int))
```


morgen-peschke:
Note to self: figure out if #1243 can be done in a binary-compatible way to keep the reported errors from getting unmanageable.

morgen-peschke:
@zarthross : I think I found a way to make it work, however before propagating these changes into the shapeless stuff, I wanted to get feedback on the approach.

`coreJVM/mimaReportBinaryIssues` comes up clean, which is promising.

It's currently sitting in a PR against my branch: morgen-peschke/circe#1

hamnis:
Does currently does not seem to have a way forward, closing. If this was done in error, please reopen.

PR #2078, opened 2023-01-26
Normalize Module Name

isomarcte:
This commit adds a new function to the `build.sbt` which normalizes the module name in the cases where the module name is derived from a path with a sub-directory, and thus will include a `'/'` character. These characters are replaced with `'-'`, which is both valid and probably want we intended in the first place.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -87,15 +87,16 @@ lazy val allSettings = Seq(
   Compile / scalastyleSources ++= (Compile / unmanagedSourceDirectories).value
 )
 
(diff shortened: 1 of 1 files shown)


isomarcte:
@zarthross @zmccoy this should fix the failure in https://github.com/circe/circe/pull/2072

zarthross (dismissed):

zarthross (approved):

zmccoy:
Thanks @isomarcte <3

Issue #2088, opened 2023-02-09
StackOverflowError since 0.14.4


xuwei-k:
### `build.sbt`

```scala
libraryDependencies += "io.circe" %% "circe-generic" % "0.14.4"

scalaVersion := "3.2.2"
```

### `Main.scala`

```scala
package example

import io.circe.generic.semiauto.deriveEncoder
import io.circe.Encoder

object Main {
  def a: List[String] = List(2).map { i =>
    case class X(x: Int)

    object X {
      implicit val encoder: Encoder[X] = deriveEncoder[X]
    }

    X.encoder.apply(X(i)).toString
  }

  def main(args: Array[String]): Unit = {
    println(a)
  }
}
```

### `sbt run` with circe 0.14.3

```
[info] running example.Main 
List({
  "x" : 2
})
[success] Total time: 1 s, completed Feb 9, 2023, 7:13:44 AM
```

### `sbt run` with circe 0.14.4

```
[info] running example.Main 
Exception in thread "sbt-bg-threads-1" java.lang.StackOverflowError
	at example.Main$.example$Main$$$_$X$2(Main.scala:10)
	at example.Main$X$3$.<init>(Main.scala:11)
	at example.Main$.X$lzyINIT1$1(Main.scala:10)
	at example.Main$.example$Main$$$_$X$2(Main.scala:10)
	at example.Main$X$3$.<init>(Main.scala:11)
	at example.Main$.X$lzyINIT1$1(Main.scala:10)
	at example.Main$.example$Main$$$_$X$2(Main.scala:10)
	at example.Main$X$3$.<init>(Main.scala:11)
	at example.Main$.X$lzyINIT1$1(Main.scala:10)
	at example.Main$.example$Main$$$_$X$2(Main.scala:10)
	at example.Main$X$3$.<init>(Main.scala:11)
	at example.Main$.X$lzyINIT1$1(Main.scala:10)
	at example.Main$.example$Main$$$_$X$2(Main.scala:10)
	at example.Main$X$3$.<init>(Main.scala:11)
	at example.Main$.X$lzyINIT1$1(Main.scala:10)
	at example.Main$.example$Main$$$_$X$2(Main.scala:10)
	at example.Main$X$3$.<init>(Main.scala:11)
	at example.Main$.X$lzyINIT1$1(Main.scala:10)
	at example.Main$.example$Main$$$_$X$2(Main.scala:10)
	at example.Main$X$3$.<init>(Main.scala:11)
[...]

PR #2092, opened 2023-02-09
Fix Issue #2088

zarthross:
Fix the runtime stack overflow in issue #2088 but inlining more of the mirrors in the new macro methods.


zmccoy (approved):

zmccoy:
Failure is the docs build that we know is an issue currently.

PR #2098, opened 2023-02-14
fixed 2096

### LINKED ISSUE #2096: [Bug] discriminator fails on class hierarchies with more than one level.
The code somehow picked up the middle-level abstract type name as the value for the discriminator rather than the concrete type. 

Here is a test
```scala
  {
    given Configuration = Configuration.default.withDiscriminator("type")
    sealed trait GrandParent derives ConfiguredCodec
    object GrandParent:
      given Eq[GrandParent] = Eq.fromUniversalEquals

    sealed trait Parent extends GrandParent

    case class Child(a: Int) extends Parent

    test("Codec for hierarchy of more than 1 level with discriminator should encode and decode correctly") {
      val child: GrandParent = Child(1)
      val json = Encoder.AsObject[GrandParent].apply(child)
      val result = Decoder[GrandParent].decodeJson(json)
      assert(result === Right(child), result)
    }
  }
```

The decode result:
```
Left(
  value = DecodingFailure at : type Parent has no class/object/case named 'Parent'.
)
```
cc @Lasering, sorry to bother you again. 

Update: I submitted PR #2098  to fix.

kailuowang:
Fixes #2096 
First time playing with Scala 3 derivation. Apologize in advance for potential silly mistake.


modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredEncoder.scala
```diff
@@ -22,8 +22,10 @@ trait ConfiguredEncoder[A](using conf: Configuration) extends Encoder.AsObject[A
     val (constructorName, json) = encodeElemAt(index, a, conf.transformConstructorNames)
     conf.discriminator match
       case None => JsonObject.singleton(constructorName, json)
-      case Some(discriminator) =>
-        json.asObject.getOrElse(JsonObject.empty).add(discriminator, Json.fromString(constructorName))
+      case Some(discriminator) => {
+        val jo = json.asObject.getOrElse(JsonObject.empty)
+        if(jo.contains(discriminator)) jo else jo.add(discriminator, Json.fromString(constructorName))
```
kailuowang: This is to prevent the Encoder of the mid-level from overriding the discriminator. Although, I am not sure how stable this order of Encoder is. There might be better ways.

Lasering:
You need to employ the same strategy in the Encoder otherwise the discriminator will not have the correct value.

kailuowang:
> You need to employ the same strategy in the Encoder otherwise the discriminator will not have the correct value.

I think the new implementation addressed this concern.

zarthross:
@kailuowang, thanks for the work on this!  The tests appear to be hanging for Scala 3 in CI, mind taking a look?  I tried locally and got the same result.

kailuowang:
> @kailuowang, thanks for the work on this! The tests appear to be hanging for Scala 3 in CI, mind taking a look? I tried locally and got the same result.

@zarthross thanks! I think it's fixed (weird it didn't hang for me before)

modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredDecoder.scala
```diff
@@ -19,10 +20,22 @@ trait ConfiguredDecoder[A](using conf: Configuration) extends Decoder[A]:
 
   /** Decodes a class/object/case of a Sum type handling discriminator and strict decoding. */
   private def decodeSumElement[R](c: HCursor)(fail: DecodingFailure => R, decode: Decoder[A] => ACursor => R): R =
+
     def fromName(sumTypeName: String, cursor: ACursor): R =
-      constructorNames.indexOf(sumTypeName) match
-        case -1 => fail(DecodingFailure(s"type $name has no class/object/case named '$sumTypeName'.", cursor.history))
-        case index => decode(elemDecoders(index).asInstanceOf[Decoder[A]])(cursor)
+      def findDecoder(decoder: Decoder[?], searched: List[ConfiguredDecoder[?]]): Option[Decoder[?]] =
```
zarthross: It looks to me like the logic in here doesn't take into consideration if the `ConfiguredDecoder` is a Sum or Product decoder, and will accidentally use the decoder for a field from a Product type rather than only looking at the class names of the Sums.  

Here is 2 tests, the second is without a configured discriminator and works fine:

```scala

  {
    sealed trait GreatGrandParent
    object GreatGrandParent:
      given Eq[GreatGrandParent] = Eq.fromUniversalEquals[GreatGrandParent]

    sealed trait GrandParent extends GreatGrandParent
    case class Uncle(Child: Int) extends GrandParent  // The field name, `Child` matches a existing case class in the hierarchy and is important for the tests.
    sealed trait Parent extends GrandParent
    
    case class Child(a: Int, b: String) extends Parent

    test("Codec for hierarchy of more than 2 level with discriminator should encode and decode correctly, even if a parent's sibling has a field with the same name as a Child type") {
      given Configuration = Configuration.default.withDiscriminator("type")
      given Codec.AsObject[GreatGrandParent] = ConfiguredCodec.derived[GreatGrandParent]

      val child: GrandParent = Child(1, "a")
      val json = Encoder.AsObject[GreatGrandParent].apply(child)
      val result = Decoder[GreatGrandParent].decodeJson(json)
      assert(result === Right(child), result)
    }

    test("Codec for hierarchy of more than 2 level should encode and decode correctly, even if a parent's sibling has a field with the same name as a Child type") {
      given Configuration = Configuration.default
      given Codec.AsObject[GreatGrandParent] = ConfiguredCodec.derived[GreatGrandParent]

      val child: GrandParent = Child(1, "a")
      val json = Encoder.AsObject[GreatGrandParent].apply(child)
      val result = Decoder[GreatGrandParent].decodeJson(json)
      assert(result === Right(child), result)
    }
  }
```

Here is the result:

```
  + Codec for hierarchy of more than 1 level with discriminator should encode and decode correctly 0.007s
==> X io.circe.ConfiguredDerivesSuite.Codec for hierarchy of more than 2 level with discriminator should encode and decode correctly, even if a parent's sibling has a field with the same name as a Child type  0.047s munit.FailException: /Users/darren.gibson/Projects/circe/circe/modules/tests/shared/src/test/scala-3/io/circe/ConfiguredDerivesSuite.scala:467
466:      val result = Decoder[GreatGrandParent].decodeJson(json)
467:      assert(result === Right(child), result)
468:    }
Left(
  value = DecodingFailure at : Int
)
    at munit.FunSuite.assert(FunSuite.scala:11)
    at io.circe.ConfiguredDerivesSuite.$init$$$anonfun$27(ConfiguredDerivesSuite.scala:467)
  + Codec for hierarchy of more than 2 level should encode and decode correctly, even if a parent's sibling has a field with the same name as a Child type 0.026 
```

zarthross (changes requested):

modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredDecoder.scala
kailuowang: Good catch, thank you @zarthross. I was trying to avoid breaking binary compatibility with those hacky implementations. 
I just found of a way to expand the internal API while preserving bin compat. 
Also added your test.

modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredDecoder.scala
```diff
@@ -19,10 +20,22 @@ trait ConfiguredDecoder[A](using conf: Configuration) extends Decoder[A]:
...
+        decoder match {
+          case cd: ConfiguredDecoder[?] with SumOrProduct if !searched.contains(cd) && cd.isSum =>
```
zarthross: I feel like its probably faster to check if `cd` is a sum type first and know if you need to bother checking the list.
```suggestion
          case cd: ConfiguredDecoder[?] with SumOrProduct if cd.isSum && !searched.contains(cd) =>
```

modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredDecoder.scala
zarthross: I wonder if we can pre-optimize this at initialization time by pre-walking the tree and making a `Map[Discriminator, Decoder[?]]` and storing it in a `lazy val` rather than effectively re-walking the tree for every decode?

I know our examples here aren't particularly deep, but when we do often get users using circe in unexpected ways, with deep nested ADTs, or decoders in tight loops where performance matters a lot.

kailuowang: Good point. Just made the change.  Also noticed that I no longer need to store searched Decoders to avoid stack overflow in recursive structures since the chain of recursive must involve a `Product`, which breaks the recursive loop.

modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredDecoder.scala
```diff
@@ -14,15 +16,30 @@ trait ConfiguredDecoder[A](using conf: Configuration) extends Decoder[A]:
...
+  private lazy val decodersDict: Map[String, Decoder[?]] = {
+    def findDecoderDict(p: (String, Decoder[?])): List[(String, Decoder[?])] =
+      p._2 match {
+        case cd: ConfiguredDecoder[?] with SumOrProduct if cd.isSum =>
+          cd.constructorNames.zip(cd.elemDecoders).flatMap(findDecoderDict)
+        case _ => List(p)
+      }
+    constructorNames.zip(elemDecoders).flatMap(findDecoderDict).toMap
+  }
+
```
zarthross, line 28: I'm a little concerned that this doesn't handle recursive types such as:

```
sealed trait Tree; 
case class Branch(l: Tree, r: Tree) extends Tree; 
case object Leaf extends Tree;
```

But then I tried it locally and something else unrelated blows up first in the inlining, so this implementation is probably sufficient for now.

zarthross (dismissed):
Thanks again for the work on this!  Greatly appreciated!

modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredDecoder.scala
kailuowang: Thanks! There was a recursive test stackoverflowed with my implementation before I changed to the code to stop traversing the tree at a `Product` decoder.  I am not sure if that change fixed it for the above scenario, but if we need to come back to this again feel free to ping me. Will be happy to continue improve it.

zmccoy (dismissed):

modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredDecoder.scala
Lasering: Having a test for the code above would solve any possible concern :)

kailuowang:
@zarthross, I just added a test for recursive ADT. Looks to me that it's ready for merge?

zmccoy (approved):

zarthross (approved):

zarthross:
Thanks @kailuowang

kailuowang:
> Thanks @kailuowang

Thank you! Is there, by any chance, a release scheduled in the near future?

PR #2099, opened 2023-02-16
Add Dassault Systems to list of adopters

batakpout:
We are using circe in our [Medidata](https://www.medidata.com/en/) Product :)


zmccoy (dismissed):
@batakpout Thanks for the addition!

zmccoy (approved):

PR #2107, opened 2023-03-03
Optimize JSON equality fail fast

agourlay:
I maintain a library that performs a large number of `JSON` equality tests to the extend that it shows in performance profiles.

tomap

The main issue is that a Scala `Map` is created every time two `JsonObject` backed by a `LinkedHashMapJsonObject` are compared.

The first commit proposes to detect possible differences by first comparing the size of the fields which is O(1), thus avoiding the creation of additional Maps.

A second commit applies a similar strategy to the `JArray` comparison to shortcut the Vector traversals. 

Those two optimizations should be able to save a large number of allocations on trivial failing comparisons.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Json.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Json.scala
@@ -584,13 +584,14 @@ object Json {
    */
   private[this] def isReal(value: Float): Boolean = java.lang.Float.isFinite(value)
 
-  private[this] final def arrayEq(x: Seq[Json], y: Seq[Json]): Boolean = {
+  private[this] final def arrayEq(x: Vector[Json], y: Vector[Json]): Boolean = {
+    if (x.length != y.length) return false
     val it0 = x.iterator
     val it1 = y.iterator
-    while (it0.hasNext && it1.hasNext) {
+    while (it0.hasNext) {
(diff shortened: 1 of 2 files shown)


modules/core/shared/src/main/scala/io/circe/Json.scala
```diff
@@ -584,13 +584,14 @@ object Json {
...
+    if (x.length != y.length) return false
     val it0 = x.iterator
     val it1 = y.iterator
-    while (it0.hasNext && it1.hasNext) {
+    while (it0.hasNext) {
       if (Json.eqJson.neqv(it0.next(), it1.next())) return false
     }
-    it0.hasNext == it1.hasNext
+    true
   }
```
zarthross, line 595: I wonder if we could just call into the standard library equality check for vectors and make this a shorter/faster implementation?

agourlay: Interesting point, I think it would definitely make the code shorter, not sure about squeezing extra performance.

At the same time, I do not have the context regarding why it was done like this in first place.
The spirit of the PR is to shortcut the existing implementation, not to modify it.
 
Pushing for a bigger change would require a JMH benchmark IMHO.

zarthross (approved):

zmccoy (approved):

PR #2115, opened 2023-03-19
Quick shortcut Optimisation: avoid StringBuilder

diesalbla:
Given that the defaults for the parameters passed to concat are empty strings, it may often be the case that we would not need to allocate the StringBuilder.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Printer.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Printer.scala
@@ -86,19 +86,27 @@ final case class Printer(
     final def onNumber(value: JsonNumber): Unit = writer.append(value.toString)
   }
 
-  private[this] final def concat(left: String, text: String, right: String): String =
-    if (left.isEmpty && right.isEmpty)
-      text
-    else {
-      val builder = new StringBuilder()
-      builder.append(left)
-      builder.append(text)
-      builder.append(right)
-      builder.toString
-    }
-
   private[this] final val pieces: Printer.PiecesAtDepth =
-    if (indent.isEmpty)
+    if (indent.isEmpty) {
+      var builder: StringBuilder = null
+
+      def concat(left: String, text: String, right: String): String =
+        if (left.isEmpty && right.isEmpty)
+          text
+        else {
+          val size = left.length + text.length + right.length
+          if (builder == null)
+            builder = new StringBuilder(Math.max(16, size))
+          else {
+            builder.setLength(0)
+            builder.ensureCapacity(left.length + text.length + right.length)
+          }
+          builder.append(left)
+          builder.append(text)
+          builder.append(right)
+          builder.toString
+        }
+
       new Printer.ConstantPieces(
         Printer.Pieces(
           concat(lbraceLeft, openBraceText, lbraceRight),
@@ -111,7 +119,7 @@ final case class Printer(
           concat(colonLeft, colonText, colonRight)
         )
       )
-    else
+    } else
       new Printer.MemoizedPieces(indent) {
         final def compute(i: Int): Printer.Pieces = {
           val builder = new StringBuilder()
--- a/modules/core/shared/src/main/scala/io/circe/Printer.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Printer.scala
@@ -86,13 +86,16 @@ final case class Printer(
     final def onNumber(value: JsonNumber): Unit = writer.append(value.toString)
   }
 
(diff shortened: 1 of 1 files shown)


zarthross:
2 quick things, can 1 we get a benchmark result showing that this actually improves something, and 2 can you branch this off series/0.14 instead of series/0.15.

diesalbla:
Get-rebasing onto 0.14 should be no problem. 

> can 1 we get a benchmark result showing that this actually improves something

What makes it likely it would not improve things? It is saving the allocation of an object, the `StringBuilder`, which internally allocates an `Array[Byte]`. A different question is whether that use case is common in industrial usage.

armanbilge:
> What makes it likely it would not improve things?

The introduction of a conditional can negatively impact performance.

> It is saving the allocation of an object, the `StringBuilder`, which internally allocates an `Array[Byte]`.

Sure, but for all we know the JIT could be using escape analysis to do scalar replacement.

That's not to say this is/isn't a win, it's just really hard to know without empirical data :)

diesalbla:
> The introduction of a conditional can negatively impact performance.

Boolean tests between stack variables are negligible. Loading the array length field into the stack has a bit of cost, but one that would be paid anyway. This would always be lower than allocating an object and an array; plus the cost of garbage-collecting it afterwards.

> Sure, but for all we know the JIT could be using escape analysis to do scalar replacement.

How would that work? Would that turn the `StringBuilder` fields into stack variables? Even in that case, it cannot inline the internal array, as Java arrays are always in the heap.

armanbilge:
https://shipilev.net/jvm/anatomy-quarks/18-scalar-replacement/

diesalbla:
@armanbilge Looks like a good reference, thanks.

armanbilge:
> Boolean tests between stack variables are negligible.

To clarify: it's not the cost of executing the predicate. It's the cost of the branch itself.

https://www.baeldung.com/java-branch-prediction#what_are_the_hazards

Again, this is not to say your changes are not an optimization. It may very well be an excellent one! There's just non-trivial things involved, and it requires benchmarks to really be sure :)

zmccoy:
@diesalbla Added the on hold label for now while we're figuring this out.  Thank you!

diesalbla:
> To clarify: it's not the cost of executing the predicate. It's the cost of the branch itself.


So, regarding branch prediction, that refers to the pipeline of instructions in a CPU. That is, that if branch prediction fails it has to pull a few instructions back from that pipeline. However, allocating an object into heap, even if that is just the Thread Local Allocation Buffer (TLAB) may take a few dozens of those CPU instructions, in looking for an available spot, then changing the state of the heap (a few memory writes), then cleaning up that memory. Later on, collecting that memory also issues a lot of CPU instructions, far more of what a failed branch prediction takes. It is on a different scale.

diesalbla:
@zarthross What hypothesis should be verified using a benchmark? Checking that avoiding allocating memory would result in less memory consumption seems trivial, the question is how much of a proportion is it that it reduces. Is the concern, on the other hand, how much the conditional charges on the case where the optimisation does not apply?

rossabaker:
If you're microoptimizing that, sizing the StringBuilder's constructor would help... and scalac might give it to you for free with `left + text + right`.

diesalbla:
> If you're microoptimizing that, sizing the StringBuilder's constructor would help... and scalac might give it to you for free with left + text + right.

I am not sure I understand... Is that a suggestion to do further changes, or to do something completely different?

rossabaker:
The constructor to StringBuilder takes an Int argument.  A correct size prevents wasting bytes by being too big or copies by being too small.

diesalbla:
Well, the PR is starting to pull a thread... a few changes: 
- Made `concat` into a local nested method.
- Extracted the `StringBuilder` used in `concat` as a shared value, so 
- Delay initialisation of that `StringBuilder` until and unless it is first needed. 
- Use the sizes of the strings to compute its size.

> The constructor to StringBuilder takes an Int argument. A correct size prevents wasting bytes by being too big

It is worth mentioning that, since that is an array of bytes, the memory waste from 16 to 32 is not too great, and for GC purposes its size is not a problem since a byte array is itself a single terminal node (no outgoing links).

krux02:
As an outsider, this is a pretty obvious case. It is a trivial check for a potential big performance improvement, it is just that there is some cost in code reability that should be justified. For me (who doesn't know the codebase) it would just be enough to know that calls to concat with empty left and right actually do happen.

Issue #2122, opened 2023-04-01
Providing cats.Defer instances for Decoder and Encoder to make working with recursive structures easier


morgen-peschke:
Certain shapes of data are difficult to write correct `Decoder`/`Encoder` instances for, and providing a `Defer` instance for `Decoder` and `Encoder` can make this much simpler to do correctly.

# Problem Statement
The easiest way to see this is working through an example. I'll use the same data structure and sample JSON for each example:

```scala
import cats.Show
import io.circe.Json
import io.circe.syntax._

sealed trait Tree[A]
object Tree {
  final case class Branch[A](left: Tree[A], right: Tree[A]) extends Tree[A]
  final case class Leaf[A](value: A) extends Tree[A]

  implicit def show[A]: Show[Tree[A]] = Show.fromToString

  val json: Json =
    Json.obj(
      "left" -> Json.obj(
        "left" -> Json.obj(
          "left" -> Json.obj("value" := 1),
          "right" -> Json.obj("value" := 2)
        ),
        "right" -> Json.obj("value" := 3)
      ),
      "right" -> Json.obj("value" := 4)
    )
}
```

## Attempt 1: Write it by the book

A naive decoder, written using the examples from the [`ADT` section of the guide](https://circe.github.io/circe/codecs/adt.html) doesn't work as expected. Because the implicit definition has to be a `def`, "tying the knot" doesn't work, and this produces a `StackOverflowError` at runtime:

```scala
implicit def branchDecoder[A: Decoder]: Decoder[Branch[A]] =
  Decoder.forProduct2[Branch[A], Tree[A], Tree[A]]("left", "right")(Branch.apply)

implicit def leafDecoder[A: Decoder]: Decoder[Leaf[A]] =
  Decoder[A].at("value").map(Leaf(_))

implicit def treeDecoder[A: Decoder]: Decoder[Tree[A]] =
  List[Decoder[Tree[A]]](
    Decoder[Branch[A]].widen,
    Decoder[Leaf[A]].widen
  ).reduce(_ or _)

println {
  try Decoder[Tree[Int]].tryDecodeAccumulating(Tree.json.hcursor).show
  catch {
    case e: StackOverflowError => e
  }
}
```

[scastie](https://scastie.scala-lang.org/tp95rAhDR0WH9sQyjeDMZA)

## Attempt 2: Force it to be lazy

[...]

PR #2123, opened 2023-04-01
Add Defer instances for Decoder and Encoder

morgen-peschke:
Implements #2122


modules/tests/shared/src/test/scala/io/circe/DecoderSuite.scala
```diff
@@ -763,4 +764,6 @@ class DecoderSuite extends CirceMunitSuite with LargeNumberDecoderTestsMunit {
     assert(result.isInvalid)
     assertEquals(result.swap.toOption.map(_.size), Some(2))
   }
+
+  checkAll("Defer[Decoder]", DeferTests[Decoder].defer[MiniInt])
```
zarthross, line 784: Can you add some property tests for recursive datastructures like what was used in the example?  Just to verify further the stack safety and serve as an example for users on how to user defer?

modules/core/shared/src/main/scala/io/circe/Decoder.scala
```diff
@@ -1433,11 +1443,27 @@ object Decoder
       protected[this] final def parseUnsafe(input: String): ZoneOffset = ZoneOffset.of(input)
     }
 
+  private final case class DeferredDecoder[A](decoder: () => Decoder[A]) extends Decoder[A] {
+    override def apply(c: HCursor): Result[A] = decodeAccumulating(c).toEither.leftMap(_.head)
```
zarthross: I know it seems redundant, but can you reimplement the function you have in `decodeAccumulating` here as well?  The reason is `apply`/non-accumulating should fail FAST/on the first error, while this implementation still runs all the decoders before failing, but only returns the first error.

zarthross (changes requested):

zmccoy:
Could we add some documentation as well to the docs for this?  You could easily transform what you have in the accommodating issue and it would be great for discoverability of this.

modules/core/shared/src/main/scala/io/circe/Decoder.scala
morgen-peschke: This was a good opportunity to add the caching that was brought up in [typelevel/cats#4414](https://github.com/typelevel/cats/pull/4414)

modules/tests/shared/src/test/scala/io/circe/DecoderSuite.scala
morgen-peschke: I added non-property test, as generating recursive data structures are kind of annoying to write `Gen`s for.

Not impossible though, so if you feel strongly that they need to be property tests, I can take a pass at converting them.

zarthross:
@morgen-peschke Thanks for adding the tests.  It looks like this build failed MIMA for some reason, do you mind taking a look?

modules/core/shared/src/main/scala/io/circe/Decoder.scala
```diff
@@ -1433,11 +1443,27 @@ object Decoder
...
+  }
+
   /**
    * @group Instances
    */
-  implicit final val decoderInstances: SemigroupK[Decoder] with MonadError[Decoder, DecodingFailure] =
-    new SemigroupK[Decoder] with MonadError[Decoder, DecodingFailure] {
+  implicit final val decoderInstances
+    : SemigroupK[Decoder] with MonadError[Decoder, DecodingFailure] with Defer[Decoder] =
+    new SemigroupK[Decoder] with MonadError[Decoder, DecodingFailure] with Defer[Decoder] {
```
armanbilge: I think you'll need to do something like this to fix bincompat

```suggestion
  @deprecated
  def decoderInstances: MonadError[Decoder, DecodingFailure] = decoderInstancesBinCompat
  
  implicit final val decoderInstancesBinCompat = 
    : SemigroupK[Decoder] with MonadError[Decoder, DecodingFailure] with Defer[Decoder] =
    new SemigroupK[Decoder] with MonadError[Decoder, DecodingFailure] with Defer[Decoder] {
```

modules/tests/shared/src/test/scala/io/circe/DecoderSuite.scala
zarthross: I think we need this test to have some depth to it.  We need to have the test ensure the stack safety of the implementation of defer.  Its probably fine to use `List` as the type the way you did, but using a generated list of some significant length should do the trick.

modules/core/shared/src/main/scala/io/circe/Decoder.scala
zarthross: Alternatively, we could just make a separate val for the `Defer` couldn't we?  `Defer` has no parents and no children so there shouldn't be any real need for it to be part of `decoderInstances` right?

armanbilge: Sure, good point!

zarthross: Also, does the order matter on those `with` clauses?  Maybe moving `Defer[Decoder]` to before `MonadError[Decoder, DecodingFailure]` instead of after?

armanbilge: I'm not sure. But I do know that Scala 2 and Scala 3 have different algorithms for this, so it may be impossible to reconcile in a way to make all versions happy. We had a similar issue in Cats (actually we accidentally broke bincompat on Scala 3 since weren't checking it back then ... )

armanbilge: Ref:
- https://github.com/typelevel/cats/issues/4062#issuecomment-984064544

modules/tests/shared/src/test/scala/io/circe/DecoderSuite.scala
morgen-peschke: Clarification question: are you concerned about the stack safety of the `Defer` instance, or the `Decoder` it's used to write?

The former is tested by the `Defer` laws: 
```
  + Defer[Decoder]: defer.defer Identity 0.156s
  + Defer[Decoder]: defer.defer does not evaluate 0.001s
  + Defer[Decoder]: defer.defer is stack safe 0.319s
  + Defer[Decoder]: defer.defer matches fix 0.429s
```

The later I think may be out of our control.

hamnis:
would you mind fixing the merge conflics here?

hamnis:
failure is genuine: https://github.com/circe/circe/actions/runs/9078292566/job/24945019131?pr=2123#step:13:2785

modules/tests/shared/src/test/scala/io/circe/DecoderSuite.scala
```diff
@@ -779,4 +780,55 @@ class DecoderSuite extends CirceMunitSuite with LargeNumberDecoderTestsMunit {
...
+    // (1 for each instance, and 1 that gets created but not called because because the last "cdr" field is missing)
+    assertEquals(counter, 1)
+  }
+
+  test("Decoder.recursive should be stack safe") {
+    // Pretty much the same as the one in the other test, we just don't need to count for this one
+    implicit def uglyListDecoder[A: Decoder]: Decoder[List[A]] =
+      Decoder.recursive[List[A]] { implicit recurse =>
+        Decoder.instance { c =>
+          (c.downField("car").as[A], c.downField("cdr").as[Option[List[A]]]).mapN(_ :: _.getOrElse(Nil))
```
johnynek: I don't think this is stack safe, and it looks like the tests fail.

So, recurse is doing to be used at `c.downField("cdr").as[Option[List[A]]]` but that's nested inside of a `mapN`. While nesteding `Defer` is stack safe in the construction you have here, composing with map2, which is what is happening here, isn't. Something like: `lazy val d = map2(a, defer(d))(_ :: _)`, `map2` would need to somehow unroll that loop. It's not easy to see how that could happen.

So, I think the stack depth will be the same depth as the object graph.

cats-parse has the same issue on parsing (which is the same basic typeclass as Decoder).

johnynek: If you ask me, we should merely add a comment to recursive and remove this test. Many structures have depth = log(size), so they are perfectly safe to parse with this. If you have a linear depth object (such as a linked list), this isn't suitable if the lists are big.

morgen-peschke: Something like this?

> Note: while a `Decoder` written using `Decoder.recursive` can prevent unneeded creation of `Decoder` instances when recursing, the resulting `Decoder` cannot be guaranteed to be stack-safe.

johnynek: yeah, something like that seems right.

modules/core/shared/src/main/scala/io/circe/Encoder.scala
```diff
@@ -23,40 +23,39 @@ import cats.data.Chain
...
+import java.time.OffsetDateTime
+import java.time.OffsetTime
+import java.time.Period
+import java.time.Year
+import java.time.YearMonth
+import java.time.ZoneId
+import java.time.ZoneOffset
+import java.time.ZonedDateTime
+import java.time.format.DateTimeFormatter
+import java.time.format.DateTimeFormatter.ISO_LOCAL_DATE_TIME
```
johnynek: personally, I don't like these import reformats. I think grouped imports make it easier to see the package level dependencies and I would rather use `{ }` to the maximum vs the minimum.

morgen-peschke: I don't have strong opinions on it, other than "let the tooling handle it" 😅 

In this case, it was the result of running `scalafixAll` and `scalafmtAll` (then reverting the changes to the files I hadn't touched).

morgen-peschke: I can revert the changes manually if you find it particularly objectionable, but it might be worth revisiting the formatting config if that is the case 🤷🏻

satorg: Also, in my experience, unfolded imports tend to trigger less conflicts on merges and rebases. I guess this could one of the reasons why it was made a default behavior for the OrganizeImports rule in scalafix.

morgen-peschke: I needed to re-run the CI, so I've rolled back the import reformat

zarthross (dismissed):

hamnis (approved):

PR #2127, opened 2023-04-18
feat: Add helpers to create nullable properties from `Option[A]`

OddKristensen:
It struck me that if `Json.fromString("foo")` is a quick and easy way to construct the JSON property for the string, then it should be just as easy to do that for an `Option[String]`. So I added a bunch of helpers that account for this.

You can argue that this generalizes to be:
```scala
final def propertyOrNull[A](value: Option[A])(f: A => Json): Json = value.fold(Json.Null)(f)
```

I did consider adding this generalization, but seemed a little too general perhaps.


modules/tests/shared/src/test/scala/io/circe/JsonSuite.scala
```diff
@@ -269,6 +269,30 @@ class JsonSuite extends CirceMunitSuite with FloatJsonTests {
     assert(result === expected && resultAlias === expected)
   }
 
+  test("fromStringOrNull return Null on None") {
+    assertEquals(Json.fromStringOrNull(None), Json.Null)
+  }
```
zarthross, line 274: Can you also add tests for the `Some` side here?  As it stands now we could accidentally always return null from `from*OrNull` and the tests would pass.

zarthross (dismissed):

zarthross (approved):

Issue #2104, opened 2023-02-27
Codec.from changes decoder behaviour


jakubdaniel:
Assuming
```scala
val e = io.circe.Encoder.instance[Option[Int]] { _ => ??? } // fake, not used in this example
val d = io.circe.Decoder.decodeOption[Int]
val c = io.circe.Codec.from(d, e)
val h = io.circe.Json.obj().hcursor
```
is
```scala
scala> d.at("x").tryDecode(h)
val res0: io.circe.Decoder.Result[Option[Int]] = Right(None)

scala> c.at("x").tryDecode(h)
val res1: io.circe.Decoder.Result[Option[Int]] = Left(DecodingFailure at .x: Missing required field)
```
expected?

I'd expect `c` to behave exactly as `d` on all inputs to any of the `Decoder` methods. But it obviously doesn't. This is because it falls back to the default implementations of `tryDecode` etc instead of properly delegating all the methods to any potential overrides already baked into `d`.

zarthross:
Looks to me like the `Codec.from` method only overrides the 2 abstract methods from Encoder and Decoder, but in reality, we override various other methods from Encoder/Decoder all the time... we need to update that `Codec.from` implementation (and any other implementations of `Codec` for that matter) so that it correctly proxies ALL methods to the correct underlying `Encoder/Decoder`

satorg:
It seems, this one was resolved via #2131 , wasn't it?

PR #2131, opened 2023-05-01
Fix 2104: Proxy remaining Codec methods

solarized-fox:
a fix wrt to #2104, I believe these are all the nonfinal methods that needed rerouting. I ran `test` and `scalastyle` in sbt, and it should be good to go


zarthross:
@solarized-fox Mind creating a unit test so we can prevent future regressions?

drewfeelsblue:
the same for `AsObject.from`
```
  val decoder = Decoder.decodeOption[Int]
  val codec   = Codec.AsObject.from[Option[Int]](decoder, null)
  val hCursor = Json.obj().hcursor

  decoder.at("x").tryDecode(hCursor) // Right(None)
  codec.at("x").tryDecode(hCursor)   // Left(DecodingFailure at .x: Missing required field)
```

`AsArray.from`?

solarized-fox:
> @solarized-fox Mind creating a unit test so we can prevent future regressions?

you mean a test to make sure `Codec.from(…).method(…) == coder.method(…)`, or a test to make sure all nonfinal methods are delegated? either way I should be able to do that, might take a hot minute though

solarized-fox:
all right, is this satisfactory as a test?

zarthross (approved):

zmccoy (approved):

PR #2138, opened 2023-05-15
Fix-2113 fix cursor`s traverse ACursor.pathToRoot

iyfedorov:
Remove redundant case in `ACursor.pathToRootlastCursorParentOrLastCursor` that skips current CursorOp parent


zarthross:
Can you add a few more tests cases?  Looks like we previously lacked tests here, and I think adding a few more deeply nested cases, and a test case with an array or two in it would help raise confidence in this fix.
Thank you!

hamnis (approved):
Looks good to me, deferring to @zarthross for merge.

zarthross (approved):

PR #2141, opened 2023-05-26
Simplify the implementation of ConfiguredCodec

cacapouh:
SSIA


modules/tests/shared/src/test/scala-3/io/circe/ConfiguredDerivesSuite.scala
```diff
@@ -420,11 +420,6 @@ class ConfiguredDerivesSuite extends CirceMunitSuite:
       List(DownField("ConfigExampleFoo"))
     )
     assert(Decoder[ConfigExampleBase].decodeJson(json) === Left(failure("unexpected fields: anotherField")))
-    assert(
-      Decoder[ConfigExampleBase].decodeAccumulating(json.hcursor) === Validated.invalidNel(
-        failure("unexpected field: anotherField")
-      )
-    )
```
zarthross: Why did you remove this test?

modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredCodec.scala
```diff
@@ -1,48 +1,13 @@
...
-          case product: Mirror.ProductOf[A] => decodeProductAccumulating(c, product.fromProduct)
-          case _: Mirror.SumOf[A]           => decodeSumAccumulating(c)
-
-  inline final def derive[A: Mirror.Of](
-    transformMemberNames: String => String = Configuration.default.transformMemberNames,
-    transformConstructorNames: String => String = Configuration.default.transformConstructorNames,
-    useDefaults: Boolean = Configuration.default.useDefaults,
-    discriminator: Option[String] = Configuration.default.discriminator,
-    strictDecoding: Boolean = Configuration.default.strictDecoding
-  ): ConfiguredCodec[A] =
```
zarthross: This cannot be removed without breaking bin-compat, can you re-add this please?

modules/tests/shared/src/test/scala-3/io/circe/ConfiguredDerivesSuite.scala
cacapouh: I apologize for the oversight. I mistakenly removed the above expression, under the false impression that `ConfiguredDecoder[ConfigExampleBase]#decodeAccumulating` would be called.

zarthross:
This appears to still be failing bin-compat, please correct any further issues, before Zach and I review.

PR #2156, opened 2023-06-11
 Fixes #2110

### LINKED ISSUE #2110: Inconsistent encoding/decoding of nested sum types
As of 0.14.5, derived encoder and decoder of a nested sum type are inconsistent. The `Decoder` can't decode a JSON encoded by the `Encoder`.

Minimal example:

```scala
import io.circe.Decoder
import io.circe.Encoder
import io.circe.parser.decode
import io.circe.syntax._

sealed trait Top

object Top {
  sealed trait Middle extends Top

  object Middle {
    case object Bottom extends Middle
  }

  implicit val decodeTop: Decoder[Top] = deriveDecoder
  implicit val encodeTop: Encoder[Top] = deriveEncoder
}

object Main {
    val s = (Top.Middle.Bottom: Top).asJson.noSpaces
    println(s)
    println(decode[Top](s))
  }
}
```

Expected output (scala:3.2.2, circe:0.14.3)

```
{"Middle":{"Bottom":{}}}
Right(Bottom)
```

Actual output (scala:3.2.2, circe:0.14.5)

```
{"Middle":{"Bottom":{}}}
Left(DecodingFailure at .Middle: type Top has no class/object/case named 'Middle'.)
```

^ In this case, the decoder can decode `{"Bottom":{}}` to `Bottom`.

---

It turned out that both of the above are also inconsistent with scala 2.13 build of circe:

Output (scala:2.13.0, circe:0.14.3/0.14.5)

```
{"Bottom":{}}
Right(Bottom)
```

So I'm not sure which encoding is correct.

MartinHH:
Restores consistent encoding/decoding of nested sum types.

Also changes the encoding to fix the inconsistency between Scala 2 and Scala 3 (as described in #2110 ).


modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredDecoder.scala
```diff
@@ -53,7 +47,14 @@ trait ConfiguredDecoder[A](using conf: Configuration) extends Decoder[A]:
...
             )
-          case Right(Some(sumTypeName)) => fromName(sumTypeName, c)
+          case Right(Some(sumTypeName)) =>
+            decodersDict
+              .get(sumTypeName)
+              .fold(
+                nameNotFound(sumTypeName, c)
+              ) { decoder =>
+                decode(decoder.asInstanceOf[Decoder[A]])(c)
+              }
```
MartinHH, line 57: FYI: this is basically the implementation of `def fromName` from previous line 35 ff.

modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredDecoder.scala
```diff
@@ -70,7 +71,11 @@ trait ConfiguredDecoder[A](using conf: Configuration) extends Decoder[A]:
                     s"expected a single key json object with one of: ${constructorNames.iterator.mkString(", ")}."
                   )
                 )
-              else fromName(sumTypeName, c.downField(sumTypeName))
+              else
+                val cursor = c.downField(sumTypeName)
+                constructorNames.indexOf(sumTypeName) match
+                  case -1    => nameNotFound(sumTypeName, cursor)
+                  case index => decode(elemDecoders(index).asInstanceOf[Decoder[A]])(cursor)
```
MartinHH, line 78: FYI: this is the implementation of `def fromName` from before https://github.com/circe/circe/pull/2098/files#diff-378e58dd77c3ec12a0e0b230ec5d55a15679e08f37f01ef83c5ee06967ca42d4 (Since that change only changed the encoding for the case where `conf.discrimniator.isDefined`, this restores the previous decoding for the `conf.discrimniator.isEmpty` case.)

MartinHH:
Failure of the CI check is not related to this change - it can be reproduced on the main branch. See #2160 2160

zarthross (approved):

zmccoy (approved):

hoangmaihuy:
@zarthross May I ask when will we have a release containing this fix? Inconsistent encode/decode in version 0.14.5 is a major blocker for our Scala 3 migration, but we also can't use 0.14.3 because that version doesn't support ConfiguredCodec derivation. Thank you a lot 😄

PR #2157, opened 2023-06-12
improve adt docs

drewfeelsblue:


docs/src/main/tut/codecs/adt.md
```diff
@@ -21,25 +21,27 @@ case class Qux(values: List[String]) extends Event
...
 object GenericDerivation {
   implicit val encodeEvent: Encoder[Event] = Encoder.instance {
-    case foo @ Foo(_) => foo.asJson
-    case bar @ Bar(_) => bar.asJson
-    case baz @ Baz(_) => baz.asJson
-    case qux @ Qux(_) => qux.asJson
+    case foo: Foo => foo.asJson
+    case bar: Bar => bar.asJson
+    case baz: Baz => baz.asJson
+    case qux: Qux => qux.asJson
```
zarthross, line 33: I like this is an improvement.

docs/src/main/tut/codecs/adt.md
```diff
@@ -21,25 +21,27 @@ case class Qux(values: List[String]) extends Event
...
   implicit val decodeEvent: Decoder[Event] =
-    List[Decoder[Event]](
+    NonEmptyList.of[Decoder[Event]](
       Decoder[Foo].widen,
       Decoder[Bar].widen,
       Decoder[Baz].widen,
       Decoder[Qux].widen
-    ).reduceLeft(_ or _)
+    ).reduceK
 }
```
zarthross: I don't think for the target audience of our documentation, that the use of `reduceK` over `reduceLeft(_ or _)` actually makes it any more clear on how we might compose decoders.

zarthross (changes requested):
We'd be happy to merge in the change around the pattern match if you can remove the other change. Thanks for the contribution.

docs/src/main/tut/codecs/adt.md
drewfeelsblue: no problems, I'll remove it. I just believed that the docs should provide with, I'd say, the most "correct" code (and here passing the function explicitly is kind of redundant)

drewfeelsblue: moreover, this code (with the `reduceLeft` on the list) can be considered as erroneous by the scapegoat's `UnsafeTraversableMethods` inspection

drewfeelsblue: removed

zarthross: Thanks for updating it.  I didn't think about the linter scapegoat, and the `UnsafeTraversableMethods`.  I think switching to a `NonEmptyList` would fix that specific issue, and showing another example with `reduceK` could be fine.  I just want to make sure our average audience understands at a glance what is going on.

zarthross (approved):

zmccoy (approved):

PR #2158, opened 2023-06-14
feat: allow to derive enumeration codec for nested sealed hierarchies

kpodsiad:
Hey folks! Current logic for deriving enum codec lacks ability to derive it for nested hierarchies. What do you think about adding it? In case answer is yes, here's the PR.

There is one caveat though, `Mirror.Of[A].ordinal` doesn't work as expected for nested hierarchies, that's why I needed to use `summonTypeTestRecursively` (credits to @odomontois and https://github.com/evolution-gaming/derivation). Here's scala-cli snippet for quick showcase what's the problem
```scala
//> using scala "3.3.0"

import scala.deriving.Mirror

sealed trait Animal
object Animal:
  sealed trait Mammal extends Animal
  case object Dog extends Mammal
  case object Cat extends Mammal
  case object Bird extends Animal

object Main extends App {
  val mirror = summon[Mirror.Of[Animal]]
  assert(mirror.ordinal(Animal.Dog) == 0)
  assert(mirror.ordinal(Animal.Cat) == 0) // <<< HERE
  assert(mirror.ordinal(Animal.Bird) == 1)
}
```

### DIFF
--- a/modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredEnumDecoder.scala
+++ b/modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredEnumDecoder.scala
@@ -4,12 +4,13 @@ import scala.deriving.Mirror
 import scala.compiletime.constValue
 import Predef.genericArrayOps
 import io.circe.{ Decoder, DecodingFailure, HCursor }
+import scala.collection.mutable.ArraySeq
 
 trait ConfiguredEnumDecoder[A] extends Decoder[A]
 object ConfiguredEnumDecoder:
   inline final def derived[A](using conf: Configuration)(using mirror: Mirror.SumOf[A]): ConfiguredEnumDecoder[A] =
-    val cases = summonSingletonCases[mirror.MirroredElemTypes, A](constValue[mirror.MirroredLabel])
-    val labels = summonLabels[mirror.MirroredElemLabels].toArray.map(conf.transformConstructorNames)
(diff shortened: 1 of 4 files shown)


kpodsiad:
@zarthross @zmccoy what do you folks think about it?

MartinHH:
Seems like their are several issues related to (scala 3) derivation for such nested "sum" hierarchies - see #2110 / #2156 and previously #2096 (all of them seemingly related to how those nested sums are "mirrored").

I guess ideally, one would take a step back and try to look at the bigger picture and try to find a better approach for all cases of scala 3 derivation for sums than fixing them one by one.

kpodsiad:
Thanks for letting me know @MartinHH. To some point I wasn't even aware of this until 2days ago we hit such errors ourselves :/ I can take a look at them and try to tackle nested sum problem, as soon at company I'm working at we might have more advanced cases. We're in the process of cross-compilation between Scala 2 & 3 and so far circe is one the biggest obstacle, I'll try to come up with some actionable actions.

MartinHH:
I had a second (and third) look myself and came to the conclusion  that the use cases and implementations of `ConfiguredDe/EnCoder` and `ConfiguredEnumDe/Encoder` are so different that one might as well go ahead and solve the "nesteds sum" problem in two different ways (i.e. keep following the route that was chosen for `ConfiguredDe/EnCoder` in #2096 and go ahead with the approach of this PR for `ConfiguredEnumDe/Encoder`). But that's just my opinion and in the end, it's up to the maintainers of this project...

hamnis:
needs conflict fixing

PR #2173, opened 2023-07-11
Add `autoRecursiveDerivation` config option

### LINKED ISSUE #2126: Support semiauto derivation without recursive derivation
In scala 2, semiauto derivation requires that any nested case classes also have instances defined explicitly, while in scala 3 derivation automatically recurses to perform derivation for the nested case classes ([the `Mirror.Of` branch here](https://github.com/circe/circe/blob/590c93959c24b7deac00e9b11ae61d0a692b6920/modules/core/shared/src/main/scala-3/io/circe/derivation/package.scala#L28-L32)), e.g.

```scala
case class Foo()
case class Bar(foo: Foo)
// scala 2 - doesn't compile because there's no Decoder[Foo]
io.circe.generic.semiauto.deriveDecoder[Bar]
// scala 3 - compiles
io.circe.Decoder.derived[Bar]
```

I understand this is convenient in many cases, but I feel like there are a number of reasons it's less then ideal -- for example, I may define a `Decoder[Foo]` with custom behavior, and then someone later on without context could delete it, see that all code still compiles, and assume that everything is fine when it's not.

Is this behavior by design? If so, would you at least consider alternate derivation methods that skip the recursion?

mrdziuban:
Resolves #2126

This adds an `autoRecursiveDerivation: Boolean` config option that controls whether derivation should auto-recurse when it encounters a nested type with a `Mirror` instance. The default behavior is unchanged -- derivation will auto-recurse. When the config option is set to `false`, auto-recursion will only happen for types with a `Mirror.Singleton` instance, i.e. `case object`s. This allows for code like this to still compile:

```scala
sealed trait Foo
case object Bar extends Foo

given conf: Configuration = Configuration.default.withoutAutoRecursiveDerivation

// Compiles even though we haven't defined an `Encoder[Bar.type]`
Encoder.derivedConfigured[Foo]
```

## Note on binary compatibility

The change to `Configuration` is currently not binary compatible -- is this an issue? If so, I think there are ways to get it to be binary compatible but they'd likely be verbose

### DIFF
--- a/modules/core/shared/src/main/scala-3/io/circe/derivation/Configuration.scala
+++ b/modules/core/shared/src/main/scala-3/io/circe/derivation/Configuration.scala
@@ -21,7 +21,8 @@ case class Configuration(
   transformConstructorNames: String => String = Predef.identity,
   useDefaults: Boolean = false,
   discriminator: Option[String] = None,
-  strictDecoding: Boolean = false
+  strictDecoding: Boolean = false,
+  autoRecursiveDerivation: Boolean = true
 ):
   def withTransformMemberNames(f: String => String): Configuration = copy(transformMemberNames = f)
   def withSnakeCaseMemberNames: Configuration = withTransformMemberNames(renaming.snakeCase)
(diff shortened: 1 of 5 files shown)


mrdziuban:
This doesn't work as intended unfortunately -- [`if (conf.autoRecursiveDerivation)`](https://github.com/circe/circe/pull/2173/files#diff-37d49aeba25ea78414564f8967c2861e222f8231dbc25fc05a272e2251a29f4bR33) would need to be an `inline if`, which would mean that users would have to define their `Configuration`s as `inline given`s.

Closing this and will follow up with a different approach.

PR #2183, opened 2023-07-31
Add `*NoAutoRecursion` variants of scala 3 derivation methods

mrdziuban:
This is an alternative approach to solving the same problem as #2173 (see https://github.com/circe/circe/pull/2173#issuecomment-1658900145 for an explanation of why it didn't work).

This adds `derivedNoAutoRecursion` and `derivedConfiguredNoAutoRecursion` methods to mirror the existing `derived` and `derivedConfigured` methods, but to perform derivation without auto-recursing for nested types that have a `Mirror` instance. As in #2173, the default behavior is unchanged -- derivation will auto-recurse -- and the `*NoAutoRecursion` variants will still auto-recurse for types with a `Mirror.Singleton` instance. As an added benefit, this is binary compatible while #2173 wasn't.


mrdziuban:
@zarthross @zmccoy do you have any thoughts or feedback on this? The auto-recursive derivation has already bitten me a couple times in my codebase, I'd love to be able to avoid it

modules/core/shared/src/main/scala-3/io/circe/derivation/package.scala
```diff
@@ -25,28 +25,58 @@ private[circe] inline final def summonLabels[T <: Tuple]: List[String] =
     case _: EmptyTuple => Nil
     case _: (t *: ts)  => constValue[t].asInstanceOf[String] :: summonLabels[ts]
 
-private[circe] inline final def summonEncoders[T <: Tuple](using Configuration): List[Encoder[_]] =
+private[circe] inline final def summonEncoders[T <: Tuple](inline autoRecurse: Boolean)(using
+  Configuration
+): List[Encoder[_]] =
```
zarthross: I have reservations about changing the signature of any package private methods, we've had them break bincompat in surprising ways that mima refuses to verify for us.  Can you find all the package private ones you've changed and add deprecated versions that call your new one? 

Once you've gotten through that, @zmccoy and I will look over it.

armanbilge: > we've had them break bincompat in surprising ways that mima refuses to verify for us

Yes, probably those methods should be marked with the `@binaryAPI` annotation currently going through SIP.

- https://github.com/scala/improvement-proposals/pull/58

mrdziuban: @zarthross addressed!

mrdziuban:
@zarthross @zmccoy do you have any feedback on this?

modules/core/shared/src/main/scala-3/io/circe/derivation/package.scala
```diff
@@ -25,28 +25,70 @@ private[circe] inline final def summonLabels[T <: Tuple]: List[String] =
...
     case decodeA: Decoder[A] => decodeA
     case _: Mirror.Of[A]     => ConfiguredDecoder.derived[A]
   }
 
+private[circe] inline final def summonDecoderNoAutoRecurse[A](using conf: Configuration): Decoder[A] =
+  summonFrom {
+    case decodeA: Decoder[A] => decodeA
+    case m: Mirror.Of[A] =>
+      inline m match {
+        case _: Mirror.Singleton => ConfiguredDecoder.derivedNoAutoRecursion[A]
```
Lasering: Isn't Mirror.Singleton only available for case objects?

mrdziuban: Yeah, this is intentional. From the PR description:

> the *NoAutoRecursion variants will still auto-recurse for types with a Mirror.Singleton instance

and from the description in #2173 

> auto-recursion will only happen for types with a `Mirror.Singleton` instance, i.e. `case object`s. This allows for code like this to still compile:

```scala
sealed trait Foo
case object Bar extends Foo

given conf: Configuration = Configuration.default.withoutAutoRecursiveDerivation

// Compiles even though we haven't defined an `Encoder[Bar.type]`
Encoder.derivedConfigured[Foo]
```

Lasering: Why not make it more general? By default auto derive encoders/decoders for nested types if they are not available. Using withoutAutoRecursiveDerivation fail if a encoder/decoder is not found and just do auto derivation for the "first level".

mrdziuban: That's exactly what's implemented here. The default `Encoder.derived` and `Decoder.derived` methods will still auto derive nested types if they're not available. The new `Encoder.derivedNoAutoRecursion` and `Decoder.derivedNoAutoRecursion` methods will fail if an `Encoder`/`Decoder` for a nested type is not found. The only exception for the `*NoAutoRecursion` variants is for `case object`s, that's what this `case` statement handles

zarthross: Adding some tests showing what does and does not auto-recurse (including compiles and does not compile tests) would help make this more clear, and prevent future regressions.

mrdziuban: Good idea @zarthross, I just added a test suite to demonstrate what compiles and what doesn't

mrdziuban:
@zarthross checking in again here, I added the tests you were hoping to see. Do you have any additional feedback?

modules/tests/shared/src/test/scala-3/io/circe/DerivedNoAutoRecursionSuite.scala
```diff
@@ -0,0 +1,132 @@
...
+  checkAll("Codec[Foo]", CodecTests[Foo].codec)
+  checkAll("Codec[Box[Foo]]", CodecTests[Box[Foo]].codec)
+  checkAll("Codec[Bar]", CodecTests[Bar].codec)
+  checkAll("Codec[Box[Bar]]", CodecTests[Box[Bar]].codec)
+  checkAll("Codec[Adt1]", CodecTests[Adt1].codec)
+  checkAll("Codec[Adt2]", CodecTests[Adt2].codec)
+
+  test("Nested case classes cannot be derived") {
+    assert(!typeChecks("Decoder.derivedNoAutoRecursion[Quux]"))
+    assert(!typeChecks("Encoder.AsObject.derivedNoAutoRecursion[Quux]"))
```
zarthross: Why did you use `typeChecks` instead of `compileErrors`?

modules/tests/shared/src/test/scala-3/io/circe/DerivedNoAutoRecursionSuite.scala
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  test("Nested ADTs cannot be derived") {
+    assert(!typeChecks("Decoder.derivedNoAutoRecursion[Adt3]"))
+    assert(!typeChecks("Encoder.AsObject.derivedNoAutoRecursion[Adt3]"))
+  }
```
zarthross, line 159: Hmm... I actually think the fact this test passes is a bug.  Circe Generic and Generic Extras has no problems deriving the children (and grandchildren) of a trait, and we should allow that by default here as well. (Example [here](https://scastie.scala-lang.org/Tw91DCsxSkKBOLHqoliggQ))

modules/tests/shared/src/test/scala-3/io/circe/DerivedNoAutoRecursionSuite.scala
mrdziuban: I didn't know about `compileErrors`, updated to use it

modules/tests/shared/src/test/scala-3/io/circe/DerivedNoAutoRecursionSuite.scala
mrdziuban: Interesting. This goes directly against what "semiauto" means to me, which is that any subtypes of a trait must already have an instance in scope. `case object`s are an exception in my mind because deriving for them will always produce the same instance -- one that (de)serializes to `{}`.

I'm basing my desired behavior off of how [circe-derivation](https://github.com/circe/circe-derivation) worked, rather than circe generic. The latest release doesn't have my fix (https://github.com/circe/circe-derivation/pull/303) included, but if it did then the behavior in the scastie you shared would only work if there was a `Codec[Foo]` already defined.

zarthross: I think its more convenient, and understood by more users to work the way circe-generic and circe-generic-extras behaves here, especially since circe-generic is part of circe's main repo.

mrdziuban: Okay, I implemented this. Now instances for any child (or grand...child) of a `sealed trait` will be automatically derived, not just `case object`s. I added a test to verify that this works, and to verify that this still can't be derived for because there are no instances for `Nested`, which are needed to derive for `Class1`

```scala
sealed trait Adt5
object Adt5 {
  case class Nested()
  case class Class1(nested: Nested) extends Adt5
  case object Object1 extends Adt5
}
```

modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredEncoder.scala
```diff
@@ -54,10 +54,18 @@ trait ConfiguredEncoder[A](using conf: Configuration) extends Encoder.AsObject[A
...
+  ): ConfiguredEncoder[A] =
     new ConfiguredEncoder[A] with SumOrProduct:
       lazy val elemLabels: List[String] = summonLabels[mirror.MirroredElemLabels]
-      lazy val elemEncoders: List[Encoder[?]] = summonEncoders[mirror.MirroredElemTypes]
+      lazy val elemEncoders: List[Encoder[?]] =
+        summonEncoders[mirror.MirroredElemTypes](
+          autoRecurse,
+          inline mirror match
+            case _: Mirror.ProductOf[A] => false
+            case _: Mirror.SumOf[A]     => true
```
mrdziuban, line 67: I was hoping to share this logic with `isSum` below by making `isSum` an `inline val`, e.g.

```scala
inline val isSum =
  inline mirror match
    case _: Mirror.ProductOf[A] => false
    case _: Mirror.SumOf[A]     => true

lazy val elemEncoders: List[Encoder[?]] = summonEncoders[mirror.MirroredElemTypes](autoRecurse, isSum)
```

but then I get this error:

```scala
 79 |    given encoder: Encoder.AsObject[Adt1] = Encoder.AsObject.derivedNoAutoRecursion
    |                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |Cannot reduce `inline if` because its condition is not a constant value: this.isSum
```

I'm sure there's something I'm misunderstanding about `inline val`s, just wanted to call it out

modules/tests/shared/src/test/scala-3/io/circe/DerivedNoAutoRecursionSuite.scala
Lasering: I think semiauto should derive encoder/decoders/codecs for all the sum types of a type. And auto derivation should do it for sum and product types.

hamnis:
I think this should be the default behaviour and merits a a new major version.
Once 0.15 development starts, this should be changed to `.derived`

I have found it very surprising that `.derived` autoderives anything beyond one level.

mrdziuban:
@zarthross @zmccoy there seem to be a lot of different opinions on what the default behavior should be for semiauto derivation, so I created [a poll](https://forms.gle/ocQ35aL7a1Z6CCDc9) to try to gather more feedback from users so we can make an informed decision on the best path forward.

## Questions

There were 3 questions:

1. How should semiauto derivation work by default for case classes?
    1. It should fail if any case class field does not have a Decoder/Encoder, e.g. https://gist.github.com/mrdziuban/8bf7f99c362ff8e89c4bc8535b952442
    2. It should automatically derive a Decoder/Encoder for case class fields when possible, e.g. https://gist.github.com/mrdziuban/4921b51180ba20dd0c8789861f20705b
2. How should semiauto derivation work by default for sealed traits?
    1. It should fail if any subtype does not have a Decoder/Encoder, e.g. https://gist.github.com/mrdziuban/936a8d2787c0ccce369af9cc26ba4229
    2. It should automatically derive a Decoder/Encoder for case objects, but not for case classes or sub sealed traits, e.g. https://gist.github.com/mrdziuban/10648b5bed11ca292366f7a498df44b7
    3. It should automatically derive a Decoder/Encoder for case objects and sub sealed traits, but not for case classes, e.g. https://gist.github.com/mrdziuban/5670257e0ec111e517f972a9c8819527
    4. It should automatically derive a Decoder/Encoder for all subtypes, e.g. https://gist.github.com/mrdziuban/02b94ce717eb8a9a9cce599a442d44a4
    5. Other (users could enter a different option)
4. Any other thoughts?

## Current implementation

As of now, the `derived`/`derivedConfigured` methods in circe implement option 1.ii for case classes and option 2.iv for sealed traits. The `*NoAutoRecursion` methods in this pull request implement option 1.i for case classes and option 2.iv for sealed traits.

## Responses

The poll got 107 responses, below is a summary for each question. You can view all responses here: https://docs.google.com/spreadsheets/d/1L4En_2HnJFFPP8haDTsLAr06xDLsZbW54Nw6JoNNBj8/edit?usp=sharing

### case classes

Users think that semiauto derivation should not automatically derive for case class fields, and instead should fail if any field does not have a `Decoder`/`Encoder`.

- 70/107 (65.4%) of users chose option 1.i
- 37/107 (34.6%) of users chose option 1.ii

### sealed traits

Users were more split on this, but it was pretty even between options 2.i (not deriving for any subtypes) and 2.iv (deriving for all subtypes). Note that one user left this blank, so there were only 106 responses.

- 46/106 (43.4%) of users chose option 2.iv
- 43/106 (40.6%) of users chose option 2.i
- 9/106 (8.5%) of users chose option 2.iii
- 4/106 (3.8%) of users chose option 2.ii
- 5 users chose "Other", and 4 of them entered their own option
    - derive for everything that extends the sealed trait but not sub classes or sealed traits
    - It should automatically derive a Decoder/Encoder for case objects and for case classes (with all parameters already having a Decoder/Encoder instance), but not for sub sealed traits
    - It should fail if any subtype does not have a Decoder/Encoder, and it should demand an explicit config object determining whether there is/is not a discriminator and how to derive that discriminator
    - Not sure about this one, but the current behaviour for case classes in question 1 is really annoying

### Any other thoughts?

I'm including all responses (collapsed) below, even if they go off topic or suggest changes that would massively overhaul circe's architecture (which I'm not advocating for by any means).

<details>
  <summary>expand responses</summary>

- Circe is most useful when it has a similar user experience to Jackson - i.e. that it can handle virtually anything and the only time explicit code is necessary is to influence the output JSON form. Adherence to a specific JSON spec should be enforced by unit tests, not compile-time limitations of the encoder.
- Full auto makes you subservient to your json schema
- Failure to compile on missing encoders is the only sane way to do semi-automatic derivation.  Anything else is a bug waiting to happen because something should have some custom encoding. It is also much more explicit.
- I'd like to see `Encoder.AsObject` removed as well
- Semi-auto derivation requires typeclass evidence exists for all nested types, having it automatically derive evidence doesn't make sense. Auto-derivation is wasteful and adds to compile times, which are already slow for Scala on sbt.
- never auto-derive
- Great initiative! I always wondered why are members and subtypes not derived when all I had to do was just call .derive[] on them, without custom code.
- IMO, every library should use "sensible defaults", with minimum configuration/parameters.  
- Ask yourself: what would an average user EXPECT to happen when calling a piece of code.
    
    Right now Circe requires lots of boilerplate and, in my opinion, confusing configuration.
    I would get rid of the implicit configuration COMPLETELY.
    Scala has support for wild identifiers (with backticks) like `this: crazyAPI _: has spaces`: String
    When using (especially lurking, from where??, implicit) configuration, and automapping camelCase/under_score, I cant Ctrl+F that identifier.
    
    Re polymorphic stuff, I prefer to have a default of field `@type`.
    And I'd use case-class/object/enum-case simple name.
    If you want override, put an annotation like @discriminator("my_type").
    
    This not only simplifies using the library, but also implementation is simpler, less stuff to support, less bugs, less security exploits etc.
    
    I've implemented a library on these principles.
    https://github.com/sake92/tupson
- I answered "How should semiauto derivation work by default for sealed traits?" with "It should automatically derive a Decoder/Encoder for all subtypes", because that would allow for elegant derivation for `enum`s that are new in Scala 3.
    "It should fail if any subtype does not have a Decoder/Encoder" would be too verbose and inelegant with enums.
- Semiauto decoding should be "Fail safe" by default. It should not be possible to (de)serialize a type that hasn't *very* explicitly been opted in. 
    
    Codecs for sealed-trait variants should be 'private' by default - only visible to the parent classes codec. 
    
    It should  *not* be possible to accidentally use a child class codec directly, since this would skip the discriminator. 
    
    (Since the type being serialized might be inferred, it's possible a change in one location could inadvertently result in a different type being inferred in a completely different location, resulting in a child-class codec being used without realizing. This can be extremely dangerous, and difficult to detect and diagnose)
    
    Anything less than this, and you'll end up on the CVSS list next to Jackson Databind.
- I have a clear personal preference on this, but I do acknowledge that those who have other preferences also have valid reasons (e.g. fear of accidentally deriving stack-unsafe codecs for recursive structures).
    
    Hence, I think library authors should allow library users to decide this, i.e. they should allow to configure the behavior of Decoder.derived / Encoder.AsObject.derived.
- I would also take Scala 3 enums into account - I replaced my sealed traits with them. And there should be a distinction between 'simple' once that are just Strings and the other once that need an object with discriminator.
- Full auto derivation is dangerous and i consider it unsuitable for production code. Semi auto derivation is only acceptable for production code insofar as it does not auto-derive anything (such as case class fields). Some of the options offered in this poll would make semi auto derivation unsafe by this definition.
- An important thing for me in semiauto derivation is a clear error in case something fails. If possible, the error should explicitly say which field/subtype is the source of the issue.
- Would be nice to have consistent behavior across 
- if there's different use cases we could have deriveDecoder.withAutoFields(true) or something, unless that would confuse matters.
    The benefit of semiauto for me is to make sure if someone adds a field it doesn't just get randomly serialised if not appropriate.  Hence me being so strict.
- With the strict (shallow) variant, I can be sure that my special encoder for a case class field is used, and not an automatically derived one.
    But a separate deep automatic derivation could be useful, too.
    Thanks for asking!
- Didn't realize this was missing from circe for Scala 3.  Hopefully in place before we migrate.

</details>

## Conclusions

Given the results, I think it makes sense to change the default behavior of `case class` derivation to not auto-recurse, in a new major version as @hamnis suggested.

`sealed trait`s are more up in the air. I feel fairly strongly about not requiring explicit derivation for `case object`s, as it would introduce a lot more boilerplate for (IMO) little gain -- I'm wondering if users who took the poll didn't fully understand that consequence of option 2.i. My personal preference is for option 2.iii, but I would also be okay with 2.iv.

Lasering:
My two cents:

1.ii  + 2.iv = auto derivation. 
2.i This looks like no derivation at all.
2.ii and 2.iii are tricky. In the derivation code how do you distinguish from case objects, case classes, and sub sealed traits?

How does the generic-extras module handles things in Scala2?

mrdziuban:
> In the derivation code how do you distinguish from case objects, case classes, and sub sealed traits?

At least in scala 3, I think this could be done with an `inline match` on a `Mirror` instance for each subtype. A `Mirror.Singleton` would be a `case object`, `Mirror.Product` would be a `case class`, and `Mirror.Sum` would be a sub sealed trait.

That said, @armanbilge asked about my preference for 2.iii [on Discord](https://discord.com/channels/632277896739946517/877550996191084554/1181288679822610523), and I realized I don't have any strong reasoning for it other than the fact that it's just what I'm familiar with, since I use circe-derivation and that's how it works. I'm fully on board with 2.iv assuming it follows the pattern he described:

> I think my preference is for 2.iv, subject to the same rule 1.i that case classes don't just auto-derive all the way down

armanbilge:
> I think it makes sense to change the default behavior of `case class` derivation to not auto-recurse

👍 I also want to point out that [this was likely the intended use of `derives` by the Scala 3 team](https://discord.com/channels/632150470000902164/875868146949554207/1113598380824866859), not the current implementation.

> My personal preference is for option 2.iii, but I would also be okay with 2.iv.

I prefer 2.iv. where `case class`es do not auto-recurse as per 1.i.

---

I haven't had a chance to study to the binary implications of changing the default behavior of `derives`, but I suspect it could be possible to roll the changes out without breaking binary-compatibility. Assuming that this is true ...

> in a new major version as @hamnis suggested.

Honestly this is the hardest question of all. This is definitely a major source-breaking change, and bumping to a new major version would properly express that to users. That seems good.

However, at the same time, tooling would treat this release as binary-incompatible. So users must either silence version conflicts in their build and/or we force lots of libraries to upgrade and release against the new version, when it's not really necessary. That seems bad. Especially because upgrading a library may require non-trivial work to deal with all the source-breaking changes, when the current release can still be used perfectly fine.

What exactly is the migration path? Can we scalafix users to the new syntax for the old behavior?

Lasering:
Enums should also be taken into consideration. If I'm not mistaken:

The `A` here will be a `Mirror.Singleton`:
```scala
enum Foo:
  case A
```

The `Mix` here will be listed in the `Mirror.Sum.MirroredElemTypes` for the type `Color`, and there will be a `Mirror.Product` for `Mix`:
```scala
enum Color(val rgb: Int):
  case Mix(mix: Int) extends Color(mix)
```

Also enums are [desugared](https://docs.scala-lang.org/scala3/reference/enums/desugarEnums.html#) to sealed abstract classes (with case classes) so in essence a sealed trait hierarchy, or ,in simple enums like Foo, to fields of the Foo companion class. The reason why this is important is that the way enums and sealed trait hierarchies are derived should be consistent.

At least for me, choosing how many levels (in the sum hierarchy) to derive for is a much more tricky subject. And to make things more confusing many people see the `Mirror.Product.MirroredElemTypes` as the last level in this hierarchy.

For me it makes sense that both enum `Foo` and enum `Color` are derived successfully using semiauto derivation. Which, to make things consistent, implies sealed trait hierarchies with one level should also be successfully derived with semiauto derivation. More concretely this should work (notice circe already defines a decoder for all the types used, if that was not the case the derivation should fail):
```scala
sealed trait Adt
case class AdtFoo(i: Int, s: String = "abc") extends Adt
case class AdtBar(xs: List[Boolean]) extends Adt
case object AdtQux extends Adt

Decoder.derived[Adt]
```

Now for cases like these is where I find it to be a tricky subject.
```scala
sealed trait NestedAdt
sealed trait NestedAdt1 extends NestedAdt
sealed trait NestedAdt2 extends NestedAdt
final case class NestedAdtFoo(i: Int, s: String = "abc") extends NestedAdt1 with NestedAdt2
case class NestedAdtBar(xs: List[Boolean]) extends NestedAdt
case object NestedAdtQux extends NestedAdt1
```


It seems like the main pain point is that the current derivation derives encoders/decoders for the `Mirror.Product.MirroredElemTypes`, when that is something that only auto derivation should do.

Historical note: when I implemented https://github.com/circe/circe/pull/1800 I hadn't a clear distinction in my head from auto vs semiauto derivation (especially regarding sum vs products types). My implementation was a suggestion on how to go about it, since most people that looked at the PR had no problems with it, that implementation ended up being merged. However knowing what I know now, semiauto derivation would only derive encoders/decoders for the sum types, and auto derivation would be semiauto + derivation for product types.

mrdziuban:
@zarthross @zmccoy are you open to changing the behavior of semiauto derivation to follow the pattern described in 2.iv of my notes above, https://github.com/circe/circe/pull/2183#issuecomment-1839047060?

If so, I have a much simpler branch that does so here: https://github.com/mblink/circe/compare/series/0.14.x...less-auto-recursive-derivation. AFAICT this would be a binary compatible, but not source compatible, change.

hamnis:
> @zarthross @zmccoy are you open to changing the behavior of semiauto derivation to follow the pattern described in 2.iv of my notes above, [#2183 (comment)](https://github.com/circe/circe/pull/2183#issuecomment-1839047060)?
> 

> If so, I have a much simpler branch that does so here: [mblink/circe@series/0.14.x...less-auto-recursive-derivation](https://github.com/mblink/circe/compare/series/0.14.x...less-auto-recursive-derivation). AFAICT this would be a binary compatible, but not source compatible, change.

Hi, new maintainer here.

As long as we do not change the serialization format, I would be open for this.
I will defer to @zarthross here though.
The scalafix idea to mitigate the source incompatibility would be interesting to explore, but I do not have the time to do that. 

Regardless, I will merge #2230 today, so we should build on that.

zarthross:
Sorry for the late response... personal life has gotten in the way of open source recently, but hopefully I'm back to reviewing items more frequently.

As for what Semi-Auto means, I agree with the polling consensus, and I may have mis-communicated previously.

Option 1.I  and 2.IV are Semi-auto and therefor what `derives` should be doing.  Option 1.II is Full-Auto, or at least how we used the terms in scala 2's generic and generic-extras.  

I want to add the extra caveat that 2.IV should FULLY recurse a sealed trait hierarchy no matter how deeply nested.

For clarity:
```
sealed trait Animal
final case class Fox extends Animal
sealed trait Dog
final case class Corgi(color: String) extends Dog

// This should succeed.
io.circe.Decoder.derived[Animal]
```

zarthross:
As far as versioning goes, if we can keep the PR bin-compat and do a 0.14 release we should be fine.  Yes its source incompat, but we should just make a big deal in the release notes about it and give clear instructions on how to fix.  

I'd rather not wait for 0.15 for something like this, or release an early 0.15 just for this, because it would ripple through the ecosystem.

If someone has time to write a scalafix that would be cool, but not sure the best way to do that in this case.

mrdziuban:
> I want to add the extra caveat that 2.IV should FULLY recurse a sealed trait hierarchy no matter how deeply nested.

Agreed, and the branch I linked above handles that. I won’t be able to do it this week, but early next week I can update it and open a PR

mrdziuban:
@hamnis @zarthross I just opened #2258 with the changes we've discussed. If that looks good, feel free to close this at any point

PR #2187, opened 2023-08-23
Improve derivation support for recursive ADTs

### LINKED ISSUE #2101: [Bug] Recursive ADTs fail to Derive
This simple recursive data structure fails to derive an Encoder:
```
sealed trait Tree; 
case class Branch(l: Tree, r: Tree) extends Tree; 
case object Leaf extends Tree;
```

With this error:
```
Welcome to Scala 3.2.2 (17.0.6, Java OpenJDK 64-Bit Server VM).
Type in expressions for evaluation. Or try :help.
                                                                                
scala> import _root_.io.circe._
                                                                                
scala> sealed trait Tree; case class Branch(l: Tree, r: Tree) extends Tree; case object Leaf extends Tree;;Encoder.AsObject.derived[Tree]
-- Error: ----------------------------------------------------------------------
  1 |sealed trait Tree; case class Branch(l: Tree, r: Tree) extends Tree; case object Leaf extends Tree;;Encoder.AsObject.derived[Tree]
    |                                                                                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |                   Maximal number of successive inlines (32) exceeded,
    |                   Maybe this is caused by a recursive inline method?
    |                   You can use -Xmax-inlines to change the limit.
    |---------------------------------------------------------------------------
    |Inline stack trace
    |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
    |This location contains code that was inlined from rs$line$2:1
 20 |    case _: Mirror.Of[A]     => ConfiguredEncoder.derived[A]
    |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
    |This location contains code that was inlined from rs$line$2:1
 15 |    case _: (t *: ts)  => summonEncoder[t] :: summonEncoders[ts]
    |                          ^^^^^^^^^^^^^^^^
    |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
    |This location contains code that was inlined from rs$line$2:1
 32 |      lazy val elemEncoders: List[Encoder[?]] = summonEncoders[mirror.MirroredElemTypes]
    |                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
    |This location contains code that was inlined from rs$line$2:1
 20 |    case _: Mirror.Of[A]     => ConfiguredEncoder.derived[A]
    |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
    |This location contains code that was inlined from rs$line$2:1
 15 |    case _: (t *: ts)  => summonEncoder[t] :: summonEncoders[ts]
    |                          ^^^^^^^^^^^^^^^^
    |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
    |This location contains code that was inlined from rs$line$2:1
 32 |      lazy val elemEncoders: List[Encoder[?]] = summonEncoders[mirror.MirroredElemTypes]
    |                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
    |This location contains code that was inlined from rs$line$2:1
 20 |    case _: Mirror.Of[A]     => ConfiguredEncoder.derived[A]
    |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
    |This location contains code that was inlined from rs$line$2:1
 15 |    case _: (t *: ts)  => summonEncoder[t] :: summonEncoders[ts]
    |                          ^^^^^^^^^^^^^^^^
    |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
    |This location contains code that was inlined from rs$line$2:1
 32 |      lazy val elemEncoders: List[Encoder[?]] = summonEncoders[mirror.MirroredElemTypes]
    |                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
.........
```

MartinHH:
Fixes #2101

### DIFF
--- a/modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredDecoder.scala
+++ b/modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredDecoder.scala
@@ -192,25 +192,27 @@ object ConfiguredDecoder:
   inline final def derived[A](using conf: Configuration)(using
     inline mirror: Mirror.Of[A]
   ): ConfiguredDecoder[A] =
-    new ConfiguredDecoder[A] with SumOrProduct:
-      val name = constValue[mirror.MirroredLabel]
-      lazy val elemLabels: List[String] = summonLabels[mirror.MirroredElemLabels]
-      lazy val elemDecoders: List[Decoder[?]] = summonDecoders[mirror.MirroredElemTypes]
-      lazy val elemDefaults: Default[A] = Predef.summon[Default[A]]
+    given result: ConfiguredDecoder[A] =
+      new ConfiguredDecoder[A] with SumOrProduct:
+        val name = constValue[mirror.MirroredLabel]
+        lazy val elemLabels: List[String] = summonLabels[mirror.MirroredElemLabels]
+        lazy val elemDecoders: List[Decoder[?]] = summonDecoders[mirror.MirroredElemTypes]
+        lazy val elemDefaults: Default[A] = Predef.summon[Default[A]]
 
-      lazy val isSum: Boolean =
-        inline mirror match
-          case _: Mirror.ProductOf[A] => false
-          case _: Mirror.SumOf[A]     => true
+        lazy val isSum: Boolean =
+          inline mirror match
+            case _: Mirror.ProductOf[A] => false
+            case _: Mirror.SumOf[A]     => true
 
-      final def apply(c: HCursor): Decoder.Result[A] =
-        inline mirror match
-          case product: Mirror.ProductOf[A] => decodeProduct(c, product.fromProduct)
-          case _: Mirror.SumOf[A]           => decodeSum(c)
-      final override def decodeAccumulating(c: HCursor): Decoder.AccumulatingResult[A] =
-        inline mirror match
-          case product: Mirror.ProductOf[A] => decodeProductAccumulating(c, product.fromProduct)
-          case _: Mirror.SumOf[A]           => decodeSumAccumulating(c)
+        final def apply(c: HCursor): Decoder.Result[A] =
+          inline mirror match
+            case product: Mirror.ProductOf[A] => decodeProduct(c, product.fromProduct)
+            case _: Mirror.SumOf[A]           => decodeSum(c)
+        final override def decodeAccumulating(c: HCursor): Decoder.AccumulatingResult[A] =
+          inline mirror match
+            case product: Mirror.ProductOf[A] => decodeProductAccumulating(c, product.fromProduct)
+            case _: Mirror.SumOf[A]           => decodeSumAccumulating(c)
+    result
 
   inline final def derive[A: Mirror.Of](
     transformMemberNames: String => String = Configuration.default.transformMemberNames,
--- a/modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredEncoder.scala
+++ b/modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredEncoder.scala
@@ -55,19 +55,21 @@ trait ConfiguredEncoder[A](using conf: Configuration) extends Encoder.AsObject[A
 
 object ConfiguredEncoder:
   inline final def derived[A](using conf: Configuration)(using inline mirror: Mirror.Of[A]): ConfiguredEncoder[A] =
-    new ConfiguredEncoder[A] with SumOrProduct:
-      lazy val elemLabels: List[String] = summonLabels[mirror.MirroredElemLabels]
-      lazy val elemEncoders: List[Encoder[?]] = summonEncoders[mirror.MirroredElemTypes]
+    given result: ConfiguredEncoder[A] =
+      new ConfiguredEncoder[A] with SumOrProduct:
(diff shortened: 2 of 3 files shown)


modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredDecoder.scala
```diff
@@ -192,25 +192,27 @@ object ConfiguredDecoder:
...
-
-      final def apply(c: HCursor): Decoder.Result[A] =
-        inline mirror match
-          case product: Mirror.ProductOf[A] => decodeProduct(c, product.fromProduct)
-          case _: Mirror.SumOf[A]           => decodeSum(c)
-      final override def decodeAccumulating(c: HCursor): Decoder.AccumulatingResult[A] =
-        inline mirror match
-          case product: Mirror.ProductOf[A] => decodeProductAccumulating(c, product.fromProduct)
-          case _: Mirror.SumOf[A]           => decodeSumAccumulating(c)
+    given result: ConfiguredDecoder[A] =
```
MartinHH, line 195: FYI: note that I only added this line and line 215 - anything in between is just an additional layer of indentation.

modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredEncoder.scala
```diff
@@ -55,19 +55,21 @@ trait ConfiguredEncoder[A](using conf: Configuration) extends Encoder.AsObject[A
 
 object ConfiguredEncoder:
   inline final def derived[A](using conf: Configuration)(using inline mirror: Mirror.Of[A]): ConfiguredEncoder[A] =
-    new ConfiguredEncoder[A] with SumOrProduct:
-      lazy val elemLabels: List[String] = summonLabels[mirror.MirroredElemLabels]
-      lazy val elemEncoders: List[Encoder[?]] = summonEncoders[mirror.MirroredElemTypes]
+    given result: ConfiguredEncoder[A] =
```
MartinHH, line 58: FYI: note that I only added this line and line 72 - anything in between is just an additional layer of indentation.

zarthross (dismissed):

zmccoy (dismissed):

zarthross:
@MartinHH, Looks like even if we get the compiler to generate these codecs at compile time, we will have a runtime issue with trees of sufficient size blowing up the stack.  I pushed some test changes to limit the test size so we can address this later, but on my machine i could upwards of 10,000 depth fine... in CI that number is much smaller, making me somewhat concerned about the real world limits.  

As it stands right now, users will be forced to do something other than derivation, which i think is better than merging this right now and possibly getting a stack overflow at runtime.

If we still wanted to get this in, we need to change the way the encoder/decoder works to be more like Defer/Fix https://github.com/circe/circe/pull/2123 or some other trick to trampoline on the stack rather than blow it up.

MartinHH:
> As it stands right now, users will be forced to do something other than derivation, which i think is better than merging this right now and possibly getting a stack overflow at runtime.

Not trying to talk you into merging this (I really don't mind as I wasn't planning to use this), but it might be worth pointing out that the following already compiles (resulting in a similarly unsafe codec):

```
sealed trait Tree derives Codec.AsObject
case class Branch(l: Tree, r: Tree) extends Tree
case object Leaf extends Tree
```
This as well:
```
sealed trait Tree
case class Branch(l: Tree, r: Tree) extends Tree
case object Leaf extends Tree
given Codec[Tree] = Codec.AsObject.derived[Tree]
```

Maurycyt:
On the one hand, I cannot wait for this aspect of circe in Scala 3 to be improved (and I am saddened by the lack of activity for almost 2 months), but I think this is the right place to point out a weird peculiarity which may be worth considering before merging anything.

@MartinHH mentioned, that derivation for his `Tree` structure compiles. While Martin is right, that is the following compiles:
```scala
import io.circe.Codec

sealed trait Tree
case class Branch(l: Tree, r: Tree) extends Tree
case object Leaf extends Tree
given Codec[Tree] = Codec.AsObject.derived[Tree]
```
the following doesn't:
```scala
import io.circe.Codec

sealed trait Tree
case class Branch(l: Tree, r: Tree) extends Tree
case object Leaf extends Tree
case class TreeWrapper(tree: Tree)
given Codec[TreeWrapper] = Codec.AsObject.derived[TreeWrapper]
```
The latter piece of code fails to compile with the infamous `Maximal number of successive inlines (32) exceeded` error. And the only thing that changed was moving from deriving for a recursive type to deriving for a type which contained a recursive type.

Issue #1965, opened 2022-05-25
Scala 3 `derives Encoder.AsObject` is broken with primitive type parameters


armanbilge:
This program compiles:

```scala
//> using scala "3.1.2"
//> using lib "io.circe::circe-core::0.14.2"

import io.circe.*

case class Foo[A](a: A) derives Encoder.AsObject
case class Bar(i: Int) derives Encoder.AsObject

@main def main =
  println(Encoder[Foo[Bar]].apply(Foo(Bar(42))))
```
and prints:
```
{
  "a" : {
    "i" : 42
  }
}
```

However, if you change the `main` function to
```scala
@main def main =
   println(Encoder[Foo[Int]].apply(Foo(42)))
```

Then we get this compile error:

```
[error] ./bug.scala:10:28: no implicit argument of type io.circe.Encoder[Foo[Int]] was found for parameter instance of method apply in object Encoder.
[error] I found:
[error] 
[error]     Foo.derived$AsObject[Int](
[error]       io.circe.Encoder.AsObject.importedAsObjectEncoder[Int](
[error]         /* missing */
[error]           summon[io.circe.export.Exported[io.circe.Encoder.AsObject[Int]]]
[error]       )
[error]     )
[error] 
[error] But no implicit values were found that match type io.circe.export.Exported[io.circe.Encoder.AsObject[Int]].
[error] 
[error] One of the following imports might make progress towards fixing the problem:
[error] 
[error]   import io.circe.Encoder.AsArray.importedAsArrayEncoder
[error]   import io.circe.Encoder.AsObject.importedAsObjectEncoder
[error]   import io.circe.Encoder.AsRoot.importedAsRootEncoder
[error] 
[error]   println(Encoder[Foo[Int]].apply(Foo(42)))
[error]                            ^
```

djx314:
Now it seems that case class with type parameters should use
```scala
given [T](using Encoder[T]): Encoder.AsObject[Foo[T]] = deriveEncoder
given [T](using Decoder[T]): Decoder[Foo[T]]              = deriveDecoder
```
in companion.
And case class without type parameters can use derives.

djx314:
I find two more problem.
```scala
given [T](using Encoder[T]): Encoder.AsObject[DesuResult[T]] = Encoder.AsObject.derived
given [T](using Decoder[T]): Decoder[DesuResult[T]]          = Decoder.derived
```
will compile success.
And
```scala
val f = Foo(42) // first set to a var
f.asJson
```
[...]

PR #2206, opened 2023-10-19
Add scala 3 auto derivation

Lasering:
Adds the missing automatic derivation in Scala 3.

When no Configuration is in the implicit scope `Configuration.default` will be used, aka, its the same as doing `derives Encoder.AsObject`:
```scala
import io.circe.derivation.auto.given
import io.circe.syntax.given
new Foo(5, "asd").asJson.noSpaces // prints: {"a":5,"b":"asd"}
```

If a Configuration is in the implicit scope then it will be used, aka, its the same as doing `derives ConfiguredEncoder`:
```scala
import io.circe.derivation.auto.given
import io.circe.syntax.given
given Configuration = Configuration.default.withTransformMemberNames(_.toUpperCase())
new Foo(5, "asd").asJson.noSpaces // prints: {"A":5,"B":"asd"}
```

Decoders and Codecs are also automatically derived.

If this is something that would be accepted I can add some tests.

Also this is a precursor to a PR I intend to do documenting the Scala 3 derivation.


zarthross:
@Lasering I think this is something we could add in, if you don't mind adding the tests that would be great!

Lasering:
Nice! This PR should be built on top of https://github.com/circe/circe/pull/2183 or at least using part of the same machinery. Maybe it makes more sense to have a bigger PR that fixes all the derivation problems:

- Auto derivation is recursive
- Non-auto derivation is not recursive.
- Both share mostly the same implementation with a boolean to toggle the recursivity. Like in https://github.com/circe/circe/pull/2183
- Move the `derived` methods from `Encoder.AsObject` to `Encoder`, and from `Codec.AsObject` to `Codec`. This solves https://github.com/circe/circe/issues/1965.

MartinHH:
Are you aware of the existing scala 3 auto-derivation in the circe-generic module: https://github.com/circe/circe/blob/series/0.14.x/modules/generic/shared/src/main/scala-3/io/circe/generic/auto.scala ?

I guess it might make sense to have both:

- the existing one as convenience for migrating from scala 2
- the one from this PR as the "scala 3 way" of doing things

But maybe the scaladoc of the existing one should be updated to point users towards the new variant?

PR #2208, opened 2023-10-23
Add Encoder.at and Codec.at.

Lasering:
Adds the method `at` to Encoder:

```scala
def at(field: String): Encoder[A] = new Encoder[A] {
  override def apply(a: A): Json = Json.obj(field -> self.apply(a))
}
```

Since Codec extends both Decoder (which already has a method called at) and Encoder I also had to override it to:
```scala
final override def at(field: String): Codec[A] = Codec.from(super[Decoder].at(field), super[Encoder].at(field))
```

This is simultaneously overriding the method `Encoder.at` and `Decoder.at` (this works because Codec extends both).

Unfortunately to have this working I had to drop the modifier `final` from `Decoder.at`, not sure of the implications to MiMA.

See https://github.com/circe/circe/pull/1537 for a previous attempt.


modules/core/shared/src/main/scala/io/circe/Encoder.scala
```diff
@@ -88,6 +88,10 @@ trait Encoder[A] extends Serializable { self =>
   final def mapJson(f: Json => Json): Encoder[A] = new Encoder[A] {
     final def apply(a: A): Json = f(self(a))
   }
+
+  def at(field: String): Encoder[A] = new Encoder[A] {
```
zarthross, line 94: I think this is a reasonable ask, but can you add some scaladoc, and tests for both this `at` and the `at` in Codec?  Make sure to add some law tests to make sure an at'd Codec obeys the codec laws.

hamnis:
@Lasering If you can address the requested changes this week, this can be part of 0.14.7.

hamnis:
build error seems genuine: https://github.com/circe/circe/actions/runs/8738215116/job/24010673820#step:9:233

hamnis:
I dont think we can do this in 0.14.x since we break downstream libs `circe-generic-extras`, and who knows what else.

Perhaps we should move this to 0.15 ?

PR #2213, opened 2023-11-09
Fix transitive derivation for ConfiguredEncoder

### LINKED ISSUE #2212: Unexpected derivation of Encoder[Option[T]]
```scala
//> using scala "3"
//> using platform "jvm"
//> using dep "io.circe::circe-core::0.14.6"


object Types {
  import io.circe._
  import io.circe.derivation.ConfiguredEncoder

  given Configuration = Configuration.default.withSnakeCaseMemberNames

  case class Bar(barValue: Int)
  case class Foo(fooValue: Option[Bar]) derives ConfiguredEncoder
}

  import io.circe.syntax._
  println(Types.Foo(Some(Types.Bar(1))).asJson)
}

```
prints:
```
{
  "foo_value" : {
    "Some" : {
      "value" : {
        "bar_value" : 1
      }
    }
  }
}
```
expected:
```
{
  "foo_value" : {
    "bar_value" : 1
  }
}
```
Apparently, the `Encoder[Option[Bar]]` is derived instead of deriving only an `Encoder[Bar]` and then using the existing `io.circe.Encoder.encodeOption` for creating the `Encoder[Option[Bar]]`.

N.B.: as far as I can see, the same problem exists for all other circe typeclasses (in combination with scala 3) unless one imports `io.circe.generic.auto._`. (The latter may seem like a solution, but in case of configured derivation, it will break the configuration for child members, e.g. in above example one would get `"barValue"` instead of `"bar_value"`.)

MartinHH:
Fixes that derivation would not use existing givens if those depend on a given that needs to be derived (e.g. not using `encodeOption[A]` if the required `Encoder[A]` needs to be derived).

(Problem still prevails for all other derivable typeclasses.)

Fixes #2212

### DIFF
--- a/modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredEncoder.scala
+++ b/modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredEncoder.scala
@@ -19,6 +19,7 @@ package io.circe.derivation
 import scala.deriving.Mirror
 import scala.compiletime.constValue
 import io.circe.{ Encoder, Json, JsonObject }
+import io.circe.`export`.Exported
 
 trait ConfiguredEncoder[A](using conf: Configuration) extends Encoder.AsObject[A]:
   lazy val elemLabels: List[String]
@@ -54,8 +55,17 @@ trait ConfiguredEncoder[A](using conf: Configuration) extends Encoder.AsObject[A
           JsonObject.singleton(constructorName, json)
 
 object ConfiguredEncoder:
+
+  // needed to make derivation implicitly available when summoning elemEncoders:
+  private object GivenDerivation:
+    inline given deriveEncoder[A](using conf: Configuration)(using
+      inline A: Mirror.Of[A]
+    ): Exported[ConfiguredEncoder[A]] =
+      Exported(ConfiguredEncoder.derived[A])
+
   inline final def derived[A](using conf: Configuration)(using inline mirror: Mirror.Of[A]): ConfiguredEncoder[A] =
     new ConfiguredEncoder[A] with SumOrProduct:
+      import GivenDerivation.deriveEncoder
       lazy val elemLabels: List[String] = summonLabels[mirror.MirroredElemLabels]
       lazy val elemEncoders: List[Encoder[?]] = summonEncoders[mirror.MirroredElemTypes]
 
--- a/modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredEncoder.scala
+++ b/modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredEncoder.scala
@@ -57,7 +57,7 @@ trait ConfiguredEncoder[A](using conf: Configuration) extends Encoder.AsObject[A
 object ConfiguredEncoder:
 
   // needed to make derivation implicitly available when summoning elemEncoders:
-  private object GivenDerivation:
+  private[circe] object GivenDerivation:
     inline given deriveEncoder[A](using conf: Configuration)(using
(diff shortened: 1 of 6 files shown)


MartinHH:
If this solution approach gets approved, I would be willing to roll this out for all other typeclasses that are affected by the same problem.

modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredEncoder.scala
```diff
@@ -54,8 +55,17 @@ trait ConfiguredEncoder[A](using conf: Configuration) extends Encoder.AsObject[A
...
+  // needed to make derivation implicitly available when summoning elemEncoders:
+  private object GivenDerivation:
+    inline given deriveEncoder[A](using conf: Configuration)(using
+      inline A: Mirror.Of[A]
+    ): Exported[ConfiguredEncoder[A]] =
+      Exported(ConfiguredEncoder.derived[A])
+
   inline final def derived[A](using conf: Configuration)(using inline mirror: Mirror.Of[A]): ConfiguredEncoder[A] =
     new ConfiguredEncoder[A] with SumOrProduct:
+      import GivenDerivation.deriveEncoder
```
MartinHH, line 68: Side note: theoretically, this would allow to change the implementation of `io.circe.derivation.summonEncoder` from...
```scala
private[circe] inline final def summonEncoder[A](using Configuration): Encoder[A] =
  summonFrom {
    case _: Mirror.Of[A]     => ConfiguredEncoder.derived[A]
  }
```
...to...
```scala
private[circe] inline final def summonEncoder[A](using Configuration): Encoder[A] =
  summonInline[Encoder[A]]
```
...because derivation via `Mirror.Of[A]` is now available implictly so `summonInline[Encoder[A]]` will find it. (I tried that and all tests ran successfully.)

However: that code is also accessible via (deprecated) public API (see `io.circe.Derivation`) so I guess it's better to leave it as it is.

modules/tests/shared/src/test/scala-3/io/circe/ConfiguredDerivesSuite.scala
```diff
@@ -504,3 +504,21 @@ class ConfiguredDerivesSuite extends CirceMunitSuite:
       assert(result === Right(tree), result)
     }
   }
+
+  {
+    given Configuration = Configuration.default.withSnakeCaseMemberNames
+
+    case class Bar(barValue: Int)
+    case class Foo(fooValue: Option[Bar]) derives ConfiguredEncoder
+
```
zarthross, line 557: Can you add a codec law here for `Foo` ?
```suggestion
checkAll("Codec[Foo]", CodecTests[Foo].codec)
```

MartinHH: So I rolled out my solution approach onto the other type classes so that I can check codec laws (see new commit), but it turns out that "codec.encoder serializability" and "codec.decoder serializability" fail.

I will look into that when I find time, but if anyone has experience with that, any pointers would be welcome. (Been years since I last had to mess with java object serialization...)

MartinHH: > So I rolled out my solution approach onto the other type classes so that I can check codec laws (see new commit), but it turns out that "codec.encoder serializability" and "codec.decoder serializability" fail.
> 
> I will look into that when I find time, but if anyone has experience with that, any pointers would be welcome. (Been years since I last had to mess with java object serialization...)

Nevermind, I fixed it.

PR #2216, opened 2023-12-02
Add example for customized codecs

DeamonDev:


hamnis:
Thanks for the contribution.

@DeamonDev are you willing to fix the conflics here?

PR #2218, opened 2024-01-01
Fix microsite generation, update scala versions, update `sbt-circe-org` plugin version

CJSmith-0141:
The documentation hasn't been building on Github in a while and there are some features that I would like to document, so this is the first step in fixing the microsite. I also changed two of the badges:

- Removed gitter for Discord
- Changed the maven badge to `0.14.6` instead of the (abandoned?) `0.15.x` branch

PR #2219, opened 2024-01-07
adding a Nullable container that can represent value/null/undefined 

yurique:
Adding the `io.circe.Nullable` container to allow distinguishing between `undefined`(missing field)/`null`/actual value when encoding/decoding JSON objects.

~Also adding the `dropNoneValues` option to the `Configuration` (to be used instead of the corresponding option in the printer configuration since a printer configured to drop null values will conflict with the encoding of `Nullable`s)~

`dropNoneValues` configuration moved to a separate PR: https://github.com/circe/circe/pull/2226

---

### Rationale

Currently, there is no way to mix `undefined` (omitting the field from the JSON) and `null` values when en-/decoding, because the only way to omit a field from the JSON representation is to use a custom printer with `dropNullValues=true` - which cannot discriminate between the `null`-s we want to keep and those we want to omit.

(One of) the most common use-case for this is defining a payload for patch-like endpoints.

A relatively common way to do this using REST is to use `null` to indicate that some field needs to be "removed":

(update both `name` and `bio`)
```
PATCH /profiles/123
{
  "name": "new name",
  "bio": "new bio"
}
```

(only update `name`, keep `bio` as is)
```
PATCH /profiles/123
{
  "name": "new name",
}
```

(keep `name` as it, *remove* `bio`)
```
PATCH /profiles/123
{
  "bio": null
}
```

When we are in control of the endpoint - we can find workarounds. But when it's a third-party API, we're stuck.

With `Nullable`, the case class corresponding to this patch endpoint payload would look like something like this:

```scala
case class PatchProfile(
  name: Option[String], // can either change or leave as is
  bio: Nullable[String] // can change, leave as is, or remove
)
```

---

An example with `Nullable` fields from the tests:

```scala
  case class WithNullables(
    a: String,
    b: Nullable[String],
    c: Nullable[Int],
    d: Nullable[Boolean],
    e: Nullable[Box[String]],
    f: Nullable[List[String]],
    g: Option[Box[String]]
  ) derives Decoder,
        Encoder.AsObject

val foo =
    WithNullables(
      a = "a value",
      b = Nullable.Value("b value"),
      c = Nullable.Undefined,
      d = Nullable.Null,
      e = Nullable.Value(Box("boxed value")),
      f = Nullable.Value(List("a", "b", "c")),
      g = Some(Box("boxed in option"))
    )

val expected = Json.obj(
  "a" -> "a value".asJson,
  "b" -> "b value".asJson,
  // "c" will not be added to the object since it's a Nullable.Undefined
  "d" -> Json.Null,
  "e" -> Json.obj(
    "a" -> "boxed value".asJson
  ),
  "f" -> List("a", "b", "c").asJson,
  "g" -> Json.obj(
    "a" -> "boxed in option".asJson
  )
)

assert(foo.asJson === expected)
```

### TODO
* [x] corresponding changes in the generic-extras: https://github.com/circe/circe-generic-extras/pull/335
* [ ] decide on (better) name for `Nullable`
* [ ] decide on the API of the nullable: currently, a `Monoid` instance is provided; as well as the `toOption` / `toEither` / `isNull` / etc methods - is this enough/correct?
* [x] figure out bincompat


zarthross:
Thanks for the PR.  I haven't had a full chance to look through it, but at a glance I like the idea.  there is some prior art here to be aware of https://github.com/davenverse/patchy.  I'm not sure the name `Nullable` is great, but I'll discuss with Zach at our next meeting.  The big thing here is going to be the changes to Configuration and generic derivation.  Is it possible to split those changes into 2 PRs?

yurique:
> I'm not sure the name `Nullable` is great,

Yeah, I wasn't sure about it, either. I just went with it to prevent myself from prematurely spending 3 hours thinking about better alternatives :)

> Is it possible to split those changes into 2 PRs?

Of course! I'll do it this weekend.

yurique:
Removed the `Configuration` changes from this PR. Preparing a separate one for it.

yurique:
Wanted to point out two changes in this PR:
* https://github.com/circe/circe/pull/2219/files#diff-572f0d33079d45a31270258f1a38ec1bf73b864e6fb7295b53b39e0e50e57fe8R27
* https://github.com/circe/circe/pull/2219/files#diff-378e58dd77c3ec12a0e0b230ec5d55a15679e08f37f01ef83c5ee06967ca42d4R196

changing `using inline Mirror.Of[A]` to `using Mirror.Of[A]` (dropping `inline`).

The reason for this is it fails to compile with Scala 3.3.1 otherwise.

https://scala-lang.org/blog/2023/05/30/scala-3.3.0-released.html

> Known incompatibility: Stability of inline parameters
> ...
> There was a bug that resulted in the compiler assuming that all inline parameters are stable references. This could have led to unsound code being accepted by the compiler and potential runtime crashes.
> ...
> Such pieces of code are now rejected by the compiler. The migration for that change is simple, as it only requires removing the inline modifier from the parameter.

I'm not sure how this affects binary compatibility. Maybe it doesn't? (since those are `inline` functions).

---

I can revert these changes and revert to Scala 3.2.2, but it looks like the bump to 3.3.1 would help to move forward with this PR: https://github.com/circe/circe/pull/2218

yurique:
And it looks like the binary compatibility checks are passing now (without changes to the `Configuration`).

yurique:
A separate PR for the `dropNoneValues`: https://github.com/circe/circe/pull/2226

It's based on top of this one, so the diff includes the changes from here. 
Here's the diff agains this branch: https://github.com/yurique/circe/compare/feature/nullable...yurique:circe:feature/drop-none-values?expand=1

zarthross:
Can you add tests demonstrating how to use this with `forProduct` and hand written encoders and decoders?  Make sure to run the codec laws on those codecs as part of the tests.

yurique:
> Can you add tests demonstrating how to use this with `forProduct` and hand written encoders and decoders? Make sure to run the codec laws on those codecs as part of the tests.

there are some law tests (for scala3): https://github.com/circe/circe/pull/2219/files#diff-123ee492aac44e330aa51a8de6e8e0da093c49d99e28f26463da694167fb1b0aR290

as for `forProduct` - I don't think it will work, because of the built-in `Encoder` for `Nullable`: we can't encode the "omit the field from the containing object" behaviour in an encoder for the type of the field (and `forProduct*` is using the encoded values from the codec verbatim). Same with options - there's no way to use `forProduct` and make `None`s not appear in the json, they will be `null`s.
The special treatment for encoding/decoding of the `undefined` is implemented in the derived codecs only (same as for decoding of the `undefined` as `None` for options).

```scala
  implicit final def encodeNullable[A](implicit e: Encoder[A]): Encoder[Nullable[A]] = new Encoder[Nullable[A]] {
    final def apply(a: Nullable[A]): Json = {
      a match {
        case Nullable.Value(v)  => e(v)
        case Nullable.Null      => Json.Null
        case Nullable.Undefined => Json.Null
      }
    }
  }
```

```scala
  test("codec encode/decode correctly with Nullable") {
    case class Example(
      a: String,
      b: Nullable[String],
      c: Nullable[Int],
      d: Nullable[Boolean],
      e: Nullable[List[String]]
    )

    given Codec[Example] = Codec.forProduct5("a", "b", "c", "d", "e")(Example.apply)(p => (p.a, p.b, p.c, p.d, p.e))

    val example = Example("hello", Nullable.Null, Nullable.Value(1), Nullable.Undefined, Nullable.Value(List("a", "b", "c")))

    val encoded = example.asJson
    val decoded = encoded.as[Example]

    println(encoded.spaces2)

    assertEquals(decoded, Right(example)) // fails
  }
```

```json
{
  "a" : "hello",
  "b" : null,
  "c" : 1,
  "d" : null, // this field should not be here (undefined)
  "e" : [
    "a",
    "b",
    "c"
  ]
}
```

yurique:
in other words, the encoder for `Nullable` needs to look like something like this (`apply` of `Encoder` would have to be returning `Option[Json]` not just `Json`):

```scala
  implicit final def encodeNullable[A](implicit e: Encoder[A]): Encoder[Nullable[A]] = new Encoder[Nullable[A]] {
    final def apply(a: Nullable[A]): Option[Json] = {
      a match {
        case Nullable.Value(v)  => e(v)
        case Nullable.Null      => Some(Json.Null)
        case Nullable.Undefined => None
      }
    }
  }
```

hamnis:
would you mind fixing the merge conflicts here and we'll get a real review started.

yurique:
> would you mind fixing the merge conflicts here and we'll get a real review started.

Sure! Will do today.

hamnis:
I'm not sure we want two concepts for representing Json.Nullable things. `Option` is fine for that.

We might want something that represents the possible missing value as a type, So I propose we defines something like (scala 3 syntax):

```scala
enum UndefinedOr[+A]:
  case Defined[A](value: A) extends UndefinedOr[A]
  case Undefined extends UndefinedOr[Nothing]
```

this allows us to use `UndefinedOr[String]` which must be a string, or not defined.
or `Undefined[Option[Int]]` which may be all cases.

One drawback of this, is that the UndefinedOr does not have a "self" serialization, it can only be used as a property, which is due to the fact that we do not have undefined as part of the JSON AST.

yurique:
I wouldn't mind having this approach, if it can be done. 

Btw, there's `UndefOr` in scala-js, exactly for this purpose, so it would be nice to have a way to make `js.UndefOr` behave like this in codecs as well, not just `circe.UndefOr`.

---

What happens with just `Option` (without wrapping it in `UndefOr`) if we introduce this? Currently, an `Option` field will happily accept both `null` and `undefined` when decoding, and will always produce `null` when encoding. I guess it should stay this way, since there's a lot of code out there expecting this behaviour.

---

I'm pretty sure I was considering this approach initially. And, of course, I don't remember now why I went with the 3-case ADT instead.

hamnis:
> I wouldn't mind having this approach, if it can be done.
> 
> Btw, there's `UndefOr` in scala-js, exactly for this purpose, so it would be nice to have a way to make `js.UndefOr` behave like this in codecs as well, not just `circe.UndefOr`.
>

We could make a conversion for that, that could be explicitly imported.

> What happens with just `Option` (without wrapping it in `UndefOr`) if we introduce this? Currently, an `Option` field will happily accept both `null` and `undefined` when decoding, and will always produce `null` when encoding. I guess it should stay this way, since there's a lot of code out there expecting this behaviour.

Nothing changes with just Option, this would be additive.

yurique:
@hamnis I'm looking at this today. 

I'm not sure `UndefinedOr[A]` (and `UndefinedOr[Option[A]]`) will help with this PR's goal.

`UndefinedOr[A]` will allow serializing a field as `undefined` (just dropping it from the field completely), and we can already do this - with `Option` and `dropNullValues` for printer (which is problematic as well, but it's a different problem - https://github.com/circe/circe/pull/2226/files).

The problem I was aiming at is that there is no way to have part of the fields as `String or null` and other fields as `String or undefined` (with `String` as just an example, of course) - because, currently, we have to choose how `Option` behaves and can't configure it per field.


What could work is `NullOr`:

```scala
class MyData(
  a: String, // required
  b: Option[String], // string or undefined (with https://github.com/circe/circe/pull/2226)
  c: NullOr[String], // string or null
  d: Option[NullOr[String]], // string or null or undefined
)
```

What do you think?

yurique:
Drafted a PR with `NullOr` instead of the 3-way `Nullable`: https://github.com/circe/circe/pull/2272/files

---

For it to work correctly, the `Option` fields need to be serialized as `undefined` (not included in the JSON) instead of `null`, otherwise there is an ambiguity:

```scala
case class Data(
  field: Option{NullOr[String]]
)
```

`Data(field = Some(NullOr.Null)` will get encoded as this:

```json
{
  "field": null
}
```

And this JSON will be decoded with `field` = `None` instead of `field` = `Some(NullOr.Null)`, and there's no way to decode into `Some(NullOr.Null)` as long as `None`s are encoded as (and decoded from) `null`s.

In the draft PR I made the change to how optional fields are serialized, but there is a separate PR to make that behaviour configurable: https://github.com/circe/circe/pull/2226/files (it might be better to bundle these together)

PR #2226, opened 2024-01-13
Configuration: drop none values

yurique:


yurique:
included in https://github.com/circe/circe/pull/2272

PR #1923, opened 2022-02-24
fix Scala 3 semiauto

xuwei-k:
### before

```
Welcome to Scala 3.1.1 (1.8.0_322, Java OpenJDK 64-Bit Server VM).
Type in expressions for evaluation. Or try :help.

scala> case class A(x: Int); case class B(y: A)
// defined case class A
// defined case class B

scala> io.circe.generic.semiauto.deriveEncoder[B] // don't derive nested classes if semiauto!?
val res0: io.circe.Encoder.AsObject[B] = anon$1@1ea5f0e7
```

### after

```
Welcome to Scala 3.1.1 (1.8.0_322, Java OpenJDK 64-Bit Server VM).
Type in expressions for evaluation. Or try :help.

scala> case class A(x: Int); case class B(y: A)
// defined case class A
// defined case class B

scala> io.circe.generic.semiauto.deriveEncoder[B]
-- Error: ----------------------------------------------------------------------
1 |io.circe.generic.semiauto.deriveEncoder[B]
  |^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |cannot reduce summonFrom with
  | patterns :  case given encodeA @ _:io.circe.Encoder[A]
  | This location contains code that was inlined from Derivation.scala:53
  | This location contains code that was inlined from Derivation.scala:70
  | This location contains code that was inlined from Derivation.scala:51
  | This location contains code that was inlined from Derivation.scala:86
  | This location contains code that was inlined from semiauto.scala:27
1 error found
```

### Scala 2.x

```
Welcome to Scala 2.13.8 (OpenJDK 64-Bit Server VM, Java 1.8.0_322).
Type in expressions for evaluation. Or try :help.

scala> case class A(x: Int); case class B(y: A)
class A
class B

scala> io.circe.generic.semiauto.deriveEncoder[B]
                                              ^
       error: could not find Lazy implicit value of type io.circe.generic.encoding.DerivedAsObjectEncoder[B]
```

PR #2258, opened 2024-04-21
Align Scala 3 semiauto derivation with `circe-generic` behavior in Scala 2

### LINKED ISSUE #2111: Issue with encoding / decoding optional values when using semiauto derivation in Scala 3
**Scala:** 3.1.0
**Circe:** 0.14.3

I found an issue when encoding / decoding optional values, and I'm unsure if this is by design.

Optional values are not encoded as expected when the implicit encoder for that type is not on scope. The example below shows the effect. Is this intended?

```scala
import io.circe.Encoder

object Main extends App {
  case class Location(name: String)
  case class Person(name: String, location: Option[Location])

  {
    // Provides only a Person Encoder
    implicit lazy val personEncoder: Encoder.AsObject[Person] =
      deriveEncoder[Person]

    println(personEncoder(Person("John", Some(Location("New York")))))
    println(personEncoder(Person("Mary", None)))

// Results in a nested object when the location is Some(...):
//    {
//      "name" : "John",
//      "location" : {
//        "Some" : {
//          "value" : {
//            "name" : "New York"
//          }
//        }
//      }
//    }
//    {
//      "name" : "Mary",
//      "location" : {
//        "None$" : null
//      }
//    }
  }

  {
    // Provides both a Person Encoder and a Location Encoder
    implicit lazy val locationEncoder: Encoder.AsObject[Location] =
      deriveEncoder[Location]

    implicit lazy val personEncoder: Encoder.AsObject[Person] =
      deriveEncoder[Person]

    println(personEncoder(Person("John", Some(Location("New York")))))
    println(personEncoder(Person("Mary", None)))
    
// Results are as expected:
//    {
//      "name" : "John",
//      "location" : {
//        "name" : "New York"
//      }
//    }
//    {
//      "name" : "Mary",
//      "location" : null
//    }
  }
}

```

### LINKED ISSUE #2126: Support semiauto derivation without recursive derivation

### LINKED ISSUE #2212: Unexpected derivation of Encoder[Option[T]]

mrdziuban:
Fixes #2111 
Fixes #2126 
Fixes #2212 
Closes #1923 
Closes #2183 
Closes #2213 

Updates semiauto derivation in Scala 3 -- `Codec.AsObject.derived`, `Decoder.derived`, and `Encoder.AsObject.derived` -- to align it with `circe-generic` behavior in Scala 2 and follow the rules specified in 1.i and 2.iv of this comment: https://github.com/circe/circe/pull/2183#issuecomment-1839047060

> 1. How should semiauto derivation work by default for case classes?
    i. It should fail if any case class field does not have a Decoder/Encoder
> ...
> 2. How should semiauto derivation work by default for sealed traits?
> ...
>    iv. It should automatically derive a Decoder/Encoder for all subtypes

Some other notes:

- 2.iv is subject to the same rule as 1.i for `case class`es that are part of a `sealed trait` hierarchy -- derivation will not auto-recurse for the `case class` fields, e.g.
    ```scala
    sealed trait Color
    case object Red extends Color
    case object Green extends Color

    sealed trait Bird
    final case class Parrot(color: Color) extends Bird

    // This fails because there's no `Decoder[Color]` in scope
    io.circe.Decoder.derived[Bird]
    ```
- Multi-layered `sealed trait` hierarchies are handled properly -- derivation will auto-recurse for any nested sum type, e.g.
    ```scala
    sealed trait Animal
    final case class Fox() extends Animal
    sealed trait Dog
    final case class Corgi(color: String) extends Dog

    // This succeeds and automatically derives an instance of `Decoder[Dog]`
    io.circe.Decoder.derived[Animal]
    ```

### DIFF
--- a/modules/core/shared/src/main/scala-3/io/circe/derivation/package.scala
+++ b/modules/core/shared/src/main/scala-3/io/circe/derivation/package.scala
@@ -25,26 +25,50 @@ private[circe] inline final def summonLabels[T <: Tuple]: List[String] =
     case _: EmptyTuple => Nil
     case _: (t *: ts)  => constValue[t].asInstanceOf[String] :: summonLabels[ts]
 
+@deprecated("Use summonEncoders(derivingForSum: Boolean) instead", "0.14.7")
 private[circe] inline final def summonEncoders[T <: Tuple](using Configuration): List[Encoder[_]] =
+  summonEncoders(false)
+
+private[circe] inline final def summonEncoders[T <: Tuple](inline derivingForSum: Boolean)(using
+  Configuration
+): List[Encoder[_]] =
   inline erasedValue[T] match
     case _: EmptyTuple => Nil
-    case _: (t *: ts)  => summonEncoder[t] :: summonEncoders[ts]
+    case _: (t *: ts)  => summonEncoder[t](derivingForSum) :: summonEncoders[ts](derivingForSum)
 
+@deprecated("Use summonEncoder(derivingForSum: Boolean) instead", "0.14.7")
 private[circe] inline final def summonEncoder[A](using Configuration): Encoder[A] =
+  summonEncoder(false)
+
+private[circe] inline final def summonEncoder[A](inline derivingForSum: Boolean)(using Configuration): Encoder[A] =
   summonFrom {
     case encodeA: Encoder[A] => encodeA
-    case _: Mirror.Of[A]     => ConfiguredEncoder.derived[A]
+    case m: Mirror.Of[A] =>
+      inline if (derivingForSum) ConfiguredEncoder.derived[A]
+      else error("Failed to find an instance of Encoder[" + typeName[A] + "]")
   }
 
+@deprecated("Use summonDecoders(derivingForSum: Boolean) instead", "0.14.7")
 private[circe] inline final def summonDecoders[T <: Tuple](using Configuration): List[Decoder[_]] =
+  summonDecoders(false)
+
+private[circe] inline final def summonDecoders[T <: Tuple](inline derivingForSum: Boolean)(using
+  Configuration
+): List[Decoder[_]] =
   inline erasedValue[T] match
     case _: EmptyTuple => Nil
-    case _: (t *: ts)  => summonDecoder[t] :: summonDecoders[ts]
+    case _: (t *: ts)  => summonDecoder[t](derivingForSum) :: summonDecoders[ts](derivingForSum)
 
+@deprecated("Use summonDecoder(derivingForSum: Boolean) instead", "0.14.7")
 private[circe] inline final def summonDecoder[A](using Configuration): Decoder[A] =
+  summonDecoder(false)
+
+private[circe] inline final def summonDecoder[A](inline derivingForSum: Boolean)(using Configuration): Decoder[A] =
   summonFrom {
     case decodeA: Decoder[A] => decodeA
-    case _: Mirror.Of[A]     => ConfiguredDecoder.derived[A]
+    case m: Mirror.Of[A] =>
+      inline if (derivingForSum) ConfiguredDecoder.derived[A]
+      else error("Failed to find an instance of Decoder[" + typeName[A] + "]")
   }
 
 private[circe] inline def summonSingletonCases[T <: Tuple, A](inline typeName: Any): List[A] =
--- a/modules/core/shared/src/main/scala-3/io/circe/Derivation.scala
+++ b/modules/core/shared/src/main/scala-3/io/circe/Derivation.scala
@@ -26,18 +26,18 @@ import io.circe.derivation._
 object Derivation {
   inline final def summonLabels[T <: Tuple]: Array[String] = summonLabelsRec[T].toArray
   inline final def summonDecoders[T <: Tuple]: Array[Decoder[_]] =
-    derivation.summonDecoders[T](using Configuration.default).toArray
+    derivation.summonDecoders[T](false)(using Configuration.default).toArray
   inline final def summonEncoders[T <: Tuple]: Array[Encoder[_]] =
-    derivation.summonEncoders[T](using Configuration.default).toArray
+    derivation.summonEncoders[T](false)(using Configuration.default).toArray
 
-  inline final def summonEncoder[A]: Encoder[A] = derivation.summonEncoder[A](using Configuration.default)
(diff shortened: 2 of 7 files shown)


modules/core/shared/src/main/scala-3/io/circe/derivation/package.scala
```diff
@@ -25,26 +25,50 @@ private[circe] inline final def summonLabels[T <: Tuple]: List[String] =
...
+private[circe] inline final def summonEncoder[A](inline derivingForSum: Boolean)(using Configuration): Encoder[A] =
   summonFrom {
     case encodeA: Encoder[A] => encodeA
-    case _: Mirror.Of[A]     => ConfiguredEncoder.derived[A]
+    case m: Mirror.Of[A] =>
+      inline if (derivingForSum) ConfiguredEncoder.derived[A]
+      else error("Failed to find an instance of Encoder[" + typeName[A] + "]")
   }
 
+@deprecated("Use summonDecoders(derivingForSum: Boolean) instead", "0.14.7-BL")
```
hamnis: BL?

mrdziuban: Whoops sorry, short for my company name where I was considering publishing a separate version of this to use. Updated

hamnis (dismissed):
LGTM

perhaps name the boolean parameter applications, but that is a nit.

mrdziuban:
> perhaps name the boolean parameter applications

Good idea, updated

zarthross (commented):
I think this probably looks fine, but we need some tests around this to prevent future regressions.

mrdziuban:
👍 ported over the tests from #2183

mrdziuban:
It looks like the CI failure is unrelated to these changes, possibly related to issues with arbitrary `String`s:

```
==> X io.circe.pointer.literal.PointerInterpolatorSuite.The pointer string interpolater should work with arbitrary interpolated strings  0.277s munit.FailException: /home/runner/work/circe/circe/modules/pointer-literal/src/test/scala/io/circe/pointer/literal/PointerInterpolatorSuite.scala:63
62:
63:  property("The pointer string interpolater should work with arbitrary interpolated strings") {
64:    Prop.forAll(ScalaCheckInstances.genPointerReferenceString) { (v: String) =>

Failing seed: 2sQv2jbX3mSJSLkAIqZAf8CVtEYkC-V8ORO70IOuSMB=

  override def scalaCheckInitialSeed = "2sQv2jbX3mSJSLkAIqZAf8CVtEYkC-V8ORO70IOuSMB="

Falsified after 22 passed tests.
> Labels of failing property: 
Expected Right(/foo/~0/bar) but got Left(io.circe.pointer.PointerSyntaxError: Syntax error at 5; expected: token character)
> ARG_0: "~"
> ARG_0_ORIGINAL: "㜐臑騱ୡ≽긋嬕枖ᨙ蜛㜼쨋踆~1왘锧"
    at munit.Assertions.fail(Assertions.scala:252)
    at munit.Assertions.fail$(Assertions.scala:246)
    at munit.FunSuite.fail(FunSuite.scala:11)
    at munit.ScalaCheckSuite.$anonfun$propToTry$4(ScalaCheckSuite.scala:104)
    at scala.util.Try$.apply(Try.scala:210)
    at munit.ScalaCheckSuite.propToTry(ScalaCheckSuite.scala:104)
    at munit.ScalaCheckSuite.$anonfun$scalaCheckPropTransform$3(ScalaCheckSuite.scala:45)
    at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:475)
    at munit.Suite$$anon$1.execute(Suite.scala:28)
    at scala.concurrent.impl.Promise$Transformation.submitWithValue(Promise.scala:429)
    at scala.concurrent.impl.Promise$DefaultPromise.submitWithValue(Promise.scala:338)
    at scala.concurrent.impl.Promise$DefaultPromise.dispatchOrAddCallbacks(Promise.scala:312)
    at scala.concurrent.impl.Promise$DefaultPromise.transform(Promise.scala:128)
    at munit.internal.FutureCompat$ExtensionFuture.transformCompat(FutureCompat.scala:16)
    at munit.ScalaCheckSuite.$anonfun$scalaCheckPropTransform$2(ScalaCheckSuite.scala:47)
    at munit.Test.$anonfun$withBodyMap$1(Test.scala:34)
```

hamnis (approved):

PR #2261, opened 2024-04-25
Add back some `Encoder.AsObject`

zarthross:
Added comments to explain the current Scala 3 short comings in this area.


zarthross:
See #2228

hamnis (approved):

PR #2268, opened 2024-05-13
Do not differentiate between native and non-native for refined

hamnis:
Use 0.11.1 for both

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -25,8 +25,7 @@ ThisBuild / circeRootOfCodeCoverage := Some("rootJVM")
 val catsVersion = "2.10.0"
 val jawnVersion = "1.5.1"
 val shapelessVersion = "2.3.10"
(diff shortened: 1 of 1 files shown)


zarthross:
@hamnis Is this not a binary breaking change for users of circe-refined?  

The reason native use a newer version was because native required a newer version and did not exist when circe-refined was created, so in a way wasn't binary breaking.

hamnis:
it might be, but do we have to be so strict with regards to binary breakage in a third party integration module?

If we are, we should spin it off into an own repo, which has its own release schedule.

armanbilge:
> If we are, we should spin it off into an own repo, which has its own release schedule.

Agree. We should do this for all the circe integrations, same as we did for http4s.

hamnis:
Closing for now. We will spin off refined into a separate repo

PR #2272, opened 2024-05-27
a `NullOr` container to represent value/null

yurique:
This is the continuation of the previous attempt: https://github.com/circe/circe/pull/2219

---

Adding the `io.circe.NullOr ` container to allow representing `null` values when encoding/decoding JSON objects.

Adding `dropNoneValues` to the configuration (for scala-3), `true` by default.
...otherwise `Options` will serialize as `null` and `NullOr` will not work as intended. 

The problem: configured (derived) codecs are not part of core for scala-2: the only way was to change the default behaviour (to always omit `None` values from JSON), but there will be no way to change this without using the generic-extras (which will need to be updated to have the corresponding configuration option).

---

### Rationale

Currently, there is no way to mix `undefined` (omitting the field from the JSON) and `null` values when en-/decoding, because the only way to omit a field from the JSON representation is to use a custom printer with `dropNullValues=true` - which cannot discriminate between the `null`-s we want to keep and those we want to omit.

(One of) the most common use-case for this is defining a payload for patch-like endpoints.

A relatively common way to do this using REST is to use `null` to indicate that some field needs to be "removed":

(update both `name` and `bio`)
```
PATCH /profiles/123
{
  "name": "new name",
  "bio": "new bio"
}
```

(only update `name`, keep `bio` as is)
```
PATCH /profiles/123
{
  "name": "new name",
}
```

(keep `name` as it, *remove* `bio`)
```
PATCH /profiles/123
{
  "bio": null
}
```

When we are in control of the endpoint - we can find workarounds. But when it's a third-party API, we're stuck.

With `NullOr`, the case class corresponding to this patch endpoint payload would look like something like this:

```scala
case class PatchProfile(
  name: Option[String], // can either change or leave as is
  bio: Option[NullOr[String]] // can change, leave as is, or remove
)
```

---

### TODO
* [x] I changed how optional fields are serialized (`None`s are omitted instead of `null`) - it should be configurable instead (for those who want to keep the old behavior), but configured codecs are not part of core for scala-2: https://github.com/circe/circe/pull/2226
* [ ] update the corresponding changes in the generic-extras: https://github.com/circe/circe-generic-extras/pull/335
* [ ] better name instead of `NullOr`?
* [x] figure out bincompat issues if any


modules/generic/shared/src/main/scala-2/io/circe/generic/Deriver.scala
```diff
@@ -75,7 +75,13 @@ class Deriver(val c: whitebox.Context)
   """
 
   protected[this] def encodeField(name: String, encode: TermName, value: TermName): Tree =
-    q"($name, $encode($value))"
+    q"""if ($value == _root_.io.circe.Nullable.Undefined) {
```
hamnis: still referencing the old type

yurique: ah, yes - thanks!
I've started with the scala3 part :), was waiting for CI to discover more things for me to fix.

yurique: I'll need to remind myself about how it works here. It's easy to fix for `NullOr`, but the interaction with `Option` gets tricky.

yurique: or maybe it's simpler than I'm thinking 🤔

yurique:
copy of the last comment from the previous PR:
---

For it to work correctly, the `Option` fields need to be serialized as `undefined` (not included in the JSON) instead of `null`, otherwise there is an ambiguity:

```scala
case class Data(
  field: Option{NullOr[String]]
)
```

`Data(field = Some(NullOr.Null)` will get encoded as this:

```json
{
  "field": null
}
```

And this JSON will be decoded with `field` = `None` instead of `field` = `Some(NullOr.Null)`, and there's no way to decode into `Some(NullOr.Null)` as long as `None`s are encoded as (and decoded from) `null`s.

hamnis:
We should keep the old behavior as default until next major release.

yurique:
> We should keep the old behavior as default until next major release.

I made `dropNoneValues=false` by default, should match the current behavior

hamnis:
I am not happy with the NullOr name, but I dont really have a better suggestion.

https://www.merriam-webster.com/thesaurus/undefined does not really help either.
`MurkyOr` does not really inspire confidence.

hamnis (approved):
Leaving merging to @zarthross

yurique:
> I am not happy with the NullOr name

same with me

`Nullable`? Though it will probably conflict with other libs/user-code..

hamnis (changes requested):
needs fixing of conflicts

yurique:
> needs fixing of conflicts

done ✅

zarthross:
Thank you for your work, sorry it took so long for a review.  Also apologies ahead of the time for the long comment....

After carefully considering the `NullOr` datatype and consulting with
several other open source developers I believe we have the wrong
abstraction in this PR.

Apologies for repeating things from above, but I'm trying to be
comprehensive in my explaination of my thoughts

# `Profile`

If we start from looking at the database table your example `Profile`
type would have, it would probably look something like this:

``` sql
CREATE TABLE profile (
    "name" text NOT NULL, -- Required Name field
    "bio" text DEFAULT NULL -- Optional Bio
);
```

The natural encoding, when using something like `doobie` or even `circe`
would normally be:

``` scala
case class Profile(
  name: String, // Required Field
  bio: Option[String] // Not Required Field written as a value
)
```

# `NullOr`

The proposal in this PR is `NullOr` (summarized here)

``` scala
sealed trait NullOr[A]
case object Null extends NullOr[Nothing]
final case class Value[A](a: A) extends NullOr[A]
```

With the proposal in this PR a patch would be written like
`PatchProfile` (Seen above)

``` scala
case class PatchProfile(
  name: Option[String], // Update, or Unchanged
  bio: Option[NullOr[String]] // Update, Unchanged, Removed
)
```

### Examples

If we assume the original values of:

``` scala
Profile("name", Some("bio"))
```

The patched and updated types would then look something like this…

1.  For all fields the patch is kinda obvious

``` scala
// {"name": "new name", "bio": "new bio"}
PatchProfile(Some("new name"), Some(NullOr.Value("new bio")))
Profile("new name", Some("new bio"))
```

1.  Removing the bio and ignoring `name`

``` scala
// {"bio": null}
PatchProfile(None, Some(NullOr.Null))
Profile("name", None)
```

1.  Ignoring everything

``` scala
// { }
PatchProfile(None, None)
Profile("name", Some("bio"))
```

In my opinion ex 1 and ex 3 Probably look fine, but ex 2 looks strange.

1.  Variant

    For the sake of completeness lets look at a variant where `NullOr`
    wraps the `Option` instead.

    ``` scala
    case class PatchProfile(
      name: Option[String], // Update, or Unchanged
      bio: NullOr[Option[String]] // Update, Unchanged, Removed
    )
    ```

    1.  For all fields

    ``` scala
    // {"name": "new name", "bio": "new bio"}
    PatchProfile(Some("new name"), NullOr.Value(Some("new bio")))
    Profile("new name", Some("new bio"))
    ```

    1.  Removing the bio and ignoring `name`

    ``` scala
    // {"bio": null}
    PatchProfile(None, NullOr.Null)
    Profile("name", None)
    ```

    1.  Ignoring everything

    ``` scala
    // { }
    PatchProfile(None, NullOr.Value(None))
    Profile("name", Some("bio"))
    ```

    Example 2 Looks better, but there is an asymmetry here with `Option`
    between the `Profile` type and its `Patch` construction.

### Table

If we look at the types we are using here in tabular form we can see
clearly

|         | Option.None | NullOr.Null |
|---------|-------------|-------------|
| Patch A | Undefined   | Null        |
| A       | Null        | N/A         |

### Higher Kinds

Lets next look at the pattern of using higher kinded types. I've often
seen and used this pattern to reduce boilerplate when constructing a
`patch` type

If we construct the `Profile` type by wrapping each parameter in an
`F[]` we can start seeing a problem emerge.

``` scala
case class ProfileF[F[_]](
  name: F[String],
  bio: F[Option[String]]
)
type Profile = ProfileF[cats.id]
object Profile {
  def apply(name: String, bio: Option[String]): Profile = ???
  def unapply(p: Profile): Option[(String, Option[String])] = ???
}
```

The `cats.id` version seems fine, but if we look at `NullOr`

``` scala
type ProfilePatch = ProfileF[NullOr]
object ProfilePatch {
  def apply(name: NullOr[String], bio: NullOr[Option[String]]): Profile = ???
  def unapply(p: Profile): Option[(NullOr[String], NullOr[Option[String]])] = ???
}
```

In this case, the class has ended up with a nullable type where we
wanted an undefineable type…

If we try using `Option` instead, we end up with an Option of Option
which we are trying to avoid..

``` scala
type ProfilePatch = ProfileF[Option]
object ProfilePatch {
  def apply(name: Option[String], bio: Option[Option[String]]): Profile = ???
  def unapply(p: Profile): Option[(Option[String], Option[Option[String]])] = ???
}
```

This leads us to the problem earlier, of forcing users to use `NullOr`
in the `Profile` so null and undefined make sense again.

# `Patch` or `Undefineable` or `Defined` or `MaybeDefined` or `JMaybe` … `Bikeshed` (names are hard)

Lets for now refer to this type as `JMaybe` but I'm open to better
naming

Well define it pretty much like `NullOr`

``` scala
sealed trait JMaybe[A]
case object Undefined extends JMaybe[Nothing]
final case class Defined[A](a: A) extends JMaybe[A]
```

Lets create the patch type using this `JMaybe`

``` scala
case class PatchProfile(
  name: JMaybe[String],
  bio: JMaybe[Option[String]]
)
```

### Examples

Again If we assume the original values of:

``` scala
Profile("name", Some("bio"))
```

The patched and updated types would then look something like this…

1.  For all fields the patch is kinda obvious

``` scala
// {"name": "new name", "bio": "new bio"}
PatchProfile(Defined("new name"), Defined(Some("new bio")))
Profile("new name", Some("new bio"))
```

1.  Removing the bio and ignoring `name`

``` scala
// {"bio": null}
PatchProfile(Undefined, Defined(None))
Profile("name", None)
```

1.  Ignoring everything

``` scala
// { }
PatchProfile(Undefined, Undefined)
Profile("name", Some("bio"))
```

Now, the json types match more closely and naturally with the scala
types we see.

### Table

If we redefine the previous table for this type we see the symmetry with
null return

|         | Option.None | JMaybe.Undefined |
|---------|-------------|------------------|
| Patch A | Null        | Undefined        |
| A       | Null        | N/A              |

### Higher Kinds

and `ProfileF` works well and matches exactly the definition of
`PatchProfile`

``` scala
type ProfilePatch = ProfileF[JMaybe]
object ProfilePatch {
  def apply(name: JMaybe[String], bio: JMaybe[Option[String]]): Profile = ???
  def unapply(p: Profile): Option[(JMaybe[String], JMaybe[Option[String]])] = ???
}
```

### Decoding

If we look at implementing a decoder for `JMaybe` it should basically
just check if the cursor is failed or not, and if its not failed
delegate to the underlying decoder. The underlying decoder may or not be
an option, but thats okay.

``` scala
// Just examples, the use of `!c.incorrectFocus` vs `c.missingField` should be debated.
implicit def jmaybeDecoder[A](implicit A: Decoder[A]): Decoder[JMaybe[A]] = Decoder.instance {
  case c: HCursor => A(a).map(Defined.apply)
  case c: FailedCursor if !c.incorrectFocus => Undefined
  case c: FailedCursor => Left(DecodingFailure(MissingField, c.history))
}
```

Also, note we do not need a specific specialized
`Decoder[JMaybe[Option[A]]]` because `Option` already works on null, and
we check for undefinedness first.

Would a `Decoder[Option[JMaybe[A]]]` work? Definitely not, the
undefinedness would cause a failure before `JMaybe` decoded… You could
argue a specialized decoder in this case could help, but I'm not sure I
would encourage users to write it in the opposite order. (Should be a
warning in the docs about such a gotcha).

### Encoding

Now this is where `JMaybe` should differ from `NullOr` alot… `JMaybe`
CANNOT have an encoder. Since there is no representation for `undefined`
in circes ASTs there is simply no way to represent it… and thats a good
thing.

What we need to do is change all the generic derivations for encoders to
handle this special case of `JMaybe` and not output a key in those
cases. Users in the docs should also be told not use `dropNullKeys` in
the printer.

We would also need to add some helpers like `Encoder.forProduct` and
friends to handle `JMaybe`, and for when encoders are written using
\`Json.obj\` as is described in the
[documentation](https://circe.github.io/circe/codecs/custom-codecs.html)

As an added bonus of NOT having an encoder, `JMaybe` would not work with
circe-generic-extras until explicit support is added.. It simply would
not compile. This would help prevent users from making the mistake of
using it. Right now `NullOr` does not have this protection since it has
both an Encoder and Decoder derived for it.

yurique:
@zarthross I quickly skimmed through the comment (I will read it more carefully later today).

But I wanted to clarify a couple of things:

#### `Option of NullOr` vs `NullOr of Option`

I'm not sure I follow why we would want `NullOr[Option[A]]` over `Option[NullOr[A]]`.
With some pseudo-scala code I would write this:
```scala
case class Patch(
  name: Option[String],
  bio: Option[String | null]
)
```

but this would look weird:
```scala
case class Patch(
  name: Option[String],
  bio: null | Option[String]
)
```

In other words, when thinking about the same use-case with patches, I'm "thinking" in this order:
* do I want to change a particular field? 
  * if so - I provide a value, if not - I don't provide it (which is `Option`)
* if I do want to change the field - do I want to modify or remove it?
  * if I want to modify it, I provide the value, if not - I provide a `null`

which maps to `Option[NullOr[A]]`.

The "reverse" (`NullOr[Option[A]]`) would be this:
* do I want to remove the field or do I want to "skip it or modify it"
* ...

And the first step alone breaks the "intuition" for me. Why am I deciding whether I want to remove the field **before** I decide if I want to touch it at all?

#### `JMaybe`

This is the part I need to read more carefully. But so far I don't see where `Undefined` is coming from.

Did you mean to have `Undefined` as one of the cases for `JMaybe`? 
If so, in the first attempt I did have a 3-way encoding (`Nullable: .Value / .Null / .Undefined`) - https://github.com/circe/circe/pull/2219

zarthross:
> Option of NullOr vs NullOr of Option

The point of that exploration is that `NullOr[Option[A]]` is awkward and cannot work, sorry if that wasn't clear.

> JMaybe `Undefined`.

YUP, i typo'd the sealed trait definition for `JMabye` and put `Null` in there... and the rest of the examples used `Undefined`.  I have updated the above.  Sorry for the confusion.


TL;DR of the above though... `Option.None` IS `null` for all intents and purposes in Scala.  `NullOr` adds confusion by making 2 kinds of `null` depending on context.  This is why i propose we create a special construct to encode the `Undefined` instead, and do our best to make doing the wrong thing (Encoders of Undefined) hard to do and the right things easy.

yurique:
> `Option.None` IS `null` for all intents and purposes in Scala

Hmm, I would see `None` as something closer to `undefined` rather than `null` 

* `None` - undefined (missing, etc)
* `Some("test")` - value
* `Some(null)` - `null` (for those who are really brave!)

---

The `JMaybe` thing is clearer now 👍 

You're suggesting to use `Option` for value/null, and `JMaybe` for defined/undefined, right?

yurique:
Here's another angle (I'm trying to recall things I've considered, it's been a while).

#### Current state

* when decoding an `Option` field, `undefined` and `null` are the same in circe - both will get decoded as `None`,
* when encoding, `undefined` and `null` are *almost* the same: all `None`s will be encoded as either `null` or as `undefined` - depending on `Printer.dropNullValues`,
* we have a lot of `Option`s in our model

And we normally don't care about `null`/`undefined` (we use `Printer.dropNullValues` when needed).
And this behavior should remain unchanged. It's impossible to change all `Option`s to `JMaybe` or anything like that.

#### Missing part

What's missing is this one niche scenario - for some of the `Option`s we do want to distinguish between `undefined` and `null` when encoding/decoding into json. 
And the solution is to change *only those* `Option`s to be `Option[NullOr[...]]`.

This made me thinking I might need to tweak this PR a little: it would be best if the user can decide what to do with regular `Option`s (encode as `null` or `undefined`, decode both `undefined` and `null` as `None`) **independently** from what to do with the special case of `Option[JsonOr[...]]` 🤔

yurique:
>  And the solution is to change only those Options to be Option[NullOr[...]].

Not saying `JMaybe[Option[...]]` wouldn't work. Just thinking aloud. 

`JMaybe` looks very close to the `UndefOr` from scala-js. Maybe we could ask @sjrd to share some wisdom?

sjrd:
Would you care to write a short version of what you'd like my wisdom for? This is a long thread.

yurique:
> Would you care to write a short version of what you'd like my wisdom for? This is a long thread.

Sorry :)

We are contemplating the notion of `undefined` vs `null`, which "exists" in JSON - but does not exist (not representable) in circe. Edit: of course it exists in circe, I meant to say "not easily achievable / not possible with auto derivation"

One way would be to make `Option` represent `undefined`/value, and have an additional container (`NullOr`) to represent `null`/value. So we could have an optional, nullable field like `something: Option[NullOr[String]]`.

Another way is the other way around, kind of - to use `Option` to represent `null`/value, and introduce a separate container to represent `undefined`/value (which would look very similar to `UndefOr`). So the optional, nullable field would look like `something: UndefOr[Option[String]]`.

In this PR I went with the former, and we're discussing if that's the right way.

yurique:
I think I convinced myself it should be the other way around.

Here's why: if this was a scala-js-only library, we would naturally represent an optional, nullable value as `UndefOr[Option[String]]`, right?

So that's probably what makes sense to do here.

So we would have this:
* `UndefOr[String]` - `undefined` or value (no nulls)
* `UndefOr[Option[String]]` - `undefined`, `null` or value
* `Option[String]` - `undefined`, `null` or value, but don't distinguish `null` and `undefined` (like it is currently).



This leads me to an additional question: how do we make sure `UndefOr` from scala-js can be used as well, out of the box?

sjrd:
Yes, that seems like the best course of action.

In order to use Scala.js' `js.UndefOr`, the easiest is probably for *your* `UndefOr` to be defined in platform-dependent way. On the JVM it would be its own thing, but on JS it would be a type alias to `js.UndefOr`.

yurique:
> to be defined in platform-dependent way

That's what I was thinking, too  👍

yurique:
`type UndefOr[+A] = A | Unit` is causing problems in js 🤔 

```
[error]     |No best given instance of type io.circe.Encoder[scala.scalajs.js.UndefOr[Option[String]]] was found.
[error]     |I found:
[error]     |
[error]     |    io.circe.Encoder.encodeUndefOr[A](
[error]     |      /* ambiguous: both method encodeOption in object Encoder and method encodeUndefOr in object Encoder match type io.circe.Encoder[A] */
[error]     |        summon[io.circe.Encoder[A]]
[error]     |    )
[error]     |
[error]     |But both method encodeOption in object Encoder and method encodeUndefOr in object Encoder match type io.circe.Encoder[A].
```

yurique:
should I expect something like this to work?
```scala
private[circe] inline final def summonEncoders[T <: Tuple](inline derivingForSum: Boolean)(using
  Configuration
): List[Encoder[_]] =
  inline erasedValue[T] match
    case _: EmptyTuple => Nil

    // special case here
    case _: (scala.scalajs.js.UndefOr[t] *: ts)  => Encoder.encodeUndefOr(summonEncoder[t](derivingForSum)) :: summonEncoders[ts](derivingForSum)

    case _: (t *: ts)  => summonEncoder[t](derivingForSum) :: summonEncoders[ts](derivingForSum)
```


it does compile, but breaks some tests (unrelated to this PR)

yurique:
Looks like `case _: (scala.scalajs.js.UndefOr[t] *: ts)` matches everything

Does anyone know how to distinguish it? :)
the code:
```scala
private[circe] inline final def summonEncoders[T <: Tuple](inline derivingForSum: Boolean)(using
  Configuration
): List[Encoder[_]] =
  inline erasedValue[T] match
    case _: EmptyTuple => Nil
    case _: (scala.scalajs.js.UndefOr[t] *: ts) =>
      Encoder.encodeUndefOr(summonEncoder[t](derivingForSum)) :: summonEncoders[ts](derivingForSum)
    case _: (t *: ts)  => // never happens
      summonEncoder[t](derivingForSum) :: summonEncoders[ts](derivingForSum)
```

PR #2273, opened 2024-05-30
Add missing rule to "unserializableCodec" in `CodecTests`

satorg:
Rule "consistency with Codec.from" was introduced to `CodecTests` in #2131.
However, it was only added to the "codec" rule set but not the "unserializableCodec" one.
It doesn't seem though that the rule is specific to Java-level serialization,
therefore many tests that call for the "unserializableCodec" rule set only, miss the opportunity to check against that rule.


hamnis (approved):
Thanks!

PR #2275, opened 2024-06-01
Update versions for Scala Native 0.5

keynmol:
Remaining dependencies:
- https://github.com/cquiroz/scala-java-time/pull/508 (likely needs a Scala Native release, if workaround can't be found for tests)
- https://github.com/fthomas/refined/pull/1274 - needs a release


hamnis:
this is already done in #2254 , but thanks.

keynmol:
Ah thanks, I'll add a comment there instead

PR #2278, opened 2024-06-04
Fix stack overflow when deriving for local classes

### LINKED ISSUE #2263: `StackOverflowError` if local class since 0.14.7 with Scala 3
build.sbt

```scala
libraryDependencies += "io.circe" %% "circe-generic" % "0.14.7"

scalaVersion := "3.3.3"

run / fork := true
```

Main.scala

```scala
package example

import io.circe.Decoder
import io.circe.Json
import io.circe.syntax._

object Main {
    case class A(x: Int)
    object A {
      implicit val decoder: Decoder[A] = deriveDecoder[A]
    }

    println(Json.obj("x" -> Json.fromInt(3)).as[A])
  }
}
```

`sbt run` result

```
[info] running (fork) example.Main 
[error] Exception in thread "main" java.lang.StackOverflowError
[error] 	at example.Main$.example$Main$$$_$A$2(Main.scala:11)
[error] 	at example.Main$A$3$.<init>(Main.scala:12)
[error] 	at example.Main$.A$lzyINIT1$1(Main.scala:11)
[error] 	at example.Main$.example$Main$$$_$A$2(Main.scala:11)
[error] 	at example.Main$A$3$.<init>(Main.scala:12)
[error] 	at example.Main$.A$lzyINIT1$1(Main.scala:11)
[error] 	at example.Main$.example$Main$$$_$A$2(Main.scala:11)
[error] 	at example.Main$A$3$.<init>(Main.scala:12)
[error] 	at example.Main$.A$lzyINIT1$1(Main.scala:11)
[error] 	at example.Main$.example$Main$$$_$A$2(Main.scala:11)
[error] 	at example.Main$A$3$.<init>(Main.scala:12)
[error] 	at example.Main$.A$lzyINIT1$1(Main.scala:11)
[error] 	at example.Main$.example$Main$$$_$A$2(Main.scala:11)
[error] 	at example.Main$A$3$.<init>(Main.scala:12)
[error] 	at example.Main$.A$lzyINIT1$1(Main.scala:11)
[error] 	at example.Main$.example$Main$$$_$A$2(Main.scala:11)
```

mrdziuban:
Fixes #2263

There seem to be two things that cause the stack overflow:

1. Passing any values derived from the `Mirror`'s type members, e.g. `constValue[mirror.MirroredLabel]`, to a non-`inline` method as strict, by-value parameters
2. Referring to any of a `Mirror`'s type members as path-dependent types, e.g. `mirror.MirroredLabel`

The workaround for the first issue is pretty straightforward -- all values derived from the `Mirror` should be passed as by-name parameters.

The second issue is more complex. Marking the `Mirror` parameter as `inline` fixes it, which is why circe 0.14.6 didn't have the issue, but it only worked because 0.14.6 was using Scala 3.2.2. In Scala 3.3 and later using an `inline` parameter for path-dependent types results in an error:

```scala
(mirror : scala.deriving.Mirror.Of[A]) is not a valid type prefix, since it is not an immutable path
```

The solution is to take type parameters that represent the type members of the `Mirror` and use the type params in place of the path-dependent types, e.g.

```scala
def derived[A, L <: String](using inline m: Mirror.Of[A] { type MirroredLabel = L })
```

I don't think we want to do this in `def derived` though as it would break compatibility and proliferate to every method that calls `derived`.

Instead, this introduces a new `LazyMirror` type (definitely open to ideas on better names) that uses this pattern and captures the values from the `Mirror` that are necessary to support derivation. The `derived` methods then take a `LazyMirror` in place of a `Mirror`.

### DIFF
--- a/modules/core/shared/src/main/scala-3/io/circe/Derivation.scala
+++ b/modules/core/shared/src/main/scala-3/io/circe/Derivation.scala
@@ -16,10 +16,7 @@
 
 package io.circe
 
-import scala.compiletime.constValue
 import scala.deriving.Mirror
-import Predef.genericArrayOps
-import cats.data.{ NonEmptyList, Validated }
 import io.circe.derivation._
 
 @deprecated(since = "0.14.4")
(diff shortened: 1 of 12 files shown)


modules/core/shared/src/main/scala-3/io/circe/Derivation.scala
```diff
@@ -133,37 +130,62 @@ private[circe] trait DerivedDecoder[A] extends DerivedInstance[A] with Decoder[A
 }
 
 private[circe] trait EncoderDerivation:
-  inline final def derived[A](using inline A: Mirror.Of[A]): Encoder.AsObject[A] =
+  inline final def derived[A](using inline A: LazyMirror[A], inline E: Encoders[A]): Encoder.AsObject[A] =
```
mrdziuban, line 133: I was kinda surprised this wasn't reported as binary incompatible. Do `inline` params not matter for bin compat?

modules/core/shared/src/main/scala-3/io/circe/Derivation.scala
```diff
@@ -133,37 +130,62 @@ private[circe] trait DerivedDecoder[A] extends DerivedInstance[A] with Decoder[A
...
+    inline A: LazyMirror[A],
+    inline E: Encoders[A],
     inline configuration: Configuration
   ): Encoder.AsObject[A] =
     ConfiguredEncoder.derived[A]
 
-private[circe] trait EncoderDerivationRelaxed:
+private[circe] trait EncoderDerivationRelaxedLP:
+  @deprecated("Use variant that takes a LazyMirror", "0.14.8")
   inline final def derived[A: Mirror.Of](using
```
mrdziuban, line 144: Removing this _was_ reported as a binary incompatibility, I'm thinking because the `Mirror.Of` context bound doesn't desugar to an `inline` param?

modules/core/shared/src/main/scala-3/io/circe/Derivation.scala
```diff
@@ -133,37 +130,62 @@ private[circe] trait DerivedDecoder[A] extends DerivedInstance[A] with Decoder[A
...
     configuration: Configuration = Configuration.default
   ): Encoder.AsObject[A] =
+    given lm: LazyMirror[A] = LazyMirror.fromMirror(summon[Mirror.Of[A]])
+    ConfiguredEncoder.derived[A]
+
+private[circe] trait EncoderDerivationRelaxed extends EncoderDerivationRelaxedLP:
+  inline final def derived[A: LazyMirror: Encoders](using c: ConfigurationOrDefault): Encoder.AsObject[A] =
```
mrdziuban, line 151: `ConfigurationOrDefault` is necessary because we need to retain the old `derived` method for binary compatibility, and we can't say `using configuration: Configuration = Configuration.default` here without getting an error:

```scala
in trait EncoderDerivationRelaxed, multiple overloaded alternatives of method derived define default arguments.
```

mrdziuban (commented):
I'm going to follow up shortly with tests for deriving instances for local classes

modules/core/shared/src/main/scala-3/io/circe/Derivation.scala
hamnis: We might have to check for tasty-compatiblity here 
https://github.com/scalacenter/tasty-mima

it is possible that that is a missing check in mima, @SethTisue do you know?

hamnis:
I'm not sure I understand why this is needed/wanted?

is it to avoid using the `derives` keyword?

mrdziuban:
I think it goes a bit beyond just avoiding `derives`. For example, when cross building for Scala 2 you can't use `derives` so IMO an `implicit val` is the best option. Making it an `implicit lazy val` does fix the issue without any changes necessary in circe, but I don't think it's reasonable or possible to enforce that all downstream users add `lazy`. If there were a way to identify when the error would occur and fail compilation then it may be a different story, but I can't think of a way to do so. 

Ultimately I think fixing this is worthwhile so circe remains flexible and covers as many use cases as possible. If it's left unfixed then users need to be made aware that derivation for local classes has limitations -- having everything compile successfully but then getting an error at runtime is bound to surprise people.

hamnis:
when cross building to scala2, you also do not have `Decoder|Encoder.derived` unless you add that as implicit syntax.

mrdziuban:
True, but as demonstrated in #2263 using `circe-generic` also causes the error, and it can definitely be used while cross building.

hamnis:
closing in favour of #2287

PR #2279, opened 2024-06-05
use ISO_OFFSET_DATE_TIME formatter Encoder[ZonedDateTime]

### LINKED ISSUE #2239: usage of ISO_ZONED_DATE_TIME as default formatter in encoder
The default encoder for ZonedDateTime uses the [ISO_ZONED_DATE_TIME](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ISO_ZONED_DATE_TIME) formatter, which is only "ISO-like" as it 

> extends the ISO-8601 extended offset date-time format to add the time-zone. The section in square brackets is not part of the ISO-8601 standard

This results in parsing issues if this encoder is used outside of the JVM, e.g. when using circe to encode ZonedDateTime which is returned in a REST API. E.g. by default JavaScript is not able to decode it:

```
console.log(Date.parse('2024-03-05T10:04:49.49661+01:00')); // ISO_OFFSET_DATE_TIME, ISO-8601 Europe/Berlin
console.log(Date.parse('2024-03-05T09:01:07.763452Z')); // ISO_OFFSET_DATE_TIME, ISO-8601 UTC
console.log(Date.parse('2024-03-05T08:55:22.97961Z[UTC]')); // ISO_ZONED_DATE_TIME UTC
```
results in 
```
> 1709629489496
> 1709629267763
> NaN
```

Given this, I'd argue circe should use `ISO_OFFSET_DATE_TIME` as the formatter for the default encoder to be ISO conformant and allow compatibility esp. when used outside of the JVM.

This shouldn't have an impact on the decoder as it can handle both formats

an-tex:
fixes https://github.com/circe/circe/issues/2239

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Encoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Encoder.scala
@@ -52,8 +52,7 @@ import java.time.format.DateTimeFormatter.{
   ISO_LOCAL_DATE_TIME,
   ISO_LOCAL_TIME,
   ISO_OFFSET_DATE_TIME,
-  ISO_OFFSET_TIME,
-  ISO_ZONED_DATE_TIME
+  ISO_OFFSET_TIME
 }
 import java.time.temporal.{ ChronoField, TemporalAccessor }
 import java.util.Currency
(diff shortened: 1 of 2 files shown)


plokhotnyuk:
If you don't need zone ids then just convert your value to `OffsetDateTime` and encode it.

an-tex:
@plokhotnyuk yea you're probably right. thanks for forcing me to properly learn the difference between the two ;)

PR #2284, opened 2024-06-19
Fix cursor history loss on SeqDecoder.

### LINKED ISSUE #2283: Decoder error message: loss of information
Hi,

When I ran the following code with circe `0.14.7` and `0.14.8`, I got different error messages. It looks like the last version doesn't track path history when an error happens inside an array:

```scala
import io.circe.Codec
import io.circe.Decoder
import io.circe.literal.JsonStringContext
case class MyClass(f1: String, f2: String)
object MyClass {
  implicit val codec: Codec[MyClass] = deriveCodec
}
object Main extends App {
  val result = Decoder[List[MyClass]].decodeJson(json"""
    [{
      "f1": "value1"
    }]
  """)
  println(result)
}
```

```
> Left(DecodingFailure at .f2: Missing required field) //0.14.7
> Left(DecodingFailure at [0]: Missing required field) //0.14.8
```

The bug is there in both scala 2 & 3

rlecomte:
Fixes #2283 

This PR fixes the bug introduced in #1719 on SeqDecoder cursor history. 

I tried to keep the performance optimization recently introduced. Another possibility is to revert the implementation to the original code which looks less hacky in my opinion.

```
#before
[info] Benchmark                                      Mode  Cnt      Score      Error  Units
[info] NumberParsingBenchmark.decodeManyBigDecimals  thrpt    5   1459.271 ±  371.206  ops/s
[info] NumberParsingBenchmark.decodeManyBigInts      thrpt    5   1666.720 ±  228.149  ops/s
[info] NumberParsingBenchmark.decodeManyDoubles      thrpt    5  10748.210 ±  320.140  ops/s
[info] NumberParsingBenchmark.decodeManyLongs        thrpt    5  11131.330 ± 1333.619  ops/s

#after   
[info] Benchmark                                      Mode  Cnt      Score      Error  Units
[info] NumberParsingBenchmark.decodeManyBigDecimals  thrpt    5   1503.769 ±  115.695  ops/s
[info] NumberParsingBenchmark.decodeManyBigInts      thrpt    5   1695.076 ±  449.295  ops/s
[info] NumberParsingBenchmark.decodeManyDoubles      thrpt    5  10075.492 ± 1772.034  ops/s
[info] NumberParsingBenchmark.decodeManyLongs        thrpt    5  12766.322 ±  994.195  ops/s
```


hamnis (approved):
Thanks a lot.

PR #2286, opened 2024-07-02
Reduce warnings

hamnis:
The goal is to be able to turn on fatalWarnings on compile


zarthross (dismissed):

modules/core/shared/src/main/scala/io/circe/package.scala
```diff
@@ -55,4 +55,9 @@ package io
  *  detailed introduction to the library.
  */
 // scalastyle:on
-package object circe
+package object circe {
+  private[circe] def void[A](a: => A): Unit = {
+    val _ = a
+    ()
+  }
```
zarthross, line 62: would it not be better to just put the `@nowarn` annotation in the appropriate places that do this?  This feels like an unnecessary runtime cost until the JIT whisks it away when it wills.

The docs for `@nowarn` also explicitly show examples of this kinda thing https://www.scala-lang.org/api/2.13.12/scala/annotation/nowarn.html

hamnis: the warning on 3 still persists, so we only get the supressions for scala 2.

Issue #2089, opened 2023-02-09
Bincompat problem with derivation in 0.14.4 on Scala 3


ybasket:
Hey,
first of all: Thank you for the library and the huge step forward #1800 brought. I found an issue with bincompat though:
If code using generic derivation on Scala 3 was compiled against an earlier 0.14 version of circe (in my case `sttp-apispec`), using it together with the new 0.14.4 may throw a `NoSuchMethodError`:
```
java.lang.NoSuchMethodError: 'scala.collection.Iterable io.circe.DerivedEncoder.encodedIterable$(io.circe.DerivedEncoder, scala.Product)'
```

The reason for that is that `encodedIterable` was not `inline` and got removed in #1800 (`DerivedEncoder` was kept package-private, but not the methods). Similar goes for `DerivedDecoder`, it has also a bunch of methods that are now removed.

Looking forward, one could (probably) either bring those back for bincompat and remove them in 0.15 or accept the breakage and ideally add a warning to the release notes.

ahjohannessen:
I got the same issue 👍 for the fix

zarthross:
Thanks for letting us know and for submitting the PR, we are looking into it.

PR #2287, opened 2024-07-02
Use macros to fix stack overflow when deriving for local classes

### LINKED ISSUE #2263: `StackOverflowError` if local class since 0.14.7 with Scala 3

mrdziuban:
Fixes #2263 
Supersedes and closes #2278 if merged

@hamnis this is an alternate approach to what I did in #2278. By using macros for derivation we can match on the `Mirror` and extract the `MirroredLabel`, `MirroredElemLabels`, and `MirroredElemTypes` types without causing a stack overflow.

There are two caveats (the tests will fail if either of these things change in the future):

1. Any methods that call the macros must mark their `Mirror.Of` as `inline`
2. Any parameters derived from the `Mirror` in the macro must be passed by-name, e.g. `fromProduct` in `ConfiguredDecoder.ofProduct` and `ConfiguredEncoder.ofSum`

### DIFF
--- a/modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredCodec.scala
+++ b/modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredCodec.scala
@@ -23,6 +23,40 @@ import io.circe.{ Codec, Decoder, Encoder, HCursor, JsonObject }
 
 trait ConfiguredCodec[A] extends Codec.AsObject[A], ConfiguredDecoder[A], ConfiguredEncoder[A]
 object ConfiguredCodec:
+  @deprecated("Use ofProduct and ofSum", "0.14.10")
+  private[derivation] def inline$of[A](
+    nme: String,
+    decoders: => List[Decoder[?]],
+    encoders: => List[Encoder[?]],
+    labels: List[String]
+  )(using
+    conf: Configuration,
+    mirror: Mirror.Of[A],
+    defaults: Default[A]
+  ): ConfiguredCodec[A] = mirror match
+    case mirror: Mirror.ProductOf[A] =>
+      new ConfiguredCodec[A] with SumOrProduct:
+        val name = nme
+        lazy val elemDecoders = decoders
+        lazy val elemEncoders = encoders
+        lazy val elemLabels = labels
+        lazy val elemDefaults = defaults
+        def isSum = false
+        def apply(c: HCursor) = decodeProduct(c, mirror.fromProduct)
+        def encodeObject(a: A) = encodeProduct(a)
+        override def decodeAccumulating(c: HCursor) = decodeProductAccumulating(c, mirror.fromProduct)
+    case mirror: Mirror.SumOf[A] =>
+      new ConfiguredCodec[A] with SumOrProduct:
+        val name = nme
+        lazy val elemDecoders = decoders
+        lazy val elemEncoders = encoders
+        lazy val elemLabels = labels
+        lazy val elemDefaults = defaults
+        def isSum = true
+        def apply(c: HCursor) = decodeSum(c)
+        def encodeObject(a: A) = encodeSum(mirror.ordinal(a), a)
+        override def decodeAccumulating(c: HCursor) = decodeSumAccumulating(c)
+
   private def ofProduct[A](
     nme: String,
     decoders: => List[Decoder[?]],
--- a/modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredCodec.scala
+++ b/modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredCodec.scala
@@ -17,47 +17,97 @@
 package io.circe.derivation
 
 import scala.deriving.Mirror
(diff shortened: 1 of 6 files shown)


hamnis (dismissed):
I found this easier to understand

hamnis:
Letting @zarthross decide to merge

modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredCodec.scala
```diff
@@ -17,47 +17,97 @@
...
 
 import scala.deriving.Mirror
-import scala.compiletime.constValue
+import scala.compiletime.{ constValue, summonInline }
+import scala.quoted.*
 import io.circe.{ Codec, Decoder, Encoder, HCursor, JsonObject }
 
 trait ConfiguredCodec[A] extends Codec.AsObject[A], ConfiguredDecoder[A], ConfiguredEncoder[A]
 object ConfiguredCodec:
-  private def of[A](nme: String, decoders: => List[Decoder[?]], encoders: => List[Encoder[?]], labels: List[String])(
```
zarthross, line 25: I would love to be shown wrong on this, but removing this method concerns me.

I know that when you use inline on a private method, (the original `derived` below calls this of), that there is a public accessor created, in this case it should be `inner$of`.

If we have some library, lets call if `A` that used theses macros, and some end user app called `B`.  
What happens if circe is updated to `ofProduct`, built a built with `of`... and `B` updates the version of circe to no longer include `inner#of`?  I feel like we will get a NoSuchMethodError in application `B`?

I'm worried we'll have a similar problem to https://github.com/circe/circe/issues/2089

zarthross: Not sure why mima doesn't catch this situation....

mrdziuban: Wow I didn't know about that behavior but you're definitely right, I confirmed in a fresh project. I'll put the `of` methods back

mrdziuban: Oh wait actually, just putting `of` back doesn't fix the issue. The `inline$of` method only exists in the bytecode if `of` is called from an `inline` method, so we could also need to add a dummy `inline` method that still calls it. Given that, what do you think the best course of action is @zarthross?

zarthross: Yeah, this is certainly an awkward part of scala3 inlines... we could try adding a private method with the name `inline$of`?  I'm not sure if scala 3 will allow that, or we could try a dummy inline method as you suggested... we'll just have to try a few things and check the generated bytecode....

mrdziuban: I ended up adding `inline$of` methods. They can't be `private` otherwise they cause a `java.lang.IllegalAccessError` instead of a binary incompatibility, but I made them `private[derivation]` and marked them as `deprecated`

zarthross (approved):
Thanks for the hard work on this!

PR #2290, opened 2024-07-12
Address minor compiler warnings

OddKristensen:


modules/core/shared/src/main/scala/io/circe/JsonNumber.scala
```diff
@@ -121,7 +121,8 @@ private sealed abstract class BiggerDecimalJsonNumber(input: String) extends Jso
 
   final def toLong: Option[Long] = toBiggerDecimal.toLong
   override final def toString: String = input
-  private[circe] final def appendToStringBuilder(builder: StringBuilder): Unit = builder.append(input)
+  private[circe] final def appendToStringBuilder(builder: StringBuilder): Unit =
+    builder.append(input): Unit
```
OddKristensen, line 125: Not sure if I am going overboard here by removing the _discarded non-Unit value_ errors. It feels kind of silly, yet there are quite a few of them.

PR #2291, opened 2024-07-20
Use values from a constant pool when possible

JD557:
Adds a constant pool of longs from -128 to 127 (like the JVM constant integer pool) end constant empty collections (strings, arrays and objects) to be reused when possible, to reduce allocations and memory usage.

Tested this on `file2.json` from https://www.kaggle.com/datasets/shrutimehta/zomato-restaurants-data (A 20MB JSON file with real world data) and got the following results (the `Dataset` object is the class retaining the full data, so that I can know the full retained size):

**Before**
Screenshot 2024-07-22 at 23 15 41
Screenshot 2024-07-22 at 23 16 01

**After**
Screenshot 2024-07-22 at 23 42 53
Screenshot 2024-07-22 at 23 42 39

---

In this case, there was about a 2MB drop in retained size, mostly by allocating less `JsonLong`'s

I did not perform any speed benchmarks, as I'm currently working on battery and don't have my charger 😅 

Either way, I imagine that on large datasets the speed will be highly dependent on your available heap/GC impact.

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Json.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Json.scala
@@ -500,7 +500,7 @@ object Json {
   /**
    * Create a `Json` value representing a JSON array from a collection of values.
    */
-  final def fromValues(values: Iterable[Json]): Json = 
+  final def fromValues(values: Iterable[Json]): Json =
     if (values.isEmpty) EmptyArray
     else JArray(values.toVector)
 
--- a/modules/core/shared/src/main/scala/io/circe/Json.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Json.scala
@@ -543,7 +543,7 @@ object Json {
   /**
    * Create a `Json` value representing a JSON number from an `Int`.
    */
-  final def fromInt(value: Int): Json = JNumber(JsonNumber.fromLong(value.toLong))
+  final def fromInt(value: Int): Json = JsonNumber.fromLong(value.toLong)
 
   /**
    * Create a `Json` value representing a JSON number or a null from an `Option[Int]`.
@@ -553,7 +553,7 @@ object Json {
   /**
    * Create a `Json` value representing a JSON number from a `Long`.
    */
-  final def fromLong(value: Long): Json = JNumber(JsonNumber.fromLong(value))
+  final def fromLong(value: Long): Json = JsonNumber.fromLong(value)
 
   /**
    * Create a `Json` value representing a JSON number or null from an optional `Long`.
--- a/modules/core/shared/src/main/scala/io/circe/Json.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Json.scala
@@ -479,6 +479,8 @@ object Json {
   final val Null: Json = JNull
   final val True: Json = JBoolean(true)
   final val False: Json = JBoolean(false)
+  private[this] final val EmptyArray: Json = JArray(Vector.empty)
+  private[this] final val EmptyString: Json = JString("")
 
   /**
    * Create a `Json` value representing a JSON object from key-value pairs.
@@ -498,7 +500,9 @@ object Json {
   /**
    * Create a `Json` value representing a JSON array from a collection of values.
    */
-  final def fromValues(values: Iterable[Json]): Json = JArray(values.toVector)
+  final def fromValues(values: Iterable[Json]): Json = 
+    if (values.isEmpty) EmptyArray
+    else JArray(values.toVector)
 
   /**
    * Create a `Json` value representing a JSON object from a [[JsonObject]].
@@ -515,7 +519,9 @@ object Json {
    *
    * Note that this does not parse the argument.
    */
-  final def fromString(value: String): Json = JString(value)
+  final def fromString(value: String): Json =
+    if (value.isEmpty) EmptyString
+    else JString(value)
 
   /**
    * Create a `Json` value representing a JSON String or null from an `Option[String]`.
@@ -537,7 +543,7 @@ object Json {
   /**
    * Create a `Json` value representing a JSON number from an `Int`.
    */
-  final def fromInt(value: Int): Json = JNumber(JsonLong(value.toLong))
+  final def fromInt(value: Int): Json = JNumber(JsonNumber.fromLong(value.toLong))
 
   /**
    * Create a `Json` value representing a JSON number or a null from an `Option[Int]`.
@@ -547,7 +553,7 @@ object Json {
   /**
    * Create a `Json` value representing a JSON number from a `Long`.
    */
-  final def fromLong(value: Long): Json = JNumber(JsonLong(value))
+  final def fromLong(value: Long): Json = JNumber(JsonNumber.fromLong(value))
 
   /**
    * Create a `Json` value representing a JSON number or null from an optional `Long`.
--- a/modules/core/shared/src/main/scala/io/circe/JsonNumber.scala
+++ b/modules/core/shared/src/main/scala/io/circe/JsonNumber.scala
@@ -241,10 +241,13 @@ object JsonNumber {
   * Constant pool of integer numbers between -128 to 127, similar to the JVM's constant pool for integers.
   */
   private val jsonLongConstantPool: ArraySeq[JsonLong] =
-    ArraySeq.tabulate(255)(x => JsonLong((x - 128).toLong))
+    ArraySeq.tabulate(256)(x => JsonLong((x - 128).toLong))
 
-  private[circe] final def fromLong(value: Long): JsonLong =
-    jsonLongConstantPool.applyOrElse[Int, JsonLong]((value - 128).toInt, _ => JsonLong(value))
+  private[circe] final def fromLong(value: Long): JsonLong = {
+    val idx = value + 128
+    if (idx >= 0 && idx < jsonLongConstantPool.size) jsonLongConstantPool.apply(idx.toInt)
+    else JsonLong(value)
+  }
 
   /**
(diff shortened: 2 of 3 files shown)


modules/core/shared/src/main/scala/io/circe/JsonNumber.scala
```diff
@@ -236,6 +237,15 @@ private[circe] final case class JsonFloat(value: Float) extends JsonNumber {
...
 object JsonNumber {
 
+  /**
+  * Constant pool of integer numbers between -128 to 127, similar to the JVM's constant pool for integers.
+  */
+  private val jsonLongConstantPool: ArraySeq[JsonLong] =
+    ArraySeq.tabulate(255)(x => JsonLong((x - 128).toLong))
+
+  private[circe] final def fromLong(value: Long): JsonLong =
+    jsonLongConstantPool.applyOrElse[Int, JsonLong]((value - 128).toInt, _ => JsonLong(value))
```
armanbilge: This is allocating a lambda `_ => JsonLong(value)` on every invocation. If the goal is performance, we might as well go for a raw `Array` and `if`/`else`.

JD557: Nice catch
Also, turns out that function was completely wrong and was fetching the wrong numbers, not sure how I didn't notice in my tests 🙃
On top of that, it was boxing/unboxing unnecessarily.

Fixed in ae1166a

Decompiled code:
```java
    public final JsonLong fromLong(long value) {
        long idx = value + 128L;
        if (idx >= 0L && idx < (long)this.jsonLongConstantPool().size()) {
            return this.jsonLongConstantPool().apply((int)idx);
        }
        return new JsonLong(value);
    }
```

JD557: Updated the results on the top message.

modules/core/shared/src/main/scala/io/circe/JsonNumber.scala
```diff
@@ -236,6 +237,18 @@ private[circe] final case class JsonFloat(value: Float) extends JsonNumber {
  */
 object JsonNumber {
 
+  /**
+  * Constant pool of integer numbers between -128 to 127, similar to the JVM's constant pool for integers.
+  */
+  private val jsonLongConstantPool: ArraySeq[JsonLong] =
+    ArraySeq.tabulate(256)(x => JsonLong((x - 128).toLong))
```
armanbilge:
```suggestion
  private[this] val jsonLongConstantPool: Array[JsonLong] =
    Array.tabulate(256)(x => JsonLong((x - 128).toLong))
```

modules/core/shared/src/main/scala/io/circe/JsonNumber.scala
```diff
@@ -236,6 +237,18 @@ private[circe] final case class JsonFloat(value: Float) extends JsonNumber {
...
+
+  private[circe] final def fromLong(value: Long): JsonLong = {
+    val idx = value + 128
+    if (idx >= 0 && idx < jsonLongConstantPool.size) jsonLongConstantPool.apply(idx.toInt)
```
armanbilge:
```suggestion
    if (idx >= 0 && idx < jsonLongConstantPool.length) jsonLongConstantPool.apply(idx.toInt)
```

armanbilge: Decompile after these changes and it should avoid a couple method calls from your previous result :)

JD557: Indeed, much cleaner:
```java
    public final JsonLong fromLong(long value) {
        long idx = value + 128L;
        if (idx >= 0L && idx < (long)jsonLongConstantPool.length) {
            return jsonLongConstantPool[(int)idx];
        }
        return new JsonLong(value);
    }
```

modules/core/shared/src/main/scala/io/circe/Json.scala
```diff
@@ -479,6 +479,8 @@ object Json {
   final val Null: Json = JNull
   final val True: Json = JBoolean(true)
   final val False: Json = JBoolean(false)
+  private[this] final val EmptyArray: Json = JArray(Vector.empty)
```
hamnis, line 482: private[this] will be removed in scala 3 at some point, so we should not use that
https://docs.scala-lang.org/scala3/reference/dropped-features/this-qualifier.html

plokhotnyuk:
@JD557 Use [jsoniter-scala-circe](https://github.com/plokhotnyuk/jsoniter-scala/tree/master/jsoniter-scala-circe) booster FTW! You will save much more CPU cycles and avoid much more allocations with it! It uses constants for boolean/null values and empty JSON arrays/objects when parsing. Also, caching of primitive numbers is already completed by Scala's boxing/unboxing functions in the `BoxesRunTime` class.

modules/core/shared/src/main/scala/io/circe/Json.scala
armanbilge: Sure, but it's still needed on Scala 2 for performance.

JD557: I don't have any strong opinions about this, so I'll do whatever you want, but I think I used `private[this]` because I saw that being used in other places.

(I'm on holiday, so it's not very convenient for me to link to other usages)

JD557:
> Also, caching of primitive numbers is already completed by Scala's boxing/unboxing functions in the `BoxesRunTime` class.

I admit that I'm not very good at this kind of stuff, but I think this PR still improves the status quo by working on JsonNumber directly, no? At least the extra JsonLong wrapper is not allocated.

modules/core/shared/src/main/scala/io/circe/Json.scala
JD557: I'm finally back, so I can take a look at this.

However, now I'm confused, should I change this or not?

hamnis: I suppose we will put a pin in that for now. Once scala-3 drops this I guess we will have to split scala-2 and scala-3 for these cases.

hamnis (dismissed):

modules/core/shared/src/main/scala/io/circe/JsonNumber.scala
```diff
@@ -236,6 +236,18 @@ private[circe] final case class JsonFloat(value: Float) extends JsonNumber {
...
+   * Constant pool of integer numbers between -128 to 127, similar to the JVM's constant pool for integers.
+   */
+  private[this] val jsonLongConstantPool: Array[JsonLong] =
+    Array.tabulate(256)(x => JsonLong((x - 128).toLong))
+
+  private[circe] final def fromLong(value: Long): JsonLong = {
+    val idx = value + 128
+    if (idx >= 0 && idx < jsonLongConstantPool.length) jsonLongConstantPool.apply(idx.toInt)
+    else JsonLong(value)
+  }
```
zarthross: I think it would make more sense to store `JNumber` since thats what the end user will see the most, and it'll save 1 additional allocation.

```suggestion
  private[this] val jsonLongConstantPool: Array[JNumber] =
    Array.tabulate(256)(x => JsonLong((x - 128).toLong))

  private[circe] final def fromLong(value: Long): JNumber = {
    val idx = value + 128
    if (idx >= 0 && idx < jsonLongConstantPool.length) jsonLongConstantPool.apply(idx.toInt)
    else JNumber(value)
  }
```

modules/core/shared/src/main/scala/io/circe/Json.scala
```diff
@@ -537,7 +543,7 @@ object Json {
   /**
    * Create a `Json` value representing a JSON number from an `Int`.
    */
-  final def fromInt(value: Int): Json = JNumber(JsonLong(value.toLong))
+  final def fromInt(value: Int): Json = JNumber(JsonNumber.fromLong(value.toLong))
```
zarthross: If you change the table to `JNumber` as suggested in another commented then you would also change here:

```suggestion
  final def fromInt(value: Int): Json = JsonNumber.fromLong(value.toLong)
```

modules/core/shared/src/main/scala/io/circe/Json.scala
```diff
@@ -547,7 +553,7 @@ object Json {
   /**
    * Create a `Json` value representing a JSON number from a `Long`.
    */
-  final def fromLong(value: Long): Json = JNumber(JsonLong(value))
+  final def fromLong(value: Long): Json = JNumber(JsonNumber.fromLong(value))
```
zarthross: If you change the table to `JNumber` as suggested in another commented then you would also change here:

```suggestion
  final def fromLong(value: Long): Json = JsonNumber.fromLong(value)
```

modules/core/shared/src/main/scala/io/circe/JsonNumber.scala
```diff
@@ -259,7 +271,7 @@ object JsonNumber {
     else {
       val longValue = java.lang.Long.parseLong(value)
 
-      if (value.charAt(0) == '-' && longValue == 0L) JsonDecimal(value) else JsonLong(longValue)
+      if (value.charAt(0) == '-' && longValue == 0L) JsonDecimal(value) else fromLong(longValue)
```
zarthross: If you change the table to `JNumber` as suggested in another commented then you would also change here:

```suggestion
      if (value.charAt(0) == '-' && longValue == 0L) JsonDecimal(value) else fromLong(longValue).value
```

zarthross (dismissed):
Thanks for adding this optimization!  Here some some feedback that would make this even better.

hamnis (approved):

PR #2296, opened 2024-08-16
Update munit to 1.0.1

mzuehlke:
munit-scalacheck is now versioned independently of munit

```
[error] (testsNative / update) found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
[error]
[error] 	* org.scala-native:test-interface_native0.5_2.13:0.5.4 (strict) is selected over {0.5.1}
[error] 	    +- org.scalameta:munit_native0.5_2.13:1.0.1           (depends on 0.5.4)
[error] 	    +- org.scalatest:scalatest-core_native0.5_2.13:3.2.18 (depends on 0.5.1)
[error] 	    +- org.scalacheck:scalacheck_native0.5_2.13:1.18.0    (depends on 0.5.1)
[error]
[error]
[error] this can be overridden using libraryDependencySchemes or evictionErrorLevel
```

Using `libraryDependencySchemes` wasn't working 😢 
I tried
```Scala
ThisBuild / libraryDependencySchemes ++= Seq(
    "org.scala-native" %% "test-interface" % VersionScheme.Always
)
```
and 
```Scala
ThisBuild / libraryDependencySchemes ++= Seq(
    "org.scala-native" %%% "test-interface" % VersionScheme.Always
)
```
closes: #2294

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -23,7 +23,9 @@ ThisBuild / scalafixAll / skip := tlIsScala3.value
 ThisBuild / ScalafixConfig / skip := tlIsScala3.value
 ThisBuild / circeRootOfCodeCoverage := Some("rootJVM")
 
-ThisBuild / evictionErrorLevel := Level.Info
+//This is really bad, but we need to add the _native0.5 for this: https://github.com/sbt/sbt/issues/7140
+ThisBuild / libraryDependencySchemes +=
+  "org.scala-native" %% "test-interface_native0.5" % VersionScheme.Always
 
 val catsVersion = "2.12.0"
 val jawnVersion = "1.6.0"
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -23,6 +23,8 @@ ThisBuild / scalafixAll / skip := tlIsScala3.value
 ThisBuild / ScalafixConfig / skip := tlIsScala3.value
 ThisBuild / circeRootOfCodeCoverage := Some("rootJVM")
 
+ThisBuild / evictionErrorLevel := Level.Info
```
hamnis: I dont think we should do this.
This will change the error level for all dependencies.
I would rather that we have explicit overrides for those we know are safe.

hamnis (approved):

hamnis:
thanks!

mzuehlke:
Looks like a bug / shortcoming in sbt. Thanks for finding this fix.

PR #2304, opened 2024-08-27
remove `cats.instances` imports

xuwei-k:
https://github.com/typelevel/cats/releases/tag/v2.2.0

> all Cats type class instances for standard library types are now available in implicit scope, and no longer have to be imported. This has a number of benefits, including faster compile times and fewer things to think about. Please see https://github.com/typelevel/cats/pull/3043, [this blog post](https://meta.plasm.us/posts/2019/09/30/implicit-scope-and-cats/), [the 2.2.0-M1 release notes](https://github.com/typelevel/cats/releases/tag/v2.2.0-M1), and the migration guide below for more detailed information about this change.


hamnis (approved):
Thanks!

PR #2316, opened 2024-10-04
Add dependency in literal to jawn instead of optional to parser

vandmo:
```scala
//> using scala 3.5.1
//> using dep io.circe::circe-literal:0.14.10

import io.circe.literal.*
val key = "theKey"
println(json"{ $key: 123 }")
```
fails with "class not found".
It works after adding `//> using dep io.circe::circe-parser:0.14.10`


hamnis:
actually, the dependency should be to `circe-jawn` and not `parser`.

vandmo:
> It works after adding `//> using dep io.circe::circe-parser:0.14.10`

Thanks!
It works after adding //> using dep io.circe::circe-jawn:0.14.10 as well :)
I updated the MR.

build.sbt
```diff
@@ -337,7 +337,7 @@ lazy val literal = circeCrossModule("literal", CrossType.Pure)
   .nativeSettings(
     tlVersionIntroduced := List("2.12", "2.13", "3").map(_ -> "0.14.9").toMap
   )
-  .dependsOn(core, parser % Test, testing % Test)
+  .dependsOn(core, jawn, parser % Test, testing % Test)
```
hamnis:
```suggestion
  .dependsOn(core, jawn, testing % Test)
```

hamnis (approved):

PR #2341, opened 2025-01-18
Derivation for fat structures

road21:
Currently scala3 derivations of `Encoder`, `Decoder` for long structures (more than 28 fields) is not supported (at least for standard compiler options).
This happens because of the error:

```
Maximal number of successive inlines (32) exceeded,
Maybe this is caused by a recursive inline method?
You can use -Xmax-inlines to change the limit.
```

Example: 
https://scastie.scala-lang.org/road21/DQVmLQVuQIO18zEde0Omcw/31

This pr provides derivation for longer data structures by unrolling inline loops.
The idea is stolen from:

https://github.com/getkyo/kyo/blob/c33fba7e3c4954533819d6a1fa68df414676ec0a/kyo-data/shared/src/main/scala/kyo/internal/TypeIntersection.scala#L120

And I was too lazy to unroll each `summon**` аnd made `Inliner` abstraction to implement unrolling logic only once.

### DIFF
--- a/modules/tests/shared/src/test/scala-3/io/circe/DerivesSuite.scala
+++ b/modules/tests/shared/src/test/scala-3/io/circe/DerivesSuite.scala
@@ -258,14 +258,92 @@ object DerivesSuite {
     v31: String,
     v32: String,
     v33: String
-  ) derives Encoder,
+  ) derives Encoder.AsObject,
         Decoder
+
+  object LongClass:
+    given Eq[LongClass] = Eq.fromUniversalEquals
+    given Arbitrary[LongClass] = Arbitrary {
+      for
+        s1 <- Arbitrary.arbitrary[String]
+        s2 <- Arbitrary.arbitrary[String]
+        s3 <- Arbitrary.arbitrary[String]
+        s4 <- Arbitrary.arbitrary[String]
+        s5 <- Arbitrary.arbitrary[String]
+        s6 <- Arbitrary.arbitrary[String]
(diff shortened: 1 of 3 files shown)


hamnis (dismissed):
LGTM. Leaving merge to @zarthross

modules/tests/shared/src/test/scala-3/io/circe/DerivesSuite.scala
```diff
@@ -223,6 +223,84 @@ object DerivesSuite {
...
+  ) derives Encoder,
+        Decoder
+
+  enum LongEnum derives Encoder, Decoder:
+    case v1, v2, v3, v4, v5, v6, v7, v8, v9, v10,
+      v11, v12, v13, v14, v15, v16, v17, v18, v19, v20,
+      v21, v22, v23, v24, v25, v26, v27, v28, v29, v30,
+      v31, v32, v33
+
+  enum LongSum derives Encoder, Decoder:
```
zarthross, line 348: We need law tests on these encoders/decoders so we are 100% sure they round trip, just knowing it compiles isn't good enough.

zarthross (changes requested):
Needs some law tests on the added ASTs, but otherwise looks good.

modules/tests/shared/src/test/scala-3/io/circe/DerivesSuite.scala
road21: I have added tests

zarthross (approved):

PR #2353, opened 2025-02-27
add empty/singleton implementations of `JsonObject` to optimize memory usage

jnicoulaud-ledger:
This pull request optimizes memory usage in the edge case where there are many objects with only one field, default `LinkedHashMap` container has a lot of memory overhead in such cases.

Notes:
 - it looks like existing tests cover the new implementation correctly, let me know if more are needed
 - the benchmarks for folding/printing [use a single field object](https://github.com/circe/circe/blob/ad8546ab0349583cf249256d762e4a50c247307b/modules/benchmark/src/main/scala-2/io/circe/benchmark/ExampleData.scala#L35), so this will make it biased. Not sure if I should add a more complex object / different benchmark ?

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/JsonObject.scala
+++ b/modules/core/shared/src/main/scala/io/circe/JsonObject.scala
@@ -335,7 +335,7 @@ object JsonObject {
 
     override def add(k: String, j: Json): JsonObject = new LinkedHashMapJsonObject(
       {
-        val map = new LinkedHashMap[String, Json]
+        val map = new LinkedHashMap[String, Json](2)
         map.put(field, value)
         map.put(k, j)
         map
--- a/modules/core/shared/src/main/scala/io/circe/JsonObject.scala
+++ b/modules/core/shared/src/main/scala/io/circe/JsonObject.scala
@@ -258,7 +258,7 @@ object JsonObject {
       val fieldsSeq = fields.iterator.toSeq
       if (fields.sizeIs == 1) new SingletonJsonObject(fieldsSeq.head._1, fieldsSeq.head._2)
       else {
-        val map = new LinkedHashMap[String, Json]()
+        val map = new LinkedHashMap[String, Json](fields.size)
         val it = fields.iterator
         while (it.hasNext) {
           val (key, value) = it.next()
--- a/modules/core/shared/src/main/scala/io/circe/JsonObject.scala
+++ b/modules/core/shared/src/main/scala/io/circe/JsonObject.scala
@@ -258,8 +258,8 @@ object JsonObject {
       case 1 =>
         val (key, value) = fieldsSeq.head
         new SingletonJsonObject(key, value)
-      case _ =>
-        val map = new LinkedHashMap[String, Json](fields.size)
+      case size =>
+        val map = new LinkedHashMap[String, Json](size)
         val it = fields.iterator
         while (it.hasNext) {
           val (key, value) = it.next()
--- a/modules/core/shared/src/main/scala/io/circe/JsonObject.scala
+++ b/modules/core/shared/src/main/scala/io/circe/JsonObject.scala
@@ -344,13 +344,16 @@ object JsonObject {
     override def +:(field: (String, Json)): JsonObject =
       add(field._1, field._2)
 
-    override def remove(key: String): JsonObject =
-      empty
+    override def remove(key: String): JsonObject = {
+      if (key == field) empty
+      else this
+    }
 
     override def traverse[F[_]](f: Json => F[Json])(implicit F: Applicative[F]): F[JsonObject] =
       F.map(f(value))(new SingletonJsonObject(field, _))
 
-    override def mapValues(f: Json => Json): JsonObject = new SingletonJsonObject(field, f(value))
+    override def mapValues(f: Json => Json): JsonObject =
+      new SingletonJsonObject(field, f(value))
 
     override private[circe] def appendToFolder(folder: Printer.PrintingFolder): Unit = {
       val originalDepth = folder.depth
--- a/modules/core/shared/src/main/scala/io/circe/JsonObject.scala
+++ b/modules/core/shared/src/main/scala/io/circe/JsonObject.scala
@@ -305,6 +305,10 @@ object JsonObject {
   implicit final val showJsonObject: Show[JsonObject] = Show.fromToString
   implicit final val eqJsonObject: Eq[JsonObject] = Eq.fromUniversalEquals
 
+  /**
+   * An implementation of [[JsonObject]] for objects containing a single field, with a lighter memory footprint
+   * than map based implementations.
+   */
   private[this] final class SingletonJsonObject(field: String, value: Json) extends JsonObject {
     override private[circe] def applyUnsafe(k: String): Json =
       if (k == field) value else null
--- a/modules/core/shared/src/main/scala/io/circe/JsonObject.scala
+++ b/modules/core/shared/src/main/scala/io/circe/JsonObject.scala
@@ -298,7 +298,7 @@ object JsonObject {
   /**
    * Construct an empty [[JsonObject]].
    */
-  final val empty: JsonObject = new MapAndVectorJsonObject(Map.empty, Vector.empty)
+  final val empty: JsonObject = EmptyJsonObject
 
   /**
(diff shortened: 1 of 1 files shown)


plokhotnyuk:
@jnicoulaud-ledger If you're looking to enhance the performance of parsing and serialization with circe AST even more, I'd like to introduce you to the [jsoniter-scala-circe project](https://github.com/plokhotnyuk/jsoniter-scala/blob/master/jsoniter-scala-circe/shared/src/test/scala/com/github/plokhotnyuk/jsoniter_scala/circe/JsoniterScalaCodecSpec.scala). This library can significantly reduce latency and boost throughput, especially when using the [jsoniter-core API](https://github.com/plokhotnyuk/jsoniter-scala/blob/master/jsoniter-scala-core/jvm/src/main/scala/com/github/plokhotnyuk/jsoniter_scala/core/package.scala) for parsing and serialization from/to byte arrays or java.nio.ByteBuffers. Additionally, it offers improved safety, particularly when handling BigDecimal numbers, as discussed [here](https://www.reddit.com/r/scala/comments/1gmq32s/dos_when_decoding_large_bigdecimal_values_using/).

Furthermore, the [jsoniter-scala-benchmark module](https://github.com/plokhotnyuk/jsoniter-scala/tree/master/jsoniter-scala-benchmark) can be used to verify that there are no performance regressions in your pull requests. It includes real-world message samples from Google Maps, GitHub Actions, and Twitter APIs, ensuring that your changes maintain optimal performance.

modules/core/shared/src/main/scala/io/circe/JsonObject.scala
```diff
@@ -253,19 +252,20 @@ object JsonObject {
...
-
-      while (iterator.hasNext) {
-        val (key, value) = iterator.next()
-
-        map.put(key, value)
-      }
-
-      fromLinkedHashMap(map)
+    val fieldsSeq = fields.iterator.toSeq
+    (fieldsSeq.size: @switch) match {
```
zarthross: I think you should avoid converting the `fields` into a seq, just to use `size`, use `sizeCompare` or `sizeIs` on the `Iterable` and determine if its 0 or 1, or greater instead... this way you don't have to materialize the Seq, and `sizeCompare` is cheap in comparison.

modules/core/shared/src/main/scala/io/circe/JsonObject.scala
```diff
@@ -253,19 +252,20 @@ object JsonObject {
...
+      case 0 => empty
```
zarthross: `empty` is actually sub-optimial as well, maybe we could create a dedicated `empty` type as well instead of it being `final val empty: JsonObject = new MapAndVectorJsonObject(Map.empty, Vector.empty)`

modules/core/shared/src/main/scala/io/circe/JsonObject.scala
```diff
@@ -275,17 +275,21 @@ object JsonObject {
...
-    if (map.isEmpty) {
-      empty
-    } else {
-      fromMapAndVector(map, map.keys.toVector)
-    }
+    fromMapAndVector(map, map.keys.toVector)
 
   private[circe] final def fromMapAndVector(map: Map[String, Json], keys: Vector[String]): JsonObject =
-    new MapAndVectorJsonObject(map, keys)
+    (map.size: @switch) match {
```
zarthross: Same critique on size above, use sizeCompare or sizeIs instead so a full traversal isn't used when it isn't needed.

modules/core/shared/src/main/scala/io/circe/JsonObject.scala
```diff
@@ -275,17 +275,21 @@ object JsonObject {
...
+      case 0 => empty
+      case 1 => new SingletonJsonObject(map.keys.iterator.next(), map.values.iterator.next())
+      case _ => new MapAndVectorJsonObject(map, keys)
+    }
 
   private[circe] final def fromLinkedHashMap(map: LinkedHashMap[String, Json]): JsonObject =
-    new LinkedHashMapJsonObject(map)
+    (map.size: @switch) match {
```
zarthross, line 292: Also this.

modules/core/shared/src/main/scala/io/circe/JsonObject.scala
jnicoulaud-ledger: thanks, changed

modules/core/shared/src/main/scala/io/circe/JsonObject.scala
jnicoulaud-ledger: thanks, changed to `sizeIs`

modules/core/shared/src/main/scala/io/circe/JsonObject.scala
jnicoulaud-ledger: For a Java Map, I don't think I can do better ? `size` is cached, and `keySet`/`values` are lazy

modules/core/shared/src/main/scala/io/circe/JsonObject.scala
jnicoulaud-ledger: ok, I added one

jnicoulaud-ledger:
@plokhotnyuk 
> @jnicoulaud-ledger If you're looking to enhance the performance of parsing and serialization with circe AST even more, I'd like to introduce you to the [jsoniter-scala-circe project](https://github.com/plokhotnyuk/jsoniter-scala/blob/master/jsoniter-scala-circe/shared/src/test/scala/com/github/plokhotnyuk/jsoniter_scala/circe/JsoniterScalaCodecSpec.scala). This library can significantly reduce latency and boost throughput, especially when using the [jsoniter-core API](https://github.com/plokhotnyuk/jsoniter-scala/blob/master/jsoniter-scala-core/jvm/src/main/scala/com/github/plokhotnyuk/jsoniter_scala/core/package.scala) for parsing and serialization from/to byte arrays or java.nio.ByteBuffers. Additionally, it offers improved safety, particularly when handling BigDecimal numbers, as discussed [here](https://www.reddit.com/r/scala/comments/1gmq32s/dos_when_decoding_large_bigdecimal_values_using/).

Thank you, yes I  am aware of `jsoniter-scala-circe` :+1: Note this PR is not directly about performance though but memory footprint, I have a use case where I have no choice but to hold a big `Json` object in memory with many singleton sub-objects. Code is not easily refactorable to use jsoniter or streaming decoding.

> Furthermore, the [jsoniter-scala-benchmark module](https://github.com/plokhotnyuk/jsoniter-scala/tree/master/jsoniter-scala-benchmark) can be used to verify that there are no performance regressions in your pull requests. It includes real-world message samples from Google Maps, GitHub Actions, and Twitter APIs, ensuring that your changes maintain optimal performance.

OK, I'll try to test it. Is it as easy as `publishLocal` on circe, then change version in `jsoniter-scala-benchmark` and run the scripts ?

jnicoulaud-ledger:
I ran a subset of the benchmark (`*circe*`) on scala 2.13, attaching test script and results: https://gist.github.com/jnicoulaud-ledger/ff2dfba065a0cdc3acef4285ab0570ce
Seeing pretty even regressions and improvements, not sure if due to my test conditions which were not optimal, probably needs a re-run with more iterations...
image

modules/core/shared/src/main/scala/io/circe/JsonObject.scala
```diff
@@ -275,31 +275,147 @@ object JsonObject {
...
+    override def toMap: Map[String, Json] =
+      Map(field -> value)
+
+    override def toIterable: Iterable[(String, Json)] =
+      Iterable.single(field -> value)
+
+    private def addOrReplace(k: String, j: Json, append: Boolean): JsonObject = {
+      if (k == field) new SingletonJsonObject(k, j)
+      else {
+        val fields = Map.from(Seq((field, value), (k, j)))
```
zarthross, line 372: Looks to me like these are equivalent, so lets use the simpler one.
```suggestion
        val fields = Map((field, value), (k, j))
```

zarthross:
Looks like you need to add in scala-collection-compat to get sizeIs on scala 2.12.  You may also need to play with ` sizeHintIfCheap` in scala 2.12... but I don't see that in 2.13... Maybe try collection-compat and run those benchmarks to see if there was a negative impact?

PR #2367, opened 2025-03-27
Introducing "downFields" for HCursor

otobrglez:
Hey!

I've been a delighted user of Circe for years now. Love it!

I keep observing the pattern that I often just want to fetch a field from some deeply nested value (sometimes even with a "dynamic path"). So, I'm thinking if we could perhaps add a method next to `downField`, named `downFields` that would essentially achieve the following:

```scala
val nested: Json = Json.obj("hello" -> Json.obj("world" -> Json.obj("foo" -> Json.fromString("bar"))))

// Current example with downField
val cursorA: ACursor = 
  HCursor.fromJson(nested).downField("hello").downField("world").downField("foo")

// Simpler syntax with downFields
val cursorB: ACursor = 
  HCursor.fromJson(nested).downFields("hello", "world", "foo")
```

I've attached my implementation and tests to prove it works.

Please consider it and let me know if additional discussion or work is needed.

Regards,

\- Oto

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/ACursor.scala
+++ b/modules/core/shared/src/main/scala/io/circe/ACursor.scala
@@ -229,7 +229,7 @@ abstract class ACursor(private val lastCursor: HCursor, private val lastOp: Curs
    * @return A cursor positioned at the final field if all specified fields exist,
    *         or a failed cursor if any field in the sequence cannot be found.
    */
-  def downFields(ks: String*): ACursor
+  def downFields(k: String, ks: String*): ACursor
 
   private[circe] final def pathToRoot: PathToRoot = {
     import PathToRoot._
--- a/modules/core/shared/src/main/scala/io/circe/ACursor.scala
+++ b/modules/core/shared/src/main/scala/io/circe/ACursor.scala
@@ -229,7 +229,9 @@ abstract class ACursor(private val lastCursor: HCursor, private val lastOp: Curs
    * @param ks Additional field names to navigate into, in sequence.
    * @return An updated cursor focused on the final field in the sequence, or a failed cursor if any field is not found.
    */
-  def downFields(k: String, ks: String*): ACursor
+  def downFields(k: String, ks: String*): ACursor = {
+    ks.foldLeft(this.downField(k))(_.downField(_))
+  }
 
   private[circe] final def pathToRoot: PathToRoot = {
     import PathToRoot._
--- a/modules/core/shared/src/main/scala/io/circe/ACursor.scala
+++ b/modules/core/shared/src/main/scala/io/circe/ACursor.scala
@@ -222,6 +222,15 @@ abstract class ACursor(private val lastCursor: HCursor, private val lastOp: Curs
    */
   def downField(k: String): ACursor
 
+  /**
+   * Navigate down through multiple fields in a JSON object sequentially.
+   *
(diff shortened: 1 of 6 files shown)


modules/core/shared/src/main/scala/io/circe/ACursor.scala
```diff
@@ -222,6 +222,15 @@ abstract class ACursor(private val lastCursor: HCursor, private val lastOp: Curs
...
   def downField(k: String): ACursor
 
+  /**
+   * Navigate down through multiple fields in a JSON object sequentially.
+   *
+   * @param ks The sequence of field names to navigate through.
+   * @return A cursor positioned at the final field if all specified fields exist,
+   *         or a failed cursor if any field in the sequence cannot be found.
+   */
+  def downFields(ks: String*): ACursor
```
zarthross: `downFields` should require at least one, this way you don't have to do the is empty check later.
```suggestion
  def downFields(k: String, ks: String*): ACursor
```

modules/core/shared/src/main/scala/io/circe/HCursor.scala
```diff
@@ -93,6 +93,15 @@ abstract class HCursor(lastCursor: HCursor, lastOp: CursorOp) extends ACursor(la
...
   }
 
+  final def downFields(ks: String*): ACursor = {
+    if (ks.isEmpty) {
+      fail(CursorOp.DownField(""))
+    } else {
+      val (head, tail) = ks.toList.splitAt(1)
+      tail.foldLeft(downField(head.head))(_.downField(_))
+    }
+  }
```
zarthross:
```suggestion
  final def downFields(k: String, ks: String*): ACursor = {
    ks.foldLeft(this.downField(k))(_.downField(_))
  }
```

zarthross (changes requested):

modules/core/shared/src/main/scala/io/circe/HCursor.scala
otobrglez: Thank you! 🙌

modules/core/shared/src/main/scala/io/circe/ACursor.scala
```diff
@@ -222,6 +222,15 @@ abstract class ACursor(private val lastCursor: HCursor, private val lastOp: Curs
...
   def downField(k: String): ACursor
 
+  /**
+   * Navigates down into the specified fields of a JSON object sequentially.
+   *
+   * @param k  The first field name to navigate into.
+   * @param ks Additional field names to navigate into, in sequence.
+   * @return An updated cursor focused on the final field in the sequence, or a failed cursor if any field is not found.
+   */
+  def downFields(k: String, ks: String*): ACursor
```
zarthross: To address the mima warning that showed up in CI, we need to make sure any new methods added to traits have a default implementation  I think the implementation used by `HCursor`  could probably just be moved here and solve the problem. 

```suggestion
  def downFields(k: String, ks: String*): ACursor = {
    ks.foldLeft(this.downField(k))(_.downField(_))
  }
```

otobrglez: Ok. I'll follow this suggestion. 👌

hamnis (approved):

zarthross (approved):

PR #2369, opened 2025-04-02
Fix superfluous keys not being ignored

### LINKED ISSUE #2368: Scala 3: incosistent behavior of Decoder for sealed traits in presence of additional fields.
```scala
import io.circe.literal.*

object TestCirceScala3 extends App {
  sealed trait Case

  object Case {
    case class C1(b: Int) extends Case

    case class C2(a: String) extends Case

    implicit val decoder: io.circe.Decoder[Case] = deriveDecoder
  }

  println(json"""{"C2" : {"a" : "", "extraFieldC2":  true}}""".as[Case]) // (1)
  println(json"""{"C2" : {"a" : "", "extraFieldC2":  true}, "extraField" : ""}""".as[Case]) // (2)
  println(json"""{"extraField" : "", "C2" : {"a" : "", "extraFieldC2":  true}}""".as[Case]) // (3)
}
```
prints
```
Right(C2())
Right(C2())
Left(DecodingFailure at .extraField: type Case has no class/object/case named 'extraField'.)
```

In (1) C2 is correctly decoded (`extraFieldC2` is ignored).
Same for (2), where the decoding stops before reaching `extraField`.
In (3) decoding fails.
The only difference between the inputs of (2) and (3) is the ordering of the fields in the Json object. 

Expected behavior: (3) should decode as `Right(C2())`. This is what happens in Scala 2.

MartinHH:
Fix superfluous keys not being ignored when decoding ADTs/Enums.

Fixes #2368

### DIFF
--- a/modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredDecoder.scala
+++ b/modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredDecoder.scala
@@ -23,6 +23,7 @@ import cats.data.{ NonEmptyList, Validated }
 import io.circe.{ ACursor, Decoder, DecodingFailure, HCursor }
 import io.circe.DecodingFailure.Reason.WrongTypeExpectation
 import cats.implicits.*
+import scala.annotation.tailrec
 import scala.collection.immutable.Map
 import scala.quoted.*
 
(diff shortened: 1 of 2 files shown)


zarthross (changes requested):
You need to check that `conf.strictDecoding` is false, when strict decoding is true the correct behavior is rejecting the json if it has extra fields.

MartinHH:
> You need to check that `conf.strictDecoding` is false, when strict decoding is true the correct behavior is rejecting the json if it has extra fields.

That's already checked. The only case where `fromName(...)` is called with (potentially) non-empty `fallBacks` is the else-branch of `if iter.hasNext && conf.strictDecoding` - see https://github.com/MartinHH/circe/blob/db662f79d231b3aa0475aa8756af10181e57615a/modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredDecoder.scala#L87

zarthross (approved):
`Configuration#strictDecoding should fail for sum types when the json object has more than one field` does look like it covers the area I was concerned about.

Thanks for your work!

hamnis (approved):

PR #2381, opened 2025-06-07
Simplify strict decoding and make it not fail fast in the accumulating variant

### LINKED ISSUE #2379: Decoders derived with Configuration#strictDecoding fail fast at `decodeAccumulating`
I'm expecting to get a comprehensive list of failures, as was the case in Scala 2 with https://github.com/circe/circe-generic-extras. Reproducible example with `scala-cli`:

```scala
//> using jvm 21.0.6
//> using scala 3.6.4
//> using dep io.circe::circe-core:0.14.13

import io.circe.*
import io.circe.derivation.*

case class Foo(a: String, b: Int, c: Int)

object Main {

    val fooDecoder: Decoder[Foo] = {
      given Configuration = Configuration.default.withStrictDecoding
      Decoder.derivedConfigured
    }

    val json = Json.obj(
      "a" -> Json.fromString("hello"),
      "b" -> Json.fromInt(42),
      "c" -> Json.fromString("100"), // This should cause a decoding error because c is expected to be an Int
      "d" -> Json.fromString("extra field") // This field is not in the case class Foo
    )

    val decoded = fooDecoder.decodeJson(json)
    val accDecoded = fooDecoder.decodeAccumulating(json.hcursor)

    println(decoded) // Left(DecodingFailure at : Strict decoding Foo - unexpected fields: d; valid fields: a, b, c.)
                     // ok

    println(accDecoded) // Invalid(NonEmptyList(DecodingFailure at : Strict decoding Foo - unexpected field: d; valid fields: a, b, c.))
                        // Expected to also get a DecodingFailure at .c
  }
}
```

bdmendes:
Closes #2379. This is a **breaking change** (at least for those who depend on parsing the messages) to the behavior of the `decodeAccumulating` API, for `Decoder`s derived with a configuration (Scala 3-only):

- Unexpected fields are reported under a single failure. For example, for a product `Foo(a, b)`, if a `c` and a `d` fields are given, we report `unexpected fields: c,d; valid fields: a,b.` instead of `unexpected field: c; valid fields: a,b.` and `unexpected field: d; valid fields: a,b.`. This seems cleaner to me and makes sense under the semantic "a single verification for expected fields" which is exactly what the code is doing, but was artificially split for the accumulating variant. (As far as I can tell, the accumulating variant implies "do not fail fast", not "split errors", although this is not well documented.)
- Even if the strict matching fails, expected fields are now validated. This was the behavior in Scala 2 in https://github.com/circe/circe-generic-extras and I'm not sure they are aware it got lost in https://github.com/circe/circe-generic-extras/pull/401 when resorting to the new core API. I noticed it when migrating a project to Scala 3.

Please let me know if this makes sense. I'm open to suggestions.

### DIFF
--- a/modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredDecoder.scala
+++ b/modules/core/shared/src/main/scala-3/io/circe/derivation/ConfiguredDecoder.scala
@@ -118,56 +118,52 @@ trait ConfiguredDecoder[A](using conf: Configuration) extends Decoder[A]:
   /** Ensures cursor is a json object, handles strict decoding. */
   private def decodeProductBase[R](
     c: HCursor,
-    fail: DecodingFailure => R,
-    strictFail: (List[String], IndexedSeq[String]) => R
-  )(decodeProduct: => R): R =
+    fail: DecodingFailure => R
+  )(decodeProduct: Option[DecodingFailure] => R): R =
     c.value.isObject match
       case false                        => fail(DecodingFailure(WrongTypeExpectation("object", c.value), c.history))
-      case true if !conf.strictDecoding => decodeProduct
+      case true if !conf.strictDecoding => decodeProduct(None)
       case true                         =>
         val expectedFields = elemLabels.toIndexedSeq.map(conf.transformMemberNames) ++ conf.discriminator
         val expectedFieldsSet = expectedFields.toSet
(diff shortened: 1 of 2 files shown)


hamnis:
I think this is reasonable.

hamnis (approved):

hamnis:
@zarthross do you agree?

zarthross (approved):

PR #2385, opened 2025-06-16
Replace deprecated `JavaConverters` with `CollectionConverters`

ikeyan:
This change addresses deprecation introduced in Scala 2.13, where JavaConverters was replaced in favor of CollectionConverters. Aside from the updated import path, this avoids unintended type inference issues caused by implicit conversions.
This update is limited to test code and does not affect runtime behavior.


ikeyan:
scala2.12 does not have CollectionConverters yet

PR #2414, opened 2025-11-11
Optimise `JsonObject`'s `equals` and `hashCode`

RafalSumislawski:
The current implementation of `equals` and `hashCode` is based on `toMap` which is efficient when dealing with `MapAndVectorJsonObject`, but when dealing with `LinkedHashMapJsonObject` it leads to time consuming, and allocation heavy `toMapAndVectorJsonObject` conversion. The result of the conversion is discarded, so use cases that repeatedly hash or compare `JsonObject`s keep paying that cost repeatedly. (I have a real world application that spends 51% of CPU time in GC because of that)

In this PR I propose a new implementation of `equals` and `hashCode` which is aware of the two representations of `JsonObject` and chooses to optimal way to hash/compare them without significant heap allocations. In addition to that I prepared benchmarks to justify the change and quantify the improvement.

 ### Benchmark results before (ccd1bcfc63922fe6bd6c32a8c35a0088bc91dbb5) (higher is better)

```
[info] Benchmark                                           Mode  Cnt       Score      Error  Units
[info] JsonObjectEqualityBenchmark.equalsLinkedHashMap    thrpt   20   110165.154 ±   462.699  ops/s
[info] JsonObjectEqualityBenchmark.equalsMapAndVector     thrpt   20  1565309.972 ± 32705.186  ops/s
[info] JsonObjectEqualityBenchmark.equalsMixed            thrpt   20   201005.477 ±  6505.266  ops/s
[info] JsonObjectEqualityBenchmark.hashCodeLinkedHashMap  thrpt   20   111157.147 ±  1450.116  ops/s
[info] JsonObjectEqualityBenchmark.hashCodeMapAndVector   thrpt   20   254723.666 ±  7074.196  ops/s
```

### Benchmark results after (cb6fb7c95ab2ae987632a179a15a8d57cb853f25) (higher is better)

```
[info] Benchmark                                           Mode  Cnt       Score      Error  Units
[info] JsonObjectEqualityBenchmark.equalsLinkedHashMap    thrpt   20  1027442.106 ±  36823.173  ops/s
[info] JsonObjectEqualityBenchmark.equalsMapAndVector     thrpt   20  1558090.245 ±  33794.294  ops/s
[info] JsonObjectEqualityBenchmark.equalsMixed            thrpt   20   865587.917 ±  11020.718  ops/s
[info] JsonObjectEqualityBenchmark.hashCodeLinkedHashMap  thrpt   20   254424.508 ±   3489.809  ops/s
[info] JsonObjectEqualityBenchmark.hashCodeMapAndVector   thrpt   20   264467.414 ±   3898.601  ops/s
```

### Summary

This resulted in 

* 9.3 times faster comparison of `LinkedHashMapJsonObject`, 
* 4.3 times faster comparison of `LinkedHashMapJsonObject` with `MapAndVectorJsonObject` 
* 2.2 times faster hashing of `LinkedHashMapJsonObject`

image

### DIFF
--- a/modules/benchmark/src/main/scala-2/io/circe/benchmark/JsonObjectEqualityBenchmark.scala
+++ b/modules/benchmark/src/main/scala-2/io/circe/benchmark/JsonObjectEqualityBenchmark.scala
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2024 circe
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.circe.benchmark
+
+import io.circe.{Json, JsonObject}
+import java.util.concurrent.TimeUnit
+import org.openjdk.jmh.annotations.*
+import java.util
+
+/**
+ * Benchmark equality- and hash-based operations on JSON objects
+ *
+ * The following command will run the benchmarks with reasonable settings:
+ *
+ * > sbt "benchmark/jmh:run -i 10 -wi 10 -f 2 -t 1 io.circe.benchmark.JsonObjectEqualityBenchmark"
+ */
+@State(Scope.Thread)
+@BenchmarkMode(Array(Mode.Throughput))
+@OutputTimeUnit(TimeUnit.SECONDS)
+@Fork(value = 1, jvmArgsAppend = Array("-Xms2G", "-Xmx2G", "-XX:+AlwaysPreTouch", "-XX:+UseSerialGC"))
+class JsonObjectEqualityBenchmark {
+
+  val count: Int = 100
+  val mapAndVectorObject1: JsonObject = buildMapAndVectorObject()
+  val mapAndVectorObject2: JsonObject = buildMapAndVectorObject()
+  val linkedHashMapObject1: JsonObject = buildLinkedHashMapObject()
+  val linkedHashMapObject2: JsonObject = buildLinkedHashMapObject()
+
+  private def buildMapAndVectorObject(): JsonObject = {
+    val orderedFields: Vector[(String, Json)] =
+      Vector.tabulate(count)(i => (i.toString, Json.fromInt(i)))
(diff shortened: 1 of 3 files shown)


RafalSumislawski:
@zarthross @hamnis
Quick ping on this PR—whenever you have time. Thanks!

mzuehlke:
One of the new test is failing in CI

zarthross (commented):
I think this code looks fine in principle, my only complaints here are maybe a bit on ease with which future subtypes of JsonObject might get added.

Since the base trait no longer implements equals and hashcode it'll be up to the implementer and reviewer to remember to add in the additional cases to the existing subclasses and the new one.  If they forget and we don't notice in a PR then ... not great.

I think maybe a better approach here might be to implement `cats.kernel.Eq` and `cats.kernel.Hash` and have `JsonObject` call into those implementations.  This way Eq and Hash can be implemented with a pattern match that can have exhaustivity checks done and ensure the update is made correctly.  Though I'm not sure what the performance ramifications might be, i doubt its too bad given it'll only be 1 extra call.

RafalSumislawski:
> One of the new test is failing in CI

I was to strict with trying to fix the hashCode to a specific value. I think it already produces different values depending on JVM/Scala version (not sure which one). I will redefine the test to fix that.

> I think maybe a better approach here might be to implement cats.kernel.Eq and cats.kernel.Hash and have JsonObject call into those implementations. This way Eq and Hash can be implemented with a pattern match that can have exhaustivity checks done and ensure the update is made correctly. Though I'm not sure what the performance ramifications might be, i doubt its too bad given it'll only be 1 extra call.

I'll try that and verify the performance impact.

Thanks for the review 🙇

RafalSumislawski:
@zarthross 
I can't use `Eq` to express comparison of a `LinkedHashMapJsonObject` with a `MapAndVectorJsonObject`. I still introduced pattern matches with exhaustiveness checks. Had to change `LinkedHashMapJsonObject` and `MapAndVectorJsonObject` from `private[this]` to `private`, otherwise we can't pattern match on these types. There is no noticeable perf impact.

```
[info] JsonObjectEqualityBenchmark.equalsLinkedHashMap    thrpt   20  1044907.707 ±  22617.112  ops/s
[info] JsonObjectEqualityBenchmark.equalsMapAndVector     thrpt   20  1557944.126 ±  34411.843  ops/s
[info] JsonObjectEqualityBenchmark.equalsMixed            thrpt   20   865205.733 ±  19166.916  ops/s
[info] JsonObjectEqualityBenchmark.hashCodeLinkedHashMap  thrpt   20   250597.282 ±   3936.374  ops/s
[info] JsonObjectEqualityBenchmark.hashCodeMapAndVector   thrpt   20   261374.513 ±   2024.198  ops/s
```

zarthross (approved):

PR #2433, opened 2026-02-18
Fix NumberFormatException when parsing leading-dot-zero decimals

### LINKED ISSUE #2432: Exceptions when parsing zero-valued BiggerDecimal starting with decimal point
Related to https://github.com/circe/circe/issues/600 and https://github.com/circe/circe/issues/2132, I think we're not addressing all edge cases. I found `NumberFormatException` to be thrown when parsing zero values starting with a decimal point.

Running `scala-cli` snippet below:
```
//> using dep io.circe::circe-core:0.14.15

import io.circe.Json

@main def run(): Unit = {
  val malformedValues = List(
      // Decoding failure
      "", " ", "abc", "N/A", "null", "1e", "1e-", "1e+", "1E", "1E-", "0.", "-0.", "-", "-.", "-1e", "-1e-", "-1e+",
      // No decoding failure?
      "-.5", ".5",
      // Exception
      "-.0", ".0"
    )

  malformedValues.foreach { v =>
    print(s"""Json.fromString("$v").as[Double] => """)
    try {
      val result = Json.fromString(v).as[Double]
      println(s"$result")
    } catch {
      case e: NumberFormatException =>
        println(s"THROWS ${e.getClass.getSimpleName}: ${e.getMessage}")
    }
  }
}
```
We get decoding failures for those invalid values but then get an exception when parsing `-.0` and `.0`:
```
Json.fromString("").as[Double] => Left(DecodingFailure at : Double)
Json.fromString(" ").as[Double] => Left(DecodingFailure at : Double)
...
Json.fromString("-.5").as[Double] => Right(-0.5)
Json.fromString(".5").as[Double] => Right(0.5)
Json.fromString("-.0").as[Double] => THROWS NumberFormatException: Zero length BigInteger
Json.fromString(".0").as[Double] => THROWS NumberFormatException: Zero length BigInteger
```

This is due to `BiggerDecimal.parseBiggerDecimalUnsafe`, which is used underneath. We can replicate it directly using it as well:
```
//> using dep io.circe::circe-numbers:0.14.15

import io.circe.numbers.BiggerDecimal

@main def run(): Unit = {
  val inputs = List(".0", "-.0", ".00", "-.00", ".001", "-.001",".5", "-.5")

  inputs.foreach { v =>
    print(s"""parseBiggerDecimalUnsafe("$v") => """)
    try {
      val result = BiggerDecimal.parseBiggerDecimalUnsafe(v)
      println(if (result == null) "null" else result.toString)
    } catch {
      case e: NumberFormatException =>
        println(s"THROWS ${e.getClass.getSimpleName}: ${e.getMessage}")
    }
  }
}
```
~~Additionally, as you can see above, I also found some inconsistent behaviour: we can successfully decode `-.5` and `.5`, for example. I'm not sure whether this is a design decision or not, but I expected these two cases to be handled in the same fashion as they are both numbers without a leading digit before the decimal point. Either both should throw, or both should be successfully decoded. But this probably deserves another issue, if confirmed to be unwanted behaviour.~~ See comment below.

afonsoafonsoafonso:
When parsing `.0` or `-.0`, for example, stripping the trailing zero from the unscaled string leaves an empty string, causing `new BigInteger("")` to throw before the existing zero-check is reached. This fix addresses this by assigning `BigInteger.ZERO` to `unscaled` when the unscaled string is empty or `-`.

Should fix https://github.com/circe/circe/issues/2432.

### DIFF
--- a/modules/numbers/shared/src/main/scala/io/circe/numbers/BiggerDecimal.scala
+++ b/modules/numbers/shared/src/main/scala/io/circe/numbers/BiggerDecimal.scala
@@ -374,7 +374,8 @@ object BiggerDecimal {
             }
 
           val unscaledString = integral + fractional
-          val unscaled = new BigInteger(unscaledString.substring(0, unscaledString.length - zeros))
+          val trimmed = unscaledString.substring(0, unscaledString.length - zeros)
+          val unscaled = if (trimmed.isEmpty || trimmed == "-") BigInteger.ZERO else new BigInteger(trimmed)
 
           if (unscaled == BigInteger.ZERO) {
             if (input.charAt(0) == '-') NegativeZero else UnsignedZero
--- a/modules/numbers/shared/src/test/scala/io/circe/numbers/BiggerDecimalSuite.scala
+++ b/modules/numbers/shared/src/test/scala/io/circe/numbers/BiggerDecimalSuite.scala
@@ -277,4 +277,35 @@ class BiggerDecimalSuite extends ScalaCheckSuite {
       )
     }
   }
+
+  property("numbers without a leading digit before decimal point should be parseable") {
+    val genSign = Gen.oneOf("", "-")
+    val genFractional = for {
+      nonZero <- Gen.choose('1', '9')
+      rest <- Gen.listOf(Gen.numChar)
+      position <- Gen.choose(0, rest.length)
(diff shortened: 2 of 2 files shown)


zarthross (approved):
Nice low touch fix with plenty of tests.  Great work, thanks for the contribution!

PR #2439, opened 2026-03-11
Improve error messages for NonEmptyMap and Validated decoders

a-khakimov:
Replace cryptic type-parameter placeholders with human-readable messages that describe what was expected.

Before

```scala
decode[NonEmptyMap[String, Int]]("""{}""")
// Left(DecodingFailure at : [K, V]NonEmptyMap[K, V])

Decoder.decodeValidated[String, Int]("error", "value")
  .decodeJson(parse("""{"other": 1}""").getOrElse(Json.Null))
// Left(DecodingFailure at : [E, A]Validated[E, A])
```

After

```scala
decode[NonEmptyMap[String, Int]]("""{}""")
// Left(DecodingFailure at : NonEmptyMap: expected a non-empty JSON object)

Decoder.decodeValidated[String, Int]("error", "value")
  .decodeJson(parse("""{"other": 1}""").getOrElse(Json.Null))
// Left(DecodingFailure at : Validated[E, A]: expected an object with either a failure or success key)
```

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -1188,7 +1188,7 @@ object Decoder
       final protected def createBuilder(): Builder[(K, V), SortedMap[K, V]] =
         SortedMap.newBuilder[K, V](Order.catsKernelOrderingForOrder(orderK))
     }.emap { map =>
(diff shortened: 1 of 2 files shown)


hamnis:
a test would be nice to avoid regressing

a-khakimov:
@hamnis thanks, added tests for both error messages

hamnis (approved):

PR #2440, opened 2026-03-12
Fix wrong NonEmptyX decoding failure for non-array inputs

### LINKED ISSUE #1978: Wrong `NonEmptyX` decoding failure
Hello!

I have experienced an unexpected behavior when decoding over a `NonEmptyX` type. Consider the following case:

```
final case class Tasks(tasks: List[Task])
final case class Task(name: String, description: Option[String], done: Boolean)
```

If I try to decode a string into a `Tasks` instance in both correct and wrong ways, it behaves as expected:

```
val validMessage: String = """{"tasks":[{"name":"task1","done":false},{"name":"task2","description":"this is the task 2","done":true}]}"""
val invalidMessage: String = """{"tasks":"bad value"}"""

println(decodeAccumulating[Tasks](validMessage))
// Valid(Tasks(List(Task(task1,None,false), Task(task2,Some(this is the task 2),true))))

println(decodeAccumulating[Tasks](invalidMessage))
// Invalid(NonEmptyList(DecodingFailure(Got value '"bad value"' with wrong type, expecting array, List(DownField(tasks)))))
```

But if I try to use a `NonEmptyList` (or any other `NonEmptyX` type) instead of `List`:

```
final case class Tasks(tasks: NonEmptyList[Task])
final case class Task(name: String, description: Option[String], done: Boolean)
```

then the decoding failures are not the expected ones:

```
val validMessage: String = """{"tasks":[{"name":"task1","done":false},{"name":"task2","description":"this is the task 2","done":true}]}"""
val invalidMessage: String = """{"tasks":"bad value"}"""

println(decodeAccumulating[Tasks](validMessage))
// Valid(Tasks(List(Task(task1,None,false), Task(task2,Some(this is the task 2),true))))

println(decodeAccumulating[Tasks](invalidMessage))
// Invalid(NonEmptyList(DecodingFailure(Missing required field, List(DownArray, DownField(tasks))), DecodingFailure(Missing required field, List(DownArray, DownField(tasks)))))
```

It looks like de decoding is not being able to recognize that there is a type mismatch at `tasks` field, so it is trying to go a step further considering that value a valid array and then trying to decode over a `Task`.

[Here](https://scastie.scala-lang.org/cPteneRYSteOFcvUXcLhGw) is the Scastie that I have used to test this behavior.

Thanks in advance!

a-khakimov:
Fixes #1978

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/NonEmptySeqDecoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/NonEmptySeqDecoder.scala
@@ -16,6 +16,9 @@
 
 package io.circe
 
+import cats.data.Validated
+import io.circe.DecodingFailure.Reason.WrongTypeExpectation
+
 import scala.collection.mutable.Builder
 
 private[circe] abstract class NonEmptySeqDecoder[A, C[_], S](decodeA: Decoder[A]) extends Decoder[S] { self =>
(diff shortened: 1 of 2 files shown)


hamnis:
needs a rebase or conflict resolution

a-khakimov:
@hamnis done

hamnis (approved):

zarthross (approved):

PR #2441, opened 2026-03-12
Accumulate errors in Decoder#or and Decoder#either decodeAccumulating

### LINKED ISSUE #2062: Decoder.decodeAccumulating discards errors in ADTs
The microsite shows [the standard way to write a `Decoder` for an ADT](https://circe.github.io/circe/codecs/adt.html) is to merge a list of `Decoder`s using `Decoder#or`:
```scala
  List[Decoder[Event]](
    Decoder[Foo].widen,
    Decoder[Bar].widen,
    Decoder[Baz].widen,
    Decoder[Qux].widen
  ).reduceLeft(_ or _)
```

This works when the JSON is correct, and produces misleading errors when the JSON structure is incorrect:
```scala
println(decode[Event]("""{ "i": null }""").show)
// Left(DecodingFailure at .values: Missing required field)
```
Which makes sense, as `decode` returns as single error. What makes less sense is that `decodeAccumulating` returns almost exactly the same thing:
```scala
println(decodeAccumulating[Event]("""{ "i": null }""").show)
// Invalid(NonEmptyList(DecodingFailure at .values: Missing required field))
```

A more useful error from `decodeAccumulating` could look like this:
```
Invalid(NonEmptyList(DecodingFailure at .i: Int, DecodingFailure at .s: Missing required field, DecodingFailure at .c: Missing required field, DecodingFailure at .values: Missing required field))
```

It looks like this could be accomplished with a change to `Decoder#or` (and likely a similar change to `Decoder#either`, for consistency):
```scala
// Existing
override def tryDecodeAccumulating(c: ACursor): Decoder.AccumulatingResult[AA] =
  self.tryDecodeAccumulating(c) match {
    case r @ Valid(_) => r
    case Invalid(_)   => d.tryDecodeAccumulating(c)
  }
// Proposed
override def tryDecodeAccumulating(c: ACursor): Decoder.AccumulatingResult[AA] =
  self.tryDecodeAccumulating(c) match {
    case r @ Valid(_) => r
    case Invalid(i)   => d.tryDecodeAccumulating(c).leftMap(i.concatNel)
  }
```

[Scastie with the above implemented as a Semigroup](https://scastie.scala-lang.org/doEOkJveRPKqJDBlbGU1RQ)

a-khakimov:
Fixes #2062

### DIFF
--- a/modules/core/shared/src/main/scala/io/circe/Decoder.scala
+++ b/modules/core/shared/src/main/scala/io/circe/Decoder.scala
@@ -305,7 +305,7 @@ trait Decoder[A] extends Serializable { self =>
     override def tryDecodeAccumulating(c: ACursor): Decoder.AccumulatingResult[AA] =
       self.tryDecodeAccumulating(c) match {
         case r @ Valid(_) => r
-        case Invalid(_)   => d.tryDecodeAccumulating(c)
+        case Invalid(e)   => d.tryDecodeAccumulating(c).leftMap(_.concatNel(e))
       }
   }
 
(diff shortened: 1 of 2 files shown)


a-khakimov:
Hi @hamnis, could you please take a look at this PR when you get a chance?

hamnis (approved):
This looks fine to me

a-khakimov:
@zarthross hi, would you mind reviewing my PR whenever you have time?

